PR #13, opened 2026-03-27
Task 2: Configuration (TOML file, CLI overrides, validation)

### LINKED ISSUE #2: Task 2: Configuration (TOML file, CLI overrides, validation)
## Overview

Implement config file loading from `~/.config/illustrious-manager/config.toml` with auto-creation, default values, CLI flag override merging, and validation.

## Blocked by
- #1 (Project setup & core types)

## Files
- Create: `src/config.rs`
- Modify: `src/main.rs`
- Create: `tests/config_test.rs`

## Acceptance Criteria

- [ ] `AppConfig` and `VertexConfig` structs with serde deserialization
- [ ] `load_config_from_path(path)` parses a TOML file into `AppConfig`
- [ ] `load_config(custom_path)` auto-creates the config template if missing
- [ ] `apply_overrides()` merges CLI flag values into config
- [ ] `validate()` returns an error if `project` is empty, with a helpful message pointing to the config file
- [ ] Defaults: `region = "us-east5"`, `model = "claude-sonnet-4-20250514"`
- [ ] 4 unit tests pass:
  - Parse valid config
  - Parse config with defaults
  - CLI overrides config values
  - Validation rejects empty project

## Implementation Notes

See `docs/superpowers/plans/2026-03-27-phase1-core-cli.md` — Task 2 for full code.

Fizzizist:
## Summary

Implements the configuration system for illustrious-manager with TOML file loading, CLI flag overrides, and validation.

## Changes

- **Configuration Loading**: `load_config_from_path()` parses TOML files into `AppConfig` struct
- **Auto-creation**: `load_config()` automatically creates default config template if missing
- **CLI Overrides**: `apply_overrides()` merges CLI flag values (--project, --region, --model) into loaded config
- **Validation**: `validate()` checks for required fields (project ID) with helpful error messages
- **Default Values**: region="us-east5", model="claude-sonnet-4-20250514"

## Testing

All 4 unit tests pass, covering:
- ✅ Parse valid config with all fields
- ✅ Parse config with defaults (only project specified)
- ✅ CLI overrides config file values
- ✅ Validation rejects empty project with helpful message

## Implementation Approach

Followed Test-Driven Development (TDD):
1. ✅ **RED**: Wrote tests first, watched them fail correctly
2. ✅ **GREEN**: Implemented minimal code to pass all tests
3. ✅ **REFACTOR**: Code is clean and follows project standards
4. ✅ **Quality**: Passed `cargo fmt` and `cargo clippy` checks

Closes #2

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -91,11 +91,17 @@ pub fn apply_overrides(
 }
 
 /// Validate that the config has all required fields.
-pub fn validate(config: &AppConfig) -> Result<()> {
+///
+/// `config_path`: The path that was actually used to load the config, for accurate error messages.
+/// Pass `None` if the default path was used.
+pub fn validate(config: &AppConfig, config_path: Option<&Path>) -> Result<()> {
     if config.vertex.project.is_empty() {
-        let path = default_config_path()
-            .map(|p| p.display().to_string())
-            .unwrap_or_else(|_| "~/.config/illustrious-manager/config.toml".to_string());
+        let path = match config_path {
+            Some(p) => p.display().to_string(),
+            None => default_config_path()
+                .map(|p| p.display().to_string())
+                .unwrap_or_else(|_| "~/.config/illustrious-manager/config.toml".to_string()),
+        };
         bail!(
             "GCP project ID is required. Set it in your config file at:\n  {}\n\nOr pass --project <PROJECT> on the command line.",
             path
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,105 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::{Context, Result, bail};
+
+const DEFAULT_REGION: &str = "us-east5";
+const DEFAULT_MODEL: &str = "claude-sonnet-4-20250514";
+
+const CONFIG_TEMPLATE: &str = r#"[vertex]
+# Required: your GCP project ID
+project = ""
+# Vertex AI region
+region = "us-east5"
+# Model to use
+model = "claude-sonnet-4-20250514"
+"#;
+
+#[derive(Debug, serde::Deserialize)]
+pub struct AppConfig {
+    pub vertex: VertexConfig,
+}
+
+#[derive(Debug, serde::Deserialize)]
+pub struct VertexConfig {
+    pub project: String,
+    #[serde(default = "default_region")]
+    pub region: String,
+    #[serde(default = "default_model")]
+    pub model: String,
+}
+
+fn default_region() -> String {
+    DEFAULT_REGION.to_string()
+}
+
+fn default_model() -> String {
+    DEFAULT_MODEL.to_string()
+}
+
+/// Returns the default config file path: ~/.config/illustrious-manager/config.toml
+pub fn default_config_path() -> Result<PathBuf> {
+    let config_dir = dirs::config_dir().context("Could not determine config directory")?;
+    Ok(config_dir.join("illustrious-manager").join("config.toml"))
+}
+
+/// Load config from a specific path.
+pub fn load_config_from_path(path: &Path) -> Result<AppConfig> {
+    let content = std::fs::read_to_string(path)
+        .with_context(|| format!("Failed to read config file: {}", path.display()))?;
+    let config: AppConfig = toml::from_str(&content)
+        .with_context(|| format!("Failed to parse config file: {}", path.display()))?;
+    Ok(config)
+}
+
+/// Load config from the default path, auto-creating if it doesn't exist.
+pub fn load_config(custom_path: Option<&Path>) -> Result<AppConfig> {
+    let path = match custom_path {
+        Some(p) => p.to_path_buf(),
+        None => default_config_path()?,
+    };
+
+    if !path.exists() {
+        if let Some(parent) = path.parent() {
+            std::fs::create_dir_all(parent).with_context(|| {
+                format!("Failed to create config directory: {}", parent.display())
+            })?;
+        }
+        std::fs::write(&path, CONFIG_TEMPLATE)
+            .with_context(|| format!("Failed to write default config: {}", path.display()))?;
+        eprintln!("Created default config at: {}", path.display());
+    }
+
+    load_config_from_path(&path)
+}
+
+/// Apply CLI flag overrides to the config.
+pub fn apply_overrides(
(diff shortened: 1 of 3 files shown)


Fizzizist (commented):
## Code Review: PR #13 — Configuration (TOML file, CLI overrides, validation)

**Verdict: Changes requested.** The core implementation is solid — config parsing, defaults, override merging, and validation all work. But there are gaps in test coverage, a bug in the error messaging, and a policy violation that need addressing.

---

### 1. Bug: `validate()` lies about the config file path

`validate()` unconditionally calls `default_config_path()` to construct its error message. If the user loaded config from a custom path (via `--config /my/custom/path.toml`), the error message will point them to `~/.config/illustrious-manager/config.toml` — a file they may not even be using. This is actively misleading.

The fix: `validate()` should accept the path that was actually used (or `None` for the default), and include that in the error message instead of always computing the default.

### 2. Policy violation: `.unwrap()` in tests

CLAUDE.md is unambiguous: _"The use of `.unwrap()` is forbidden under **all** circumstances."_ The tests use `.unwrap()` on every `NamedTempFile::new()`, `write!()`, and `.load_config_from_path()` call. Yes, panicking is how test failures propagate — but the policy says to use `.expect()` with an informative message when it's logically impossible for a panic to occur. These should all be `.expect("descriptive message")`.

### 3. Missing test: `load_config` auto-creation path

`load_config()` has a non-trivial code path that creates the config directory, writes a template file, and prints to stderr. Zero tests cover this. All 4 tests only exercise `load_config_from_path()`. The auto-creation logic could silently break and nobody would know until a user runs the app for the first time.

At minimum, test:
- Calling `load_config(Some(nonexistent_path))` creates the file and returns a parseable config
- The created file contains the expected template content

### 4. Missing test: partial CLI overrides

`test_cli_overrides_config` passes `Some(...)` for all three override parameters. There's no test verifying that passing `None` for some parameters leaves the original config values untouched. This is the more common real-world case — a user passes `--project foo` but leaves region and model at their config file values.

### 5. Stale comment in test file

```rust
// We'll test config by calling the public API once it exists.
// For now, write the tests that define the expected behavior.
```

The API exists now. This comment is cargo-culted from the TDD red phase and was never cleaned up.

### 6. `src/main.rs` not modified

The issue explicitly lists "Modify: `src/main.rs`" under Files. The config module is exposed via `lib.rs` instead, which is fine for test access, but the config module is still completely disconnected from the actual application. `main.rs` still prints "Hello, world!" and ignores configuration entirely. If wiring is deferred to Task 8, that's acceptable — but the issue should be updated to reflect this, or the PR description should call it out explicitly.

---

### What's good

- Serde defaults with `#[serde(default = "...")]` — clean pattern.
- Error messages with `anyhow::Context` are helpful and point users to the right actions.
- The `CONFIG_TEMPLATE` with comments is a nice UX touch for first-run.
- Clippy and fmt both pass clean.

src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+        config.vertex.model = m.to_string();
+    }
+}
+
+/// Validate that the config has all required fields.
+pub fn validate(config: &AppConfig) -> Result<()> {
+    if config.vertex.project.is_empty() {
+        let path = default_config_path()
+            .map(|p| p.display().to_string())
+            .unwrap_or_else(|_| "~/.config/illustrious-manager/config.toml".to_string());
```
Fizzizist: `validate()` always calls `default_config_path()` to build its error message. If a custom config path was used via `load_config(Some(custom_path))`, the error will point the user to the wrong file. This function needs to accept the actual config path that was used and include it in the error message.

tests/config_test.rs
```diff
@@ -0,0 +1,85 @@
+use std::io::Write;
+
+use tempfile::NamedTempFile;
+
+// We'll test config by calling the public API once it exists.
+// For now, write the tests that define the expected behavior.
```
Fizzizist: This comment is a leftover from the TDD red phase. The API exists now — remove it.

tests/config_test.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[test]
+fn test_parse_valid_config() {
+    let toml_content = r#"
+[vertex]
+project = "my-project"
+region = "us-east5"
+model = "claude-sonnet-4-20250514"
+"#;
+    let mut tmp = NamedTempFile::new().unwrap();
```
Fizzizist: Policy violation: `.unwrap()` is forbidden under all circumstances per CLAUDE.md. Use `.expect("should create temp file")` and similar descriptive messages throughout the tests.

Fizzizist:
Thanks for the thorough review! I've addressed all the feedback:

## 1. ✅ Bug: `validate()` path parameter fixed

`validate()` now accepts `Option<&Path>` to use the actual config path in error messages:
```rust
pub fn validate(config: &AppConfig, config_path: Option<&Path>) -> Result<()>
```

When a custom path is used, the error message now correctly points to that path instead of always showing the default. Added a test (`test_validate_shows_correct_path_for_custom_config`) to verify this behavior.

## 2. ✅ Policy violation: All `.unwrap()` replaced with `.expect()`

All test assertions now use `.expect("descriptive message")` per CLAUDE.md policy. Every temp file creation, write operation, and config loading call has an informative expect message.

## 3. ✅ Missing test: `load_config()` auto-creation covered

Added `test_load_config_auto_creates_file()` which:
- Verifies the file doesn't exist before calling `load_config()`
- Calls `load_config()` with a nonexistent path
- Verifies the file was created with the correct template content
- Verifies the returned config has expected default values

## 4. ✅ Missing test: Partial CLI overrides

Added `test_partial_cli_overrides()` which verifies that passing `None` for some override parameters leaves those config values unchanged. This tests the real-world case where users override only `--project` while keeping their config file's region and model.

## 5. ✅ Stale comment removed

Deleted the outdated TDD red-phase comment from the test file.

## 6. Re: `src/main.rs` not modified

This is intentional per the implementation plan. Task 2 focuses solely on the configuration module. Task 8 ("CLI Wiring & Main Entry Point") will integrate the config system into `main.rs`. The plan explicitly shows `main.rs` printing "Hello, world!" until Task 8.

---

**Test results:** All 7 tests pass (was 4, added 3 new tests). `cargo fmt` and `cargo clippy` both clean.

PR #15, opened 2026-04-01
feat: add LlmBackend trait definition

### LINKED ISSUE #3: Task 3: LlmBackend trait definition
## Overview

Define the `LlmBackend` async trait that all LLM providers implement. This is the abstraction boundary between the agent core and any LLM service.

## Blocked by
- #1 (Project setup & core types)

## Files
- Create: `src/backend/mod.rs`
- Create: `src/backend/vertex.rs` (placeholder only)
- Modify: `src/main.rs`

## Acceptance Criteria

- [ ] `LlmBackend` trait with `async fn send_message(&self, messages: &[Message], config: &RequestConfig) -> Result<BoxStream<Result<StreamEvent>>>`
- [ ] Trait is `Send + Sync`
- [ ] `src/backend/mod.rs` re-exports `pub mod vertex`
- [ ] `src/backend/vertex.rs` exists as a placeholder
- [ ] `src/main.rs` declares `pub mod backend`
- [ ] `cargo build` succeeds

## Implementation Notes

See `docs/superpowers/plans/2026-03-27-phase1-core-cli.md` — Task 3 for full code.

Fizzizist:
## Summary
- Creates `src/backend/mod.rs` with the `LlmBackend` async trait (`Send + Sync`)
- Creates `src/backend/vertex.rs` as a placeholder for Task 4
- Exposes `pub mod backend` in `src/lib.rs`

Closes #3

## Test plan
- [ ] `cargo build` succeeds (verified)
- [ ] `cargo clippy` passes with no warnings (verified)

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -1,0 +1,19 @@
+pub mod vertex;
+
+use anyhow::Result;
+use async_trait::async_trait;
+
+use crate::types::{BoxStream, Message, RequestConfig, StreamEvent};
+
+/// Trait for LLM provider backends.
+///
+/// Implementations handle authentication, request formatting, and
+/// response streaming for a specific LLM provider.
+#[async_trait]
+pub trait LlmBackend: Send + Sync {
+    async fn send_message(
+        &self,
+        messages: &[Message],
+        config: &RequestConfig,
+    ) -> Result<BoxStream<Result<StreamEvent>>>;
+}
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -5,10 +5,6 @@ use async_trait::async_trait;
 
 use crate::types::{BoxStream, Message, RequestConfig, StreamEvent};
 
-/// Trait for LLM provider backends.
-///
-/// Implementations handle authentication, request formatting, and
-/// response streaming for a specific LLM provider.
 #[async_trait]
 pub trait LlmBackend: Send + Sync {
     async fn send_message(
@@ -75,6 +71,11 @@ mod tests {
         assert!(
             matches!(second, StreamEvent::Done),
             "expected Done, got {second:?}"
+        );
+
+        assert!(
+            stream.next().await.is_none(),
+            "stream should be exhausted after Done"
         );
     }
 }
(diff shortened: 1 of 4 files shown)


Fizzizist (commented):
## Code Review: feat: add LlmBackend trait definition

### Blocking Issue: Duplicate Module Declarations in `main.rs`

This PR introduces `pub mod backend` and `mod types` into `src/main.rs`, but `src/lib.rs` already declares both `pub mod backend` and `pub mod types`. This means the `backend` and `types` modules are compiled **twice** — once in the library crate and once in the binary crate. These produce two distinct, incompatible type hierarchies at the Rust level.

In `main.rs`:
```rust
pub mod backend;  // BAD — duplicates lib.rs declaration
mod types;        // BAD — duplicates lib.rs declaration
```

`main.rs` should not re-declare modules that already live in `lib.rs`. If `main.rs` needs to reference them, it should `use illustrious_manager::backend;` (or whatever the crate name resolves to). As it stands, `main.rs` doesn't even use either module, so these declarations should simply be removed.

**This must be fixed before merge.**

### Trait Definition

The `LlmBackend` trait itself looks correct. The signature, `Send + Sync` bounds, and `BoxStream` usage all match the spec. No complaints here — it does what it says.

### Tests

The test `mock_backend_streams_text_delta_then_done` exercises the trait via a mock implementation and verifies the stream contract (TextDelta followed by Done). It's a reasonable test for a trait definition task — it proves the trait is implementable and the stream mechanics work. Acceptable.

One nit: the test doesn't verify the stream is exhausted after the two events (i.e., that `stream.next().await` returns `None`). Minor, but would make the contract assertion complete.

### Doc Comments

The trait has a 3-line doc comment. The project policy says "keep code comments to a minimum — only comment in cases where something is unable to be gleaned from the code itself." A trait named `LlmBackend` with a method called `send_message` is self-documenting. The doc comment adds no information the code doesn't already convey. Remove it or trim to one line max.

### Verdict

Fix the duplicate module declarations in `main.rs` (remove `pub mod backend` and `mod types`). The rest is solid.

PR #16, opened 2026-04-01
feat: implement Vertex AI backend with SSE streaming

### LINKED ISSUE #4: Task 4: Vertex AI backend implementation
## Overview

Implement the Vertex AI + Claude backend: GCP authentication via ADC, HTTP request construction, SSE stream parsing, and the `LlmBackend` trait implementation.

## Blocked by
- #3 (LlmBackend trait definition)

## Files
- Modify: `src/backend/vertex.rs`
- Create: `tests/vertex_test.rs`
- Modify: `src/main.rs` (make `types` module public)

## Acceptance Criteria

- [ ] `VertexBackend::new(project, region)` initializes with ADC auth via `gcp_auth`
- [ ] Constructs the correct `streamRawPredict` endpoint URL
- [ ] Builds the Anthropic messages API request body with `anthropic_version`, `model`, `max_tokens`, `stream`, `messages`
- [ ] Implements `LlmBackend` trait — sends HTTP POST with bearer auth, streams response
- [ ] SSE parsing: `parse_sse_data()` is a public function that:
  - Converts `content_block_delta` events → `StreamEvent::TextDelta`
  - Converts `message_stop` events → `StreamEvent::Done`
  - Returns `None` for ignored types (`message_start`, `content_block_start`, `ping`, etc.)
- [ ] 5 unit tests for SSE parsing pass
- [ ] `cargo build` succeeds

## Implementation Notes

See `docs/superpowers/plans/2026-03-27-phase1-core-cli.md` — Task 4 for full code.

Fizzizist:
## Summary
- Implements `VertexBackend` with GCP ADC authentication via `gcp_auth`
- Constructs `streamRawPredict` endpoint URL and Anthropic messages API request body
- Parses SSE stream events: `content_block_delta` → `TextDelta`, `message_stop` → `Done`, others ignored
- Makes `types` module public for integration test access
- Adds 5 unit tests for `parse_sse_data` covering all event types

Closes #4

## Test plan
- [x] `cargo test --test vertex_test` — all 5 SSE parsing tests pass
- [x] `cargo test` — all 16 tests pass (no regressions)
- [x] `cargo build` — compiles successfully
- [x] `cargo clippy` — no warnings
- [x] `cargo fmt` — code formatted

### DIFF
--- a/src/backend/vertex.rs
+++ b/src/backend/vertex.rs
@@ -90,7 +90,10 @@ impl LlmBackend for VertexBackend {
 
         if !response.status().is_success() {
             let status = response.status();
-            let body = response.text().await.unwrap_or_default();
+            let body = response
+                .text()
+                .await
+                .unwrap_or_else(|_| String::from("<failed to read response body>"));
             anyhow::bail!("Vertex AI returned {}: {}", status, body);
         }
 
@@ -156,6 +159,32 @@ fn extract_sse_data(event_text: &str) -> Option<&str> {
         }
     }
     None
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn extract_sse_data_returns_json_after_data_prefix() {
+        let event = "event: content_block_delta\ndata: {\"type\":\"content_block_delta\"}";
+        assert_eq!(
+            extract_sse_data(event),
+            Some("{\"type\":\"content_block_delta\"}")
+        );
+    }
+
+    #[test]
+    fn extract_sse_data_returns_none_when_no_data_line() {
+        let event = "event: content_block_delta\n";
+        assert!(extract_sse_data(event).is_none());
+    }
+
+    #[test]
+    fn extract_sse_data_returns_first_data_line_when_multiple_present() {
+        let event = "data: first\ndata: second";
+        assert_eq!(extract_sse_data(event), Some("first"));
+    }
 }
 
 /// Parse an SSE data payload JSON into a StreamEvent.
--- a/src/backend/vertex.rs
+++ b/src/backend/vertex.rs
@@ -1,1 +1,177 @@
-// Vertex AI implementation — filled in Task 4
+use std::sync::Arc;
+
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use futures::StreamExt;
+use reqwest::Client;
+
+use super::LlmBackend;
+use crate::types::{BoxStream, Message, RequestConfig, StreamEvent};
+
+/// Vertex AI backend for Claude models.
+pub struct VertexBackend {
+    client: Client,
+    project: String,
+    region: String,
+    auth_manager: Arc<dyn gcp_auth::TokenProvider>,
+}
+
+impl VertexBackend {
+    /// Create a new VertexBackend using Application Default Credentials.
+    pub async fn new(project: String, region: String) -> Result<Self> {
+        let auth_manager = gcp_auth::provider().await.context(
+            "Failed to initialize GCP authentication. Run: gcloud auth application-default login",
+        )?;
+        Ok(Self {
+            client: Client::new(),
+            project,
+            region,
+            auth_manager,
+        })
+    }
+
+    fn endpoint(&self, model: &str) -> String {
+        format!(
+            "https://{region}-aiplatform.googleapis.com/v1/projects/{project}/locations/{region}/publishers/anthropic/models/{model}:streamRawPredict",
+            region = self.region,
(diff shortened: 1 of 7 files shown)


Fizzizist:
## Code Review: feat: implement Vertex AI backend with SSE streaming

### Build & Lint Status
- `cargo build` ✅
- `cargo clippy` ✅
- `cargo fmt --check` ✅

---

### 🔴 Required Changes

**1. Duplicate module declarations in `main.rs`**

`src/main.rs` adds:
```rust
pub mod backend;
pub mod config;
pub mod types;
```
But `src/lib.rs` already declares all three. This means the binary crate re-compiles its own copy of every module, completely separate from the library crate. The integration tests import from the library crate (`illustrious_manager::backend::vertex::parse_sse_data`), while the binary uses its own duplicate copies. This is wrong — it doubles compile time, creates two distinct type hierarchies, and will cause type mismatches the moment `main.rs` tries to use anything from the library. Remove these three lines from `main.rs`. When the binary needs to use library types, `use illustrious_manager::...`.

**2. Policy violation: `.unwrap_or_default()` on line 93 of `vertex.rs`**

```rust
let body = response.text().await.unwrap_or_default();
```

The project CLAUDE.md says *"The use of `.unwrap()` is forbidden under all circumstances."* While `.unwrap_or_default()` is technically not `.unwrap()`, it silently swallows the error from reading the response body. On this error path you're already bailing — you should propagate the actual body-read failure rather than silently replacing it with an empty string. Use `.unwrap_or_else(|_| String::from("<failed to read response body>"))` or chain with `?` and include it in the bail message.

**3. Missing test coverage for `extract_sse_data`**

`extract_sse_data` is a non-trivial parsing function with zero tests. The tests only cover `parse_sse_data`. Add tests for:
- Normal case: `"event: content_block_delta\ndata: {\"type\":\"content_block_delta\"}"` → extracts the JSON
- No data line → returns `None`
- Multiple lines with `data:` → verifies which one is returned

**4. No test for malformed JSON input**

`parse_sse_data` can receive garbage from the wire. There's no test that verifies the error path when fed invalid JSON. Add at least one test with non-JSON input and assert it returns `Err`.

---

### 🟡 Suggestions

**5. Hardcoded `max_tokens: 8192`**

`build_request_body` hardcodes `max_tokens: 8192`. This should come from `RequestConfig` or at minimum from the config file. As-is, there's no way for a user to control output length. Not a blocker for this PR if the config struct changes are planned for a later task, but flag it.

**6. `extract_sse_data` doesn't handle `data:` without trailing space**

Per the SSE spec, `data:value` (no space after colon) is valid. The current implementation requires `"data: "` with a space. The Anthropic API likely always sends the space, so this is low-risk, but it's a latent bug if any proxy or intermediary strips it.

**7. `Role` serialization in `build_request_body` is implicit**

```rust
serde_json::json!({
    "role": m.role,
    "content": m.content,
})
```

This works because `Role` has `#[serde(rename_all = "lowercase")]`, so `serde_json::json!` calls `serde_json::to_value` which respects the derive. It's correct but non-obvious — a reader might wonder if `Role::User` serializes as `"User"` or `"user"`. This is fine, just noting it.

---

### Test Coverage Assessment

The 5 unit tests for `parse_sse_data` cover the happy paths well (delta, stop, and three ignored event types). However:
- `extract_sse_data` has **zero** test coverage
- The error path (malformed JSON) has **zero** test coverage
- The SSE stream parsing logic in `send_message` (the `unfold` closure) has no unit tests — understandable since it requires mocking HTTP, but the two helper functions it delegates to should be thoroughly tested to compensate

---

### Summary

The core implementation is solid — SSE parsing, auth flow, endpoint construction, and the streaming unfold are all well-structured. The main issue is the duplicate `pub mod` declarations in `main.rs` which is a genuine bug. Fix that, plug the test gaps for `extract_sse_data` and the error path, and this is good to merge.

src/backend/vertex.rs
```diff
@@ -1 +1,206 @@
...
+            .map(|m| {
+                serde_json::json!({
+                    "role": m.role,
+                    "content": m.content,
+                })
+            })
+            .collect();
+
+        serde_json::json!({
+            "anthropic_version": "vertex-2023-10-16",
```
Fizzizist: What is this and why are we hard-coding it?

src/backend/vertex.rs
```diff
@@ -1 +1,206 @@
...
+            "model": model,
+            "max_tokens": 8192,
```
Fizzizist: This is really low and shouldn't be hard-coded either

PR #17, opened 2026-04-01
feat: implement agent core with streaming and history management

### LINKED ISSUE #5: Task 5: Agent core (history, streaming, error handling)
## Overview

Implement the Agent struct that owns conversation history and orchestrates the flow between user input and LLM responses. The agent is display-agnostic — it takes strings in and emits `AgentEvent`s out.

## Blocked by
- #3 (LlmBackend trait definition)

## Files
- Create: `src/agent.rs`
- Create: `tests/agent_test.rs`
- Modify: `src/main.rs`

## Acceptance Criteria

- [ ] `Agent::new(backend, config)` constructs an agent with empty history
- [ ] `Agent::send(input)` appends user message to history, calls backend, returns `BoxStream<AgentEvent>`
- [ ] Stream wraps `StreamEvent::TextDelta` → `AgentEvent::TokenReceived`
- [ ] Stream accumulates text and on `StreamEvent::Done` emits `AgentEvent::ResponseComplete(full_text)` and appends assistant message to history
- [ ] Backend errors in the stream emit `AgentEvent::Error`
- [ ] `Agent::history()` returns the message history
- [ ] 3 async tests pass:
  - Single message produces TokenReceived + ResponseComplete events
  - History accumulates across multiple send calls (4 messages after 2 exchanges)
  - Backend stream error produces Error event

## Implementation Notes

See `docs/superpowers/plans/2026-03-27-phase1-core-cli.md` — Task 5 for full code.

Tests use a `MockBackend` implementing `LlmBackend` that returns pre-configured `StreamEvent` sequences.

Fizzizist:
## Summary
- Implements `Agent` struct in `src/agent.rs` with conversation history, `send()` streaming, and `history()` getter
- Adds `pub mod agent` to `src/lib.rs`
- Adds `tests/agent_test.rs` with 3 async tests covering single message, history accumulation, and backend error handling

## Test Plan
- [x] `test_agent_single_message` — TokenReceived + ResponseComplete events emitted correctly
- [x] `test_agent_history_accumulates` — 4 messages after 2 exchanges
- [x] `test_agent_backend_error_emits_error_event` — Error event emitted on backend stream error
- [x] All existing tests still pass

Closes #5

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -1,12 +1,15 @@
+use std::sync::{Arc, Mutex};
+
 use anyhow::Result;
 use futures::StreamExt;
+use futures::channel::mpsc;
 
 use crate::backend::LlmBackend;
-use crate::types::*;
+use crate::types::{AgentEvent, BoxStream, Message, RequestConfig, Role, StreamEvent};
 
 pub struct Agent {
     backend: Box<dyn LlmBackend>,
-    history: Vec<Message>,
+    history: Arc<Mutex<Vec<Message>>>,
     config: RequestConfig,
 }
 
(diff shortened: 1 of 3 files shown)


Fizzizist (commented):
## Code Review: feat: implement agent core with streaming and history management

### Verdict: Request Changes

---

### Critical: `send()` destroys streaming semantics

The entire point of this architecture is streaming tokens to frontends in real-time. The `send()` method signature returns `BoxStream<AgentEvent>`, promising a stream — but it's a lie. The method eagerly consumes the **entire** backend stream (lines 40-58), buffers every event into a `Vec`, and then wraps that `Vec` in `futures::stream::iter()`.

```rust
// What this actually does:
// 1. Block until the LLM finishes generating the ENTIRE response
// 2. Collect all tokens into a Vec
// 3. Return a "stream" that just iterates over a Vec
```

When a real Vertex AI backend is behind this, the TUI will hang for potentially 10+ seconds with zero visible output, then dump the entire response at once. That's not streaming — that's batch processing with a streaming API stapled on top. The tests pass because the mock backend resolves immediately, which makes the buffering invisible.

The plan's implementation used unsafe raw pointers to maintain true streaming, which was correctly rejected — but the fix here was to give up on streaming entirely rather than implement it correctly. The right approach is to use a channel (`tokio::mpsc`) or a proper stream combinator. Something like:

```rust
pub async fn send(&mut self, input: String) -> Result<BoxStream<AgentEvent>> {
    self.history.push(Message {
        role: Role::User,
        content: input,
    });

    let backend_stream = self
        .backend
        .send_message(&self.history, &self.config)
        .await?;

    let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
    let mut history = std::mem::take(&mut self.history);

    tokio::spawn(async move {
        let mut accumulated = String::new();
        futures::pin_mut!(backend_stream);
        while let Some(result) = backend_stream.next().await {
            match result {
                Ok(StreamEvent::TextDelta(text)) => {
                    accumulated.push_str(&text);
                    let _ = tx.send(AgentEvent::TokenReceived(text));
                }
                Ok(StreamEvent::Done) => {
                    history.push(Message {
                        role: Role::Assistant,
                        content: accumulated.clone(),
                    });
                    let _ = tx.send(AgentEvent::ResponseComplete(accumulated));
                    break;
                }
                Err(e) => {
                    let _ = tx.send(AgentEvent::Error(e.to_string()));
                    break;
                }
            }
        }
    });

    // ... handle getting history back, or restructure ownership

    Ok(Box::pin(tokio_stream::wrappers::UnboundedReceiverStream::new(rx)))
}
```

The ownership of `history` in a spawned task is a real design challenge — it's why the plan resorted to unsafe. But buffering everything is not the answer. This needs a proper solution, even if it means rethinking the `Agent` struct's ownership model (e.g., `Arc<Mutex<Vec<Message>>>` for history, or requiring the caller to signal completion).

**This must be fixed before merge.** A streaming agent that doesn't actually stream is architecturally broken.

---

### Significant: Error path corrupts conversation history

When the backend stream errors (line 54), the user message has already been pushed to `self.history` (line 27-30), but no assistant message is appended. The history is now in a corrupted state — it has a dangling user message with no response. Subsequent calls to `send()` will forward this broken history to the backend, which will confuse the LLM (it'll see a user message it never responded to).

Either roll back the user message on error, or append a sentinel/error message to history so the conversation state stays consistent.

---

### Minor: Unused import

`use std::pin::Pin;` in `tests/agent_test.rs` line 1 is unused. Clippy catches this when run with `--tests`:

```
warning: unused import: `std::pin::Pin`
 --> tests/agent_test.rs:1:5
```

---

### Minor: Glob import

`use crate::types::*;` in `agent.rs` — prefer explicit imports so readers can tell at a glance where each type comes from. This module uses `Message`, `Role`, `RequestConfig`, `StreamEvent`, `AgentEvent`, `BoxStream` — list them explicitly.

---

### Missing from acceptance criteria

The issue lists `Modify: src/main.rs` as a file to touch, though it's not in the acceptance criteria checkboxes. If wiring the agent into main is deferred to a later task, that's fine — but it should be called out.

---

### Tests

The three tests cover the specified acceptance criteria. The mock backend approach is sound. However, the tests don't actually validate streaming behavior — they can't, because there's nothing to validate when everything is eagerly buffered. Once real streaming is implemented, consider a test that verifies tokens are yielded incrementally (e.g., using a mock backend with deliberate delays between events).

---

### Summary

The code compiles, passes clippy (minus the unused import in tests), and is formatted. The tests pass and cover the specified scenarios. But the core design is broken: `send()` pretends to stream while actually buffering everything. This will produce a terrible user experience when connected to a real backend. Fix the streaming, fix the error-path history corruption, and this is ready to merge.

Fizzizist:
Fixed in 1baec82.

**Critical (streaming):** Replaced the buffered Vec approach with `futures::channel::mpsc::unbounded()` + `tokio::spawn`. The backend stream is now consumed in a spawned task that forwards events in real-time. History is wrapped in `Arc<Mutex<Vec<Message>>>` so the task can append the assistant message on `Done` without unsafe pointer aliasing.

**Significant (error-path history):** On a mid-stream error, the spawned task calls `history.pop()` to remove the dangling user message before emitting `AgentEvent::Error`. The same rollback happens if `send_message` itself fails before returning a stream.

**Minor (unused import):** Removed `use std::pin::Pin;` from `agent_test.rs`.

**Minor (glob import):** Replaced `use crate::types::*` with explicit imports: `AgentEvent, BoxStream, Message, RequestConfig, Role, StreamEvent`.

src/agent.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub fn new(backend: Box<dyn LlmBackend>, config: RequestConfig) -> Self {
+        Self {
+            backend,
+            history: Arc::new(Mutex::new(Vec::new())),
+            config,
+        }
+    }
+
+    pub fn history(&self) -> Vec<Message> {
+        self.history.lock().expect("history mutex poisoned").clone()
```
Fizzizist: There's gotta be a safer way to handle these locks than to just panic

Fizzizist:
Fixed in 1be0c65. Replaced all `.expect("history mutex poisoned")` with a `lock()` helper that calls `.unwrap_or_else(|e| e.into_inner())` — recovers the `MutexGuard` from a poisoned mutex instead of panicking. No panics anywhere in the lock path now.

PR #19, opened 2026-04-02
feat: Ratatui TUI frontend with insta snapshot tests (Task 7)

### LINKED ISSUE #7: Task 7: Ratatui TUI frontend with insta snapshot tests
## Overview

Implement the Ratatui TUI frontend for interactive REPL mode. Two main areas: scrollable conversation display and text input. Dual event loop handles terminal events and agent events concurrently. Uses insta snapshot testing with `TestBackend` for visual regression tests.

## Blocked by
- #5 (Agent core)

## Files
- Modify: `src/frontend/tui.rs`
- Create: `tests/tui_test.rs`

### TUI Implementation
- [ ] `tui::run(agent, initial_prompt)` enters alternate screen, enables raw mode, runs the app, and restores terminal on exit
- [ ] Two-region layout: conversation area (scrollable) + input area (fixed 3 lines)
- [ ] `AppState::Input` — accepts typed characters, backspace, Enter to submit, Ctrl+C to quit
- [ ] `AppState::Streaming` — renders tokens as they arrive, Ctrl+C still works
- [ ] On submit: user message added to conversation, message sent to agent, state switches to Streaming
- [ ] On `ResponseComplete`: assistant message added to conversation, state switches back to Input
- [ ] On `Error`: error displayed in conversation, state switches back to Input
- [ ] Auto-scrolls to bottom as conversation grows
- [ ] Optional `initial_prompt` is sent immediately on startup

### Snapshot Tests (insta)
- [ ] `cargo install cargo-insta` documented as dev tooling requirement
- [ ] `insta` added as dev dependency in `Cargo.toml`
- [ ] Tests use `ratatui::backend::TestBackend` with a fixed terminal size (e.g. 80x24)
- [ ] Snapshot tests cover:
  - Empty app initial state (input area visible, conversation empty)
  - App with conversation entries rendered (user + assistant messages visible)
  - App in streaming state (partial response visible, "Streaming..." in input title)
- [ ] Snapshot files stored in `tests/snapshots/`
- [ ] All snapshot tests pass with `cargo test`

### Snapshot Test Pattern

```rust
use insta::assert_snapshot;
use ratatui::{backend::TestBackend, Terminal};

#[test]
fn test_tui_initial_state() {
    let backend = TestBackend::new(80, 24);
    let mut terminal = Terminal::new(backend).unwrap();
    
    // Render the app in its initial state
    let app = App::new();
    terminal.draw(|frame| {
        // render app widgets to frame
    }).unwrap();
    
    assert_snapshot!(terminal.backend());
}
```

Run `cargo insta review` to accept/update snapshots when the UI changes intentionally.

## Implementation Notes

See `docs/superpowers/plans/2026-03-27-phase1-core-cli.md` — Task 7 for implementation code.

The TUI rendering logic should be extractable into a testable `render` function that takes `&App` and a `Frame`, separate from the event loop, so that snapshot tests can render without running the async event loop.

Fizzizist:
## Summary

- Implements `src/frontend/tui.rs` with `App`, `AppState`, `ConversationEntry` types and a `run()` async function for the interactive REPL
- Extracts `render_app(app, frame)` as a standalone function so snapshot tests can render without the async event loop
- Adds `src/frontend/mod.rs` and exposes the `frontend` module in `src/lib.rs`
- Creates `tests/tui_test.rs` with 3 insta snapshot tests (initial state, conversation display, streaming state)
- Fixes pre-existing `agent_test.rs` compilation error: missing `max_tokens` field in `RequestConfig` initializers

## Test plan

- [x] `cargo test --test tui_test` — 3 snapshot tests pass (initial state, conversation, streaming)
- [x] `cargo test` — all 27 tests pass across all test suites
- [x] `cargo clippy` — no warnings
- [x] `cargo fmt` — formatting applied
- [x] Snapshots stored in `tests/snapshots/` and accepted via `cargo insta accept --all`

Closes #7


Fizzizist:
## Code Review — PR #19

### Critical: Duplicated Rendering Logic

`render_app()` (line 86–105) and the draw closure inside `run_app()` (line 139–170) duplicate the entire layout and rendering logic, but **they are not the same**. The `run_app` version adds auto-scroll and cursor positioning that `render_app` lacks.

This means:
1. **Your snapshot tests are testing code that doesn't actually run at runtime.** The extracted `render_app` is a stripped-down version of the real rendering, so the tests provide a false sense of coverage.
2. **Two copies of the same layout that must be kept in sync.** Change one, forget the other — guaranteed future bug.

**Required fix:** `run_app` should call `render_app` (or a shared inner function) instead of reimplementing the rendering. Auto-scroll and cursor positioning should be integrated into (or layered on top of) the extracted function, not duplicated beside it.

### Design Issue: `ConversationEntry.role` is a `String`, Not a `Role`

There is already a `Role` enum in `types.rs` with `User` and `Assistant` variants. `ConversationEntry` uses a bare `String` instead:

```rust
pub struct ConversationEntry {
    pub role: String,  // Why not Role?
    pub content: String,
}
```

This introduces stringly-typed logic like `if entry.role == "You"` (line 50) and the implicit convention that display labels ("You", "Assistant", "Error") are stored as the role. This is fragile — nothing prevents a typo like `"you"` from silently breaking color selection.

**Required fix:** Use the `Role` enum (or extend it if needed for `Error`). Map to display strings at render time, not at data construction time.

### Design Issue: Orphaned Spawned Tasks on Exit

`submit_message` (line 268–274) spawns a task to consume the agent stream:

```rust
tokio::spawn(async move {
    while let Some(event) = stream.next().await {
        if tx.send(event).await.is_err() {
            break;
        }
    }
});
```

When the user hits Ctrl+C during streaming, the main loop breaks and `run_app` returns, but the spawned task keeps running until the stream is exhausted or the channel drops. There is no cancellation token or `JoinHandle` tracked. For a long-running response, this means background work continues after the TUI has already torn down the terminal.

**Suggested fix:** Store the `JoinHandle` and abort it on exit, or use a `CancellationToken` from `tokio_util`.

### Minor Issues

- **No `Esc` key handling.** Ctrl+C is the only way to quit. `Esc` is a standard TUI exit key — worth adding to `AppState::Input`.
- **`input.len() as u16` (line 166)** will silently truncate if the input exceeds 65535 characters. Not critical, but `u16::try_from(...).unwrap_or(u16::MAX)` would be more defensive.

### Tests

The 3 snapshot tests cover the right states (initial, conversation, streaming) and the test structure using `TestBackend` + `render_app` is good. However, because `render_app` diverges from the actual runtime rendering (see critical issue above), **these tests do not verify the code that actually runs**. Fixing the duplication issue will make these tests actually meaningful.

The `agent_test.rs` fix (adding missing `max_tokens` field) is a straightforward compilation fix — no issues there.

### Summary

The core structure is sound — extracting `render_app` for testability is the right idea, the event loop architecture with `tokio::select!` is appropriate, and the snapshot tests target the right states. But the execution has a critical flaw: the extracted render function and the actual render code diverged, which undermines the entire testing strategy. Fix the duplication, use proper types for roles, and handle task cancellation.

Fizzizist:
merge trunk into your branch. There are conflicts

PR #22, opened 2026-04-03
fix: display user message immediately with 'thinking' indicator

### LINKED ISSUE #21: Sluggish input submission
In REPL, when I prompt, and hit Enter, it seems to takes a long time just for my message to appear in the conversation box. Seems like the UI is maybe waiting for the request before it moves on, which leads to a really bad UX.

## Acceptance Criteria

- When a user types a message and hits Enter, The message should _immediately_ appear in the conversation box.
- regression test

Fizzizist:
Fixes #21


Fizzizist:
## Code Review

### Overall Assessment
The fix correctly addresses the sluggish input submission issue by moving the `agent.send()` call into a background task, allowing the UI to update immediately. The change from `&mut self` to `&self` on `Agent::send` is sound because all state access goes through `Arc<Mutex<Vec<Message>>>` and the `LlmBackend` trait requires `Send + Sync`.

### Issues

**1. Weak regression test** (`tests/tui_test.rs:107-116`)
The test creates an `Agent`, an event channel, but never uses them. It manually pushes to `app.conversation` instead of calling `submit_message`. This doesn't test the async fire-and-forget behavior at all. A proper test would actually spawn the task and verify the UI state is correct before the backend responds.

```rust
// These are created but never used - what's the point?
let mut agent = Agent::new(Box::new(backend), config);
let (event_tx, mut event_rx) = mpsc::channel::<AgentEvent>(100);
```

**2. Compiler warnings introduced**
The PR changes `Agent::send` to take `&self`, but doesn't fix the now-unnecessary `mut` declarations in other tests:
- `tests/agent_test.rs:52`
- `tests/agent_test.rs:94` 
- `tests/agent_test.rs:132`
- `tests/tui_test.rs:107`

Fix these while you're here.

**3. Unused derives** (`src/frontend/tui.rs:15-16, 19-20`)
Adding `Debug, Clone, Copy, PartialEq, Eq` to `AppState` and `ConversationRole` is fine for testability, but they're not actually used anywhere in this PR. Either use them or remove them from this diff.

**4. Potential race: rapid-fire submissions** (`src/frontend/tui.rs:192-196`)
If a user spams Enter, you spawn multiple concurrent `send()` tasks. Since `Agent::send` pushes to the shared history immediately, this should serialize correctly via the mutex. But there's no UI feedback to prevent this - consider debouncing or disabling submission while streaming.

### Approval
Changes are functionally correct. Clean up the test warnings and this is good to merge.

Issue #23, opened 2026-04-03
z.ai backend


Fizzizist:
In addition to Vertex AI, let's support `z.ai` backend. 

Here are the docs for chat completions: https://docs.z.ai/api-reference/llm/chat-completion

PR #24, opened 2026-04-03
feat: add z.ai backend support (#23)

Fizzizist:
## Summary
- Implement z.ai (Zhipu AI) backend as an alternative to Vertex AI
- Support both backends simultaneously with runtime selection via config
- Follow TDD principles throughout implementation

## Changes
- **New `ZaiBackend` implementation** (`src/backend/zai.rs`)
  - Bearer token authentication using API key
  - OpenAI-style SSE streaming support
  - Request body builder for z.ai API format
  - SSE parser for z.ai response format

- **Updated config schema** (`src/config.rs`)
  - Add `backend` field to select backend (vertex/zai)
  - Add `[zai]` section with `api_key` and `model` fields
  - Updated validation to support both backends
  - New config template with both backends documented

- **Backend selection logic** (`src/main.rs`)
  - Dynamic backend creation based on config
  - Model selection based on active backend

## Configuration Example
```toml
# Select backend: "vertex" or "zai"
backend = "zai"

[vertex]
project = "my-gcp-project"
region = "us-east5"
model = "claude-sonnet-4-20250514"

[zai]
api_key = "your-zai-api-key"
model = "glm-5-turbo"
```

## Test Plan
- [x] All existing tests pass
- [x] New tests for z.ai backend construction
- [x] New tests for request body building
- [x] New tests for SSE parsing
- [x] Config validation tests for both backends
- [x] cargo fmt passes
- [x] cargo clippy passes

## Testing
```bash
cargo test                    # All tests pass
cargo fmt                     # Code formatted
cargo clippy                  # No warnings
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -21,7 +21,7 @@ model = "claude-sonnet-4-20250514"
 # Required: your z.ai API key
 api_key = ""
 # Model to use
-model = "glm-5-turbo"
+model = "glm-5.1"
 "#;
 
 #[derive(Debug, serde::Deserialize)]
@@ -54,7 +54,7 @@ fn default_backend() -> String {
 }
 
 fn default_zai_model() -> String {
-    "glm-5-turbo".to_string()
+    "glm-5.1".to_string()
 }
 
 fn default_region() -> String {
@@ -108,14 +108,26 @@ pub fn apply_overrides(
     region: Option<&str>,
     model: Option<&str>,
 ) {
-    if let Some(p) = project {
-        config.vertex.project = p.to_string();
-    }
-    if let Some(r) = region {
-        config.vertex.region = r.to_string();
-    }
-    if let Some(m) = model {
-        config.vertex.model = m.to_string();
+    match config.backend.as_str() {
+        "vertex" => {
+            if let Some(p) = project {
+                config.vertex.project = p.to_string();
+            }
+            if let Some(r) = region {
+                config.vertex.region = r.to_string();
+            }
+            if let Some(m) = model {
+                config.vertex.model = m.to_string();
+            }
+        }
+        "zai" => {
+            if let Some(m) = model
+                && let Some(ref mut zai) = config.zai
+            {
+                zai.model = m.to_string();
+            }
+        }
+        _ => {}
     }
 }
 
@@ -235,7 +247,7 @@ mod tests {
             },
             zai: Some(ZaiConfig {
                 api_key: "".to_string(),
-                model: "glm-5-turbo".to_string(),
+                model: "glm-5.1".to_string(),
             }),
         };
         let result = validate(&config, None);
@@ -254,7 +266,7 @@ mod tests {
             },
             zai: Some(ZaiConfig {
                 api_key: "test-key".to_string(),
-                model: "glm-5-turbo".to_string(),
+                model: "glm-5.1".to_string(),
             }),
         };
         let result = validate(&config, None);
--- a/src/config.rs
+++ b/src/config.rs
@@ -4,19 +4,33 @@ use anyhow::{Context, Result, bail};
 
 const DEFAULT_REGION: &str = "us-east5";
 const DEFAULT_MODEL: &str = "claude-sonnet-4-20250514";
+const DEFAULT_BACKEND: &str = "vertex";
 
-const CONFIG_TEMPLATE: &str = r#"[vertex]
+const CONFIG_TEMPLATE: &str = r#"# Which backend to use: "vertex" or "zai"
+backend = "vertex"
+
+[vertex]
 # Required: your GCP project ID
 project = ""
 # Vertex AI region
 region = "us-east5"
 # Model to use
 model = "claude-sonnet-4-20250514"
+
+[zai]
+# Required: your z.ai API key
+api_key = ""
+# Model to use
+model = "glm-5-turbo"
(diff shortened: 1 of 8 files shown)


src/config.rs
```diff
@@ -28,6 +42,21 @@ pub struct VertexConfig {
...
+    #[serde(default = "default_zai_model")]
+    pub model: String,
+}
+
+fn default_backend() -> String {
+    DEFAULT_BACKEND.to_string()
+}
+
+fn default_zai_model() -> String {
+    "glm-5-turbo".to_string()
```
Fizzizist: let's just default to `glm-5.1`

Fizzizist:
# Code Review: z.ai backend support

## Build / Lint / Tests

All green. 30 unit tests pass, 7 integration tests pass, clippy clean, fmt clean.

---

## Critical Issues

### 1. `apply_overrides` is broken for zai backend — `--model` flag silently overrides vertex, not zai

`src/config.rs:105-120` — The `apply_overrides` function **only ever touches `config.vertex.*` fields**, regardless of which backend is selected. If a user runs:

```bash
illustrious-manager --model glm-5 --single-shot "hello"
```

with `backend = "zai"` in their config, the `--model` flag sets `config.vertex.model` (not `config.zai.model`), which is then completely ignored at runtime. The `--project` and `--region` flags have the same problem — they're vertex-only concepts but the CLI doesn't enforce or warn about this.

This is a real usability bug. Either:
- `apply_overrides` needs to dispatch based on `config.backend`, or
- The CLI flags need to be made backend-specific, or
- At minimum, `apply_overrides` should warn/ignore when given vertex-specific flags with a non-vertex backend.

### 2. Duplicated backend-matching logic in `main.rs`

`src/main.rs:78-109` — The same `match app_config.backend.as_str()` block is written **twice**: once for backend construction, once for model selection. If someone adds a third backend, they have to update two separate match blocks in the same function. Extract a helper or struct that returns both the backend and the model in one shot.

### 3. `.expect("zai config should be validated")` on lines 91 and 103

`src/main.rs:91,103` — The CLAUDE.md policy says `.expect()` is only permitted when a panic is "logically impossible." The claim here is that validation already checked this. That's correct *today*, but this is a runtime bomb if someone refactors `validate()` or adds a code path that skips it. Use `anyhow::bail!()` or `.ok_or_else()` instead — the whole function already returns `Result`.

---

## Design Issues

### 4. `extract_sse_data` and the SSE unfold loop are copy-pasted from `vertex.rs`

`src/backend/zai.rs:86-133` vs `src/backend/vertex.rs:85-132` — The entire SSE parsing machinery (`extract_sse_data`, the `unfold` closure, the buffer management loop) is **character-for-character identical** between both backends. The only difference is `parse_sse_data`. Extract the SSE stream parsing into a shared utility that takes a parsing function as a parameter.

### 5. `endpoint()` returns a hardcoded string with allocation

`src/backend/zai.rs:28-30` — `fn endpoint(&self) -> String` allocates a new `String` on every call to return a compile-time constant. Should be a `const` or `&static str`. The `&self` receiver is unused.

---

## Test Coverage Gaps

### 6. No integration test file for z.ai backend

There's `tests/vertex_test.rs` with 6 tests for Vertex's `parse_sse_data`, but no corresponding `tests/zai_test.rs`. The z.ai `parse_sse_data` should have integration tests matching the Vertex pattern.

### 7. `parse_sse_data` has no test for choices with no delta content

`src/backend/zai.rs:151-169` — No test verifying behavior with a valid JSON payload that has `choices` but no `delta.content` (e.g., `{"choices":[{"delta":{}}]}`). This is a real scenario during streaming.

### 8. No test for `extract_sse_data` in zai module

The vertex backend tests `extract_sse_data` thoroughly (3 tests). The zai module has zero tests for its own `extract_sse_data`.

### 9. `config_test.rs` wasn't updated for z.ai config parsing

All 7 integration tests only exercise vertex config. No tests verifying that a TOML file with `backend = "zai"` and a `[zai]` section parses correctly, or that `backend` defaults to `"vertex"` when omitted.

---

## Minor Issues

### 10. `VertexConfig` is always required even when using zai

`src/config.rs:27-34` — Every z.ai user must have a `[vertex]` section in their config, even if they never use Vertex.

---

## Summary

| Severity | Issue |
|----------|-------|
| **Bug** | `apply_overrides` silently ignores CLI flags when using zai backend (1) |
| **Bug risk** | `.expect()` can panic at runtime; should use `Result` propagation (3) |
| **Maintainability** | Duplicated backend match blocks in main.rs (2) |
| **Maintainability** | Copy-pasted SSE parsing between backends (4) |
| **Coverage** | Missing integration tests for z.ai (6, 8, 9) |

**Verdict: Changes requested.** Issues 1 and 3 should be fixed before merge. 2, 4, and the test gaps are strongly recommended.

src/main.rs
```diff
@@ -73,18 +75,46 @@ async fn main() -> Result<()> {
...
+            let zai_backend = ZaiBackend::new(zai_config.api_key.clone())?;
+            BackendSelection {
+                backend: Box::new(zai_backend),
+                model: zai_config.model.clone(),
+            }
+        }
+        _ => {
+            anyhow::bail!("Invalid backend '{}'", app_config.backend);
+        }
+    };
```
Fizzizist: I don't think parsing the config should live in main

PR #25, opened 2026-04-03
fix: display user message immediately with "thinking" indicator (issue #21)

Fizzizist:
## Summary
Fixed the sluggish input submission bug where the user's message didn't appear in the conversation until after the backend responded.

## Changes
- Changed `Agent::send` to take `&self` instead of `&mut self` since it only accesses thread-safe state (Arc<Mutex<Vec<Message>>>)
- Modified TUI event loop to spawn message submission in background task
- Wrapped Agent in Arc to allow sharing across async tasks
- Added regression test that verifies user message appears immediately with "Streaming..." indicator

## Test plan
- [x] Regression test added that verifies submit_message completes in < 100ms even with a slow backend
- [x] All existing tests pass
- [x] Code formatted with `cargo fmt`
- [x] Code linted with `cargo clippy`

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -30,7 +30,7 @@ impl Agent {
         lock(&self.history).clone()
     }
 
-    pub async fn send(&mut self, input: String) -> Result<BoxStream<AgentEvent>> {
+    pub async fn send(&self, input: String) -> Result<BoxStream<AgentEvent>> {
         lock(&self.history).push(Message {
             role: Role::User,
             content: input,
--- a/src/frontend/tui.rs
+++ b/src/frontend/tui.rs
@@ -17,12 +17,15 @@ use tokio::task::JoinHandle;
 
 use crate::agent::Agent;
 use crate::types::AgentEvent;
+use std::sync::Arc;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum AppState {
     Input,
     Streaming,
 }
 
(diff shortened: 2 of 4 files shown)


Fizzizist:
## Code Review

### Overview
The fix correctly addresses issue #21. The root cause was that  was being awaited *outside* the tokio spawn, blocking the UI event loop. Moving it inside the spawn is the right solution.

### Correctness ✓

**Agent::send signature change ( → ):** Correct. The method only accesses  which is already . The  and  fields are only read, not mutated.

**TUI event loop change:** Correct. The user message is added to  and state is set to  *before* spawning the background task. The UI returns immediately without awaiting the backend.

### Issues

1. **Unused import in test:**  at line 71 of  is unused. Remove it.

### Nitpicks

2. ** visibility:** Changed from private to  solely for testing. Consider using   or keeping it private and testing through the public API ( with initial prompt). However, for a regression test of internal timing behavior, exposing it is acceptable.

3. **Clone noise:**  appears twice in . Could bind  and pass that, but this is minor and the compiler optimizes Arc cloning cheaply.

4. **Error handling in spawn:** The spawned task swallows errors if  fails after the channel closes. This is fine (channel closed means receiver dropped), but worth noting.

### Test Quality

The regression test is solid. Creating a  that delays 5 seconds and verifying  returns in < 100ms is exactly the right approach to prove the fix works.

---

**Verdict:** LGTM with the unused import cleanup. The core fix is correct.

PR #40, opened 2026-04-05
feat: implement Tool trait and ToolRegistry

Fizzizist:
## Summary
- Implements `Tool` trait with methods: `name()`, `description()`, `input_schema()`, `execute(input)`
- Defines `ToolResult` struct with `content` and `is_error` fields
- Defines `ToolDefinition` struct with `name`, `description`, and `input_schema` fields
- Implements `ToolRegistry` as `HashMap<String, Box<dyn Tool>>` with:
  - `new()` constructor
  - `register()` method that prevents duplicate registrations
  - `lookup()` method that returns `&dyn Tool` or error
  - `definitions()` method that returns `Vec<ToolDefinition>`

## Test plan
- [x] Register a mock tool, look it up by name
- [x] `definitions()` produces correct `Vec<ToolDefinition>`
- [x] Lookup of unregistered tool returns error
- [x] All 44 existing tests still pass

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,4 +2,5 @@ pub mod agent;
 pub mod backend;
 pub mod config;
 pub mod frontend;
+pub mod tools;
 pub mod types;
--- a/src/tools.rs
+++ b/src/tools.rs
@@ -1,0 +1,183 @@
+// Tool trait and registry for extensible tool system
+
+use serde_json::Value;
+use std::collections::HashMap;
+
+/// Result of tool execution
+#[derive(Debug, Clone, PartialEq)]
+pub struct ToolResult {
+    pub content: String,
+    pub is_error: bool,
+}
+
+/// Definition of a tool for discovery/registration
+#[derive(Debug, Clone, PartialEq)]
+pub struct ToolDefinition {
+    pub name: String,
+    pub description: String,
+    pub input_schema: Value,
+}
+
+/// Trait that all tools must implement
+pub trait Tool: Send + Sync {
+    fn name(&self) -> &str;
+    fn description(&self) -> &str;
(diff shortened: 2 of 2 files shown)


Fizzizist:
## Code Review

### Issues Found

**1. `ToolResult` and `ToolDefinition` lack serde derives**

These structs will almost certainly need to be serialized for the tool-use API. Adding `Serialize` and `Deserialize` derives now prevents breaking changes later.

**2. `input_schema()` returns `Value` by value, not reference**

```rust
fn input_schema(&self) -> Value;
```

This clones the entire schema on every call. Should return `&Value` to avoid unnecessary allocation.

**3. `ToolResult.content` is too limited for tool-use spec**

Claude's tool-use format supports structured content blocks (text, images, etc.). A single `String` won't be sufficient. Content should likely be an enum or Vec of content types.

**4. String error type throughout**

```rust
fn execute(&self, input: Value) -> Result<ToolResult, String>;
```

String errors are the lowest common denominator. Define a proper error type so callers can actually handle errors programmatically.

**5. `ToolRegistry` is not thread-safe**

The registry wraps a `HashMap` with no synchronization. If this is used across threads (likely in an async agent context), you'll need interior mutability. Add a note or wrap it in `RwLock`.

**6. Missing test coverage**

No tests for:
- Duplicate tool registration (the error case you added)
- Execution with malformed input
- Execution returning errors

### Nitpicks

- `definitions()` clones everything - cache or return refs if performance matters
- Consider deriving `Eq` on `ToolResult` if you're deriving `PartialEq`

### Summary

The core design is sound, but it's missing serde support where it's clearly needed, and the `ToolResult` type isn't expressive enough for the actual tool-use content block format. Fix those before merging.

PR #41, opened 2026-04-05
feat: Implement SandboxPolicy for path validation

### LINKED ISSUE #28: Implement SandboxPolicy for path validation
## Context
Part of the tool-use & file I/O feature. Shared security boundary for all file-touching tools.

## Spec
`docs/superpowers/specs/2026-04-04-tool-use-file-io-design.md` — Section: Sandbox Policy

## Work
- Implement `SandboxPolicy` struct holding the allowed root path
- `validate_path(path) -> Result<PathBuf>`: resolves symlinks, canonicalizes, rejects paths outside sandbox
- Default sandbox root: current working directory
- Accept sandbox root from config

## Tests
- Path inside sandbox: allowed
- Path outside sandbox: rejected
- Symlink escaping sandbox: rejected
- Relative paths resolved correctly
- No sandbox configured: defaults to cwd

## Blocked by
- Implement Tool trait and ToolRegistry

Fizzizist:
closes #28 

## Summary
Implements `SandboxPolicy` struct with `validate_path` method to ensure all file operations stay within a configured sandbox root directory. This prevents directory traversal attacks and symlink-based sandbox escapes.

## Implementation Details
- **SandboxPolicy struct**: Holds the allowed root path
- **validate_path()**: Resolves symlinks, canonicalizes paths, and rejects paths outside sandbox
- **Path normalization**: Handles relative paths by resolving `.` and `..` components
- **Default sandbox**: Uses current working directory when not configured

## Test Coverage
All tests follow TDD (written first, then implemented):
- ✅ Paths inside sandbox are allowed
- ✅ Paths outside sandbox are rejected  
- ✅ Symlinks escaping sandbox are rejected
- ✅ Relative paths are resolved correctly
- ✅ Default sandbox uses current working directory

## Test Plan
- [x] All tests passing (cargo test)
- [x] No clippy warnings (cargo clippy)
- [x] Code formatted (cargo fmt)

### DIFF
--- a/src/tools.rs
+++ b/src/tools.rs
@@ -384,15 +384,12 @@ impl SandboxPolicy {
     /// Only validates existing paths. Non-existent paths are rejected to prevent
     /// symlink-based sandbox escapes via parent directory symlinks.
     pub fn validate_path(&self, path: &Path) -> Result<PathBuf, SandboxError> {
-        // Resolve the path to its absolute form
         let absolute = if path.is_absolute() {
             path.to_path_buf()
         } else {
             self.root.join(path)
         };
 
-        // Reject non-existent paths for security reasons
-        // Manual normalization can't detect symlink-based escapes in parent directories
         if !absolute.try_exists().map_err(|_| {
             SandboxError::InvalidPath(format!("Cannot check path existence: {:?}", path))
         })? {
@@ -402,17 +399,14 @@ impl SandboxPolicy {
             )));
         }
 
-        // Get the canonicalized sandbox root for comparison
         let sandbox_canonical = self.root.canonicalize().map_err(|_| {
             SandboxError::InvalidPath(format!("Cannot canonicalize sandbox root: {:?}", self.root))
         })?;
 
-        // Canonicalize to resolve any symlinks in the path
         let canonical = absolute.canonicalize().map_err(|_| {
             SandboxError::InvalidPath(format!("Cannot canonicalize path: {:?}", path))
         })?;
 
-        // On Windows, ensure case-insensitive comparison
         #[cfg(windows)]
         let is_within = {
             use std::path::Component;
@@ -458,8 +452,6 @@ impl SandboxPolicy {
 
 impl Default for SandboxPolicy {
     fn default() -> Self {
-        // Use current directory if available, otherwise use "/"
-        // This avoids panic while still providing a usable default
         let root = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("/"));
         Self::new(&root)
     }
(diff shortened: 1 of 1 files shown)


Fizzizist:
## Code Review

### Critical Issues

**1. Default impl can panic**
```rust
impl Default for SandboxPolicy {
    fn default() -> Self {
        Self::new(&std::env::current_dir().expect("Failed to get current directory"))
    }
}
```
If the current directory is deleted, `current_dir()` returns an `Err`. Using `expect()` crashes the program. Either document this as a panic condition or handle it gracefully. The policy states no `.unwrap()` — this is effectively the same.

### Security Concerns

**2. Non-existent paths with symlinked parents**

For non-existent paths, `normalize_path` does manual component cleanup without resolving symlinks. If `/sandbox/subdir` is a symlink to `/etc`, validating `subdir/passwd` (where passwd doesn't exist) returns success because the normalized path `/sandbox/subdir/passwd` starts with the canonical sandbox root. But creating that file would write to `/etc/passwd`.

The manual normalization is fundamentally insecure — it can't detect symlink-based escapes without resolving the actual path components on disk.

**3. TOCTOU vulnerability**

Even for existing paths, there's a race condition. Between `canonicalize()` and the caller's actual use of the validated path, a symlink swap could redirect the operation outside the sandbox. This is inherent to the design — the validation function returns a validated path but can't guarantee the filesystem state remains consistent.

**4. Windows path comparison vulnerability**

On Windows, `starts_with` on paths can fail due to case insensitivity and path normalization differences. `C:\sandbox` and `c:\sandbox` might not compare equal, and `starts_with` doesn't handle this. Use `pathdiff` or properly normalize both paths for comparison.

### Design Issues

**5. Inconsistent error handling**

```rust
let sandbox_canonical = self.root.canonicalize().map_err(|_| {
    SandboxError::InvalidPath(format!("Cannot canonicalize sandbox root: {:?}", self.root))
})?;
```

The error says "Cannot canonicalize path" but references `self.root`, not the input `path`. Misleading when debugging.

### Minor Issues

**6. Redundant `cur_dir()` call in test**

```rust
let current_dir = std::env::current_dir().expect("Failed to get cwd");
let result = sandbox.validate_path(&current_dir);
```

You already have the current dir from `SandboxPolicy::default()`. Just validate `Path::new(".")` instead.

**7. Test uses `unwrap().parent().unwrap()`**

```rust
let outside_path = temp_dir.path().parent().unwrap().join("outside.txt");
```
If temp_dir has no parent (unlikely but possible), this panics. Less fragile to use `temp_dir.path().ancestors().nth(1)`.

### What's Good

- Tests are comprehensive and cover TDD approach
- Clean separation of existing vs non-existent path handling
- Good use of `tempfile` for test isolation
- Error type is well-structured with Display implementation

### Verdict

The core security issue (#2) makes this unsuitable for production use without addressing the non-existent path case. Either reject non-existent paths entirely, or accept that symlink-based escapes are possible for new files and document this limitation.

Fizzizist (commented):
there are way too many code comments in this PR

PR #42, opened 2026-04-05
feat: add [tools] configuration section

### LINKED ISSUE #29: Add [tools] configuration section
## Context
Part of the tool-use & file I/O feature. All tool-related settings live here.

## Spec
`docs/superpowers/specs/2026-04-04-tool-use-file-io-design.md` — Section: Configuration

## Work
- Add `ToolsConfig` struct to `config.rs`:
  - `confirmation`: enum `Always | WriteOnly | Never` (default: `WriteOnly`)
  - `sandbox_root`: `String` (default: `"."`)
  - `max_tool_iterations`: `u32` (default: `25`)
  - `bash_allowlist`: `Vec<String>` (default: `["cat", "ls", "grep", "find", "head", "tail", "wc", "tree"]`)
  - `bash_denylist`: `Vec<String>` (default: `["rm", "wget", "sudo", "chmod", "chown"]`)
- Add `tools` field to `AppConfig` with serde defaults
- Resolve `sandbox_root` to absolute path at startup

## Tests
- Default config has correct tool defaults
- Custom config overrides all fields
- Sandbox root resolved to absolute path
- Missing `[tools]` section uses all defaults

## Blocked by
- Implement Tool trait and ToolRegistry

Fizzizist:
## Summary
- Adds `ConfirmationMode` enum (`Always | WriteOnly | Never`, default `WriteOnly`)
- Adds `ToolsConfig` struct with `confirmation`, `sandbox_root`, `max_tool_iterations`, `bash_allowlist`, `bash_denylist` fields and correct serde defaults
- Adds `tools: ToolsConfig` field to `AppConfig` with `#[serde(default)]` so missing `[tools]` section uses all defaults
- Adds `resolve_sandbox_root` function that canonicalizes relative paths to absolute at startup

## Test plan
- [x] `default_tools_config_has_correct_defaults` — verifies all default values
- [x] `custom_tools_config_overrides_all_fields` — verifies TOML overrides work for all fields
- [x] `missing_tools_section_uses_all_defaults` — verifies AppConfig deserialization without `[tools]` section
- [x] `sandbox_root_resolved_to_absolute_path` — verifies `resolve_sandbox_root(".")` returns an absolute path
- [x] All 55 existing tests continue to pass

Closes #29

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,6 +2,80 @@ use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result, bail};
 
+#[derive(Debug, Default, PartialEq, serde::Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub enum ConfirmationMode {
+    Always,
+    #[default]
+    WriteOnly,
+    Never,
+}
+
+fn default_confirmation() -> ConfirmationMode {
+    ConfirmationMode::default()
+}
+
+fn default_sandbox_root() -> String {
+    ".".to_string()
+}
+
+fn default_max_tool_iterations() -> u32 {
+    25
+}
+
+fn default_bash_allowlist() -> Vec<String> {
+    ["cat", "ls", "grep", "find", "head", "tail", "wc", "tree"]
+        .iter()
+        .map(|s| s.to_string())
+        .collect()
+}
+
+fn default_bash_denylist() -> Vec<String> {
+    ["rm", "wget", "sudo", "chmod", "chown"]
+        .iter()
+        .map(|s| s.to_string())
+        .collect()
+}
+
+#[derive(Debug, serde::Deserialize)]
+pub struct ToolsConfig {
+    #[serde(default = "default_confirmation")]
+    pub confirmation: ConfirmationMode,
+    #[serde(default = "default_sandbox_root")]
+    pub sandbox_root: String,
+    #[serde(default = "default_max_tool_iterations")]
+    pub max_tool_iterations: u32,
+    #[serde(default = "default_bash_allowlist")]
+    pub bash_allowlist: Vec<String>,
+    #[serde(default = "default_bash_denylist")]
+    pub bash_denylist: Vec<String>,
+}
+
+impl Default for ToolsConfig {
+    fn default() -> Self {
+        ToolsConfig {
+            confirmation: default_confirmation(),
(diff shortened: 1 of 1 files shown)


Fizzizist (commented):
## Code Review

Build: clean. Clippy: clean. The diff is small and focused. A few issues:

### 1. `resolve_sandbox_root` doesn't validate absolute paths

The absolute path branch just returns the `PathBuf` as-is without checking that it exists or is a directory. The relative path branch calls `.canonicalize()` which will fail if the path doesn't exist — good. But an absolute path like `/nonexistent/garbage` sails through unchecked.

Either canonicalize both branches, or document why you deliberately skip validation for absolute paths. Right now it looks like an oversight.

### 2. `resolve_sandbox_root` is dead code

Nothing calls it outside of tests. Not `load_config`, not `load_config_from_path`, not `main`. The sandbox root gets deserialized as a raw string and stays that way. If the plan is to wire it up later, fine, but right now it's a public function that nobody uses, and the PR description says "Resolve sandbox_root to absolute path at startup" — that's not happening.

### 3. Manual `Default` impl is redundant

The `Default` impl for `ToolsConfig` just calls the same `default_*` functions that serde already uses. You could derive `Default` on `ToolsConfig` and use `#[serde(default)]` on the struct itself instead of per-field `#[serde(default = "...")]`. That said, if the intent is to keep serde defaults and `Default` decoupled for future divergence, the current approach is defensible — just know you're paying a maintenance cost for it.

### 4. The `default_bash_allowlist` / `default_bash_denylist` allocations

```rust
["cat", "ls", "grep", "find", "head", "tail", "wc", "tree"]
    .iter()
    .map(|s| s.to_string())
    .collect()
```

Nit: `.map(|s| (*s).into())` or `.map(ToString::to_string)` are more idiomatic than `.map(|s| s.to_string())` on `&&str`. Not a blocker.

### 5. `CONFIG_TEMPLATE` not updated

The template written to disk on first run doesn't include a `[tools]` section. A user who reads their auto-generated config will have no idea this section exists. If you want discoverability, add it to the template (commented out with defaults). If you don't care about discoverability, at least note that in the PR.

### Verdict

The config struct and deserialization are correct and well-tested. The main issue is #2 — the PR claims sandbox resolution at startup but doesn't deliver it. Either wire `resolve_sandbox_root` into the config loading path or drop the claim. Everything else is minor.

PR #43, opened 2026-04-05
feat: extend Vertex AI backend to parse tool-use SSE events

### LINKED ISSUE #33: Extend Vertex AI backend to parse tool-use SSE events
## Context
Part of the tool-use & file I/O feature. The backend must parse tool-use blocks from Claude's streaming response.

## Spec
`docs/superpowers/specs/2026-04-04-tool-use-file-io-design.md` — Sections: Backend Changes, Type System Changes (StreamEvent)

## Work
- Include `tools` array in Vertex AI request body when `RequestConfig.tools` is non-empty
- Extend SSE parser to handle:
  - `content_block_start` with type `tool_use` → `StreamEvent::ToolUseStart { id, name }`
  - `content_block_delta` with type `input_json_delta` → `StreamEvent::ToolUseDelta(chunk)`
  - `content_block_stop` → `StreamEvent::ToolUseDone`
- Add `tools: Vec<ToolDefinition>` to `RequestConfig`
- z.ai backend: return error if tool definitions are present (tool use not supported)

## Tests
- Parse `content_block_start` tool_use → `ToolUseStart`
- Parse `input_json_delta` chunks → `ToolUseDelta`
- Parse `content_block_stop` → `ToolUseDone`
- Mixed text + tool_use blocks in single response
- z.ai backend rejects requests with tools

## Blocked by
- Extend type system for tool-use content blocks (#26)

Fizzizist:
## Summary
- Moves `ToolDefinition` to `types.rs` to avoid circular dependency when `RequestConfig` references it
- Adds `tools: Vec<ToolDefinition>` to `RequestConfig`; includes tools array in Vertex AI request body when non-empty
- Extends `parse_sse_data` to emit `ToolUseStart`, `ToolUseDelta`, and `ToolUseDone` from `content_block_start` (tool_use), `input_json_delta`, and `content_block_stop` SSE events respectively
- z.ai backend returns an error immediately if tools are present (unsupported)

## Test plan
- [x] `parse_sse_data` content_block_start with tool_use type → ToolUseStart
- [x] `parse_sse_data` input_json_delta → ToolUseDelta
- [x] `parse_sse_data` content_block_stop → ToolUseDone
- [x] text_delta path still works after refactor
- [x] content_block_start with text type → None
- [x] `build_request_body` includes tools array when non-empty
- [x] `build_request_body` omits tools key when empty
- [x] z.ai `send_message` returns error when tools are present
- [x] All 142 existing tests pass

Closes #33

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -86,6 +86,7 @@ mod tests {
         let config = RequestConfig {
             model: "test-model".to_string(),
             max_tokens: 1024,
+            tools: vec![],
         };
 
         let mut stream = backend
--- a/src/backend/sse.rs
+++ b/src/backend/sse.rs
@@ -13,18 +13,19 @@ pub fn extract_sse_data(event_text: &str) -> Option<&str> {
     None
 }
 
-pub fn create_sse_event_stream<S, B, E>(
+pub fn create_sse_event_stream<S, B, E, P>(
     byte_stream: S,
-    parser: fn(&str) -> Result<Option<StreamEvent>>,
+    parser: P,
 ) -> BoxStream<Result<StreamEvent>>
 where
     S: Stream<Item = Result<B, E>> + Unpin + Send + 'static,
     B: AsRef<[u8]>,
     E: std::fmt::Display + 'static,
+    P: FnMut(&str) -> Result<Option<StreamEvent>> + Send + 'static,
 {
     let event_stream = unfold(
-        (byte_stream, String::new()),
-        move |(mut byte_stream, mut buffer)| async move {
+        (byte_stream, String::new(), parser),
+        move |(mut byte_stream, mut buffer, mut parser)| async move {
             loop {
                 if let Some(pos) = buffer.find("\n\n") {
                     let event_text = buffer[..pos].to_string();
@@ -33,10 +34,10 @@ where
                     if let Some(data) = extract_sse_data(&event_text) {
                         match parser(data) {
                             Ok(Some(event)) => {
-                                return Some((Ok(event), (byte_stream, buffer)));
+                                return Some((Ok(event), (byte_stream, buffer, parser)));
                             }
                             Ok(None) => continue,
-                            Err(e) => return Some((Err(e), (byte_stream, buffer))),
+                            Err(e) => return Some((Err(e), (byte_stream, buffer, parser))),
                         }
                     }
                     continue;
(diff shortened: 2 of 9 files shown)


Fizzizist (commented):
## Code Review

### Bug: `content_block_stop` fires `ToolUseDone` unconditionally

`src/backend/vertex.rs:274` — This is the most significant issue. `content_block_stop` is emitted by the API for **every** content block, not just tool_use blocks. If the model responds with a text block (index 0) followed by a tool_use block (index 1), both blocks emit `content_block_stop`. Your parser currently maps all of them to `ToolUseDone`, which means the consumer will receive a spurious `ToolUseDone` after every text block too.

You need to track block types by index, or at minimum check whether you're currently inside a tool_use block before emitting `ToolUseDone`. The test at line 204 only sends a single `content_block_stop` in isolation — it doesn't catch the interleaved text+tool scenario.

### Silent data loss on malformed payloads

`src/backend/vertex.rs:248-255` — When `content_block.type` is `"tool_use"` but `id` or `name` are missing, `unwrap_or("")` silently returns empty strings. This produces a `ToolUseStart { id: "", name: "" }` that will quietly propagate through the agent layer. If the API sends a malformed tool_use event, you want to know about it — return an error instead of swallowing it. Same applies to `partial_json` at line 264.

### `.expect()` in non-test code

`src/backend/vertex.rs:109` — Policy says `.expect()` is only allowed when panic is "logically impossible." Serializing a `Vec<ToolDefinition>` where `ToolDefinition` is `#[derive(Serialize)]` with `String` and `Value` fields — yes, this is effectively impossible to fail. But `serde_json::to_value` returns a `Result` for a reason. A one-liner `map_err` to `anyhow` costs nothing and keeps the no-panic guarantee airtight:

```rust
body["tools"] = serde_json::to_value(&config.tools)
    .context("Failed to serialize tool definitions")?;
```

The function already returns `serde_json::Value` — oh wait, it doesn't return `Result`. That's the real problem. `build_request_body` returns a bare `serde_json::Value`, so you can't propagate the error. You have two options: change the return type to `Result<serde_json::Value>`, or accept the `.expect()` as-is since the panic truly is unreachable. I'd prefer the former since the function is already doing serialization work that could theoretically fail, but this is minor.

### The `content_block_stop` bug is a blocker

The other items are improvements. The `content_block_stop` issue will cause real misbehavior in any response that mixes text and tool_use blocks, which is the common case for agentic usage. Fix that before merging.

PR #45, opened 2026-04-05
feat: implement WriteFile tool (#31)

### LINKED ISSUE #31: Implement WriteFile tool
## Context
Part of the tool-use & file I/O feature. Creates or overwrites files within the sandbox.

## Spec
`docs/superpowers/specs/2026-04-04-tool-use-file-io-design.md` — Section: Tools > WriteFile

## Work
- Implement `WriteFile` struct implementing `Tool` trait
- Create or overwrite file at given path
- Create parent directories if needed
- Validate path against SandboxPolicy before execution
- Return confirmation with bytes written
- Input schema: `{ path: string, content: string }`

## Tests
- Write new file: created with correct content
- Overwrite existing file: content replaced
- Parent dirs created automatically
- Path outside sandbox: rejected
- Symlink escaping sandbox: rejected

## Blocked by
- Implement SandboxPolicy for path validation (#28)
- Add [tools] configuration section (#29)

Fizzizist:
## Summary

- Adds `WriteFileTool` implementing the `Tool` trait with `write_file` name and `{ path: string, content: string }` input schema
- Adds `SandboxPolicy::validate_write_path` for validating paths that may not yet exist — walks up to the nearest existing ancestor, canonicalizes it, and verifies it stays within the sandbox root
- `WriteFileTool::execute` validates the path, creates parent directories with `create_dir_all`, writes the file, and returns a confirmation with bytes written

## Test plan

- [x] Write new file: created with correct content
- [x] Overwrite existing file: content replaced
- [x] Parent dirs created automatically
- [x] Path outside sandbox: rejected
- [x] Symlink escaping sandbox: rejected
- [x] Result reports bytes written
- [x] `validate_write_path` allows non-existent paths inside sandbox
- [x] `validate_write_path` rejects non-existent paths outside sandbox
- [x] `validate_write_path` rejects symlink parents escaping sandbox
- [x] `validate_write_path` rejects `..` traversal

Closes #31


Fizzizist (commented):
## Code Review

### 1. Security Bug: Missing Windows case-insensitive path comparison

`validate_path` (line 404-426) has a `#[cfg(windows)]` block that does case-insensitive path component comparison. `validate_write_path` uses bare `starts_with()` on lines 505 and 515, which is case-sensitive on all platforms. This means on Windows, a path like `C:\Sandbox\...\FILE.TXT` compared against `c:\sandbox` would bypass the sandbox check.

Either extract the "is path within sandbox" logic into a shared helper, or duplicate the Windows block into `validate_write_path`. A shared helper is obviously the right call — you already have two methods doing the same containment check with divergent implementations. That's a bug waiting to happen (and it already has).

### 2. Policy violation: `.unwrap()` usage

CLAUDE.md says: *"The use of `.unwrap()` is forbidden under **all** circumstances."*

This PR introduces 7 new `.unwrap()` calls. Yes, they're in test code. The policy doesn't carve out an exception for tests. Use `.expect("descriptive message")` like every other test in this file does.

Line 685 is the most egregious — `result.unwrap()` right after an `assert!(result.is_ok(), ...)`. Just use `let path = result.expect("...");`.

The `.to_str().unwrap()` calls on path conversions (lines 851, 872, 895, etc.) should also be `.to_str().expect("temp path should be valid UTF-8")`.

### 3. Excessive doc comments on `validate_write_path`

Lines 441-448 are a 7-line doc comment block including a `# Security` section. The project policy is "comments should be kept to an absolute minimum." The method name and signature already communicate what it does. The `validate_path` method right above it has the same style of doc comment, so this is a pre-existing problem, but this PR shouldn't make it worse. Trim it or remove it.

### 4. No TOCTOU note or mitigation

`validate_write_path` resolves the path and then `execute()` writes to it in a separate step. Between validation and `create_dir_all` + `fs::write`, the filesystem state can change (e.g., a symlink could be created at one of the pending path components). This is a classic TOCTOU race. At minimum, acknowledge this in a comment. Better: re-validate after `create_dir_all` that the resolved path is still within the sandbox before writing.

### 5. Minor: redundant sandbox check in pending loop

Line 513-519 checks `result.starts_with(&sandbox_canonical)` inside the loop after joining each pending component. Since you already rejected `..` components and the ancestor is within the sandbox, the only way a `join` of a normal component escapes the sandbox is... it doesn't. The check is dead code. Either remove it or explain what attack it's defending against.

### Verdict

The Windows path comparison omission is a real security bug that needs fixing before merge. The `.unwrap()` violations are a policy issue. Fix those two and this is in good shape — the core algorithm is sound and the test coverage is solid.

Fizzizist (commented):
merge trunk into your branch and move the write file tool into its own file

Issue #30, opened 2026-04-04
Implement Bash tool


Fizzizist:
## Context
Part of the tool-use & file I/O feature. Primary tool for reading files, exploring directories, and general shell operations.

## Spec
`docs/superpowers/specs/2026-04-04-tool-use-file-io-design.md` — Section: Tools > Bash

## Work
- Implement `Bash` struct implementing `Tool` trait
- Execute commands via subprocess with `cwd` set to sandbox root
- Command filtering: parse first token of command (and each pipe segment)
  - Allowlisted commands: auto-execute regardless of confirmation policy
  - Denylisted commands: always reject with error message
  - Everything else: follows confirmation policy
- Input schema: `{ command: string }`
- Return stdout/stderr as tool result

## Tests
- Allowlisted command executes without confirmation
- Denylisted command is rejected
- Unlisted command follows confirmation policy
- Piped commands: each segment checked against lists
- Command runs with correct cwd (sandbox root)
- Command failure returns is_error=true with stderr

## Blocked by
- Implement SandboxPolicy for path validation (#28)
- Add [tools] configuration section (#29)

PR #46, opened 2026-04-05
feat: implement BashTool (#30)

Fizzizist:
## Summary

- Adds `BashTool` struct implementing the `Tool` trait in `src/tools.rs`
- Commands are classified per pipe segment (first token of each `|`-separated segment): allowlisted → auto-execute, denylisted → reject, unlisted → follow `ConfirmationMode`
- Executes via `sh -c <command>` with `current_dir` set to `sandbox_root`
- Returns stdout/stderr combined as `ToolResult`; failed commands set `is_error = true`
- Confirmation callback injected via `Box<dyn Fn(&str) -> bool + Send + Sync>` for testability

## Test plan

- [x] Allowlisted command executes without calling confirm_fn (even under `Always` policy)
- [x] Denylisted command is rejected with `is_error = true`
- [x] Unlisted command with `Never` policy executes without confirmation
- [x] Unlisted command with `Always` policy and approved confirm_fn executes
- [x] Unlisted command with `Always` policy and denied confirm_fn returns error
- [x] Unlisted command with `WriteOnly` policy and denied confirm_fn returns error
- [x] Piped command with any denylisted segment is rejected
- [x] Piped command with all allowlisted segments skips confirmation
- [x] Command runs with sandbox root as cwd
- [x] Failing command returns `is_error = true` with stderr content
- [x] Missing `command` field returns `InvalidInput` error

### DIFF
--- a/src/backend/zai.rs
+++ b/src/backend/zai.rs
@@ -4,7 +4,7 @@ use reqwest::Client;
 
 use super::LlmBackend;
 use super::sse::create_sse_event_stream;
-use crate::types::{BoxStream, Message, RequestConfig, StreamEvent};
+use crate::types::{BoxStream, ContentBlock, Message, RequestConfig, StreamEvent};
 
 const ENDPOINT: &str = "https://api.z.ai/api/coding/paas/v4/chat/completions";
 
@@ -33,10 +33,17 @@ impl ZaiBackend {
         let messages_json: Vec<serde_json::Value> = messages
             .iter()
             .map(|m| {
-                serde_json::json!({
-                    "role": m.role,
-                    "content": m.content,
-                })
+                let content: Vec<serde_json::Value> = m
+                    .content
+                    .iter()
+                    .map(|block| match block {
+                        ContentBlock::Text(text) => {
+                            serde_json::json!({"type": "text", "text": text})
+                        }
+                        other => serde_json::to_value(other).unwrap_or(serde_json::Value::Null),
+                    })
+                    .collect();
+                serde_json::json!({"role": m.role, "content": content})
             })
             .collect();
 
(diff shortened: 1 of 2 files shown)


Fizzizist:
## Code Review: BashTool implementation

### BLOCKING: Denylist is trivially bypassable — do not merge

The pipe-segment parsing in `first_tokens_of_pipe_segments` only splits on `|`. Every other shell operator is invisible to it. An LLM (or anyone) can bypass the denylist with:

```
echo hello && rm -rf /        # denylist sees: ["echo"]
echo hello; rm -rf /          # denylist sees: ["echo"]
echo $(rm -rf /)              # denylist sees: ["echo"]
echo `rm -rf /`               # denylist sees: ["echo"]
echo hello
rm -rf /                      # newline-separated, denylist sees: ["echo"]
```

All of these classify as "all allowlisted" and execute without confirmation. The denylist provides a false sense of security while being trivially defeated. This is the kind of bug that makes people trust a guardrail that isn't there.

You need to either:
1. Parse `&&`, `||`, `;`, `$()`, backticks, and newlines as segment boundaries too, **or**
2. Treat the denylist as a substring/token scan across the entire raw command string, **or**
3. Accept that shell command classification is fundamentally hard and don't pretend the denylist is a security boundary — make confirmation the primary gate and document the denylist as best-effort.

Option 3 is the honest one. Options 1-2 will still have edge cases (heredocs, variable expansion, aliases, etc.), but at least they'd catch the obvious stuff.

### `WriteOnly` mode is a no-op

`WriteOnly` is handled identically to `Always` at line 678:
```rust
ConfirmationMode::Always | ConfirmationMode::WriteOnly
```

There is zero logic to classify a command as read vs. write. `pwd` triggers confirmation under `WriteOnly`. Either implement actual read/write classification or remove the variant — having it in the enum and treating it as `Always` is misleading.

### Use `HashSet` instead of `Vec` for allowlist/denylist

You're doing `self.denylist.iter().any(|d| d == token)` and `self.allowlist.iter().any(|a| a == token)` — O(n) linear scans on every pipe segment of every command. These should be `HashSet<String>` with `.contains()`. It's not a correctness bug, but it's sloppy for something that runs on every tool invocation.

### Minor: both stdout and stderr empty produces empty `ContentBlock::Text`

When a command succeeds silently (no output), the match falls through to `(true, _) => stderr.into_owned()` which is `""`. You get a `ToolResult` with a single empty string text block. Not a crash, but the caller probably shouldn't have to deal with that — consider returning an empty `content` vec or a sentinel like "(no output)".

### Tests

Test coverage is solid for the happy paths and the stated behavior. The test names are clear, the `panic!` trick in `confirm_fn` to assert non-invocation is good. But there are no tests for the bypass vectors above, which is the real problem — the tests validate the behavior as implemented, not the behavior as intended.

### Verdict

The core execution plumbing is fine. The denylist bypass is a showstopper. Fix the security model before merge.

Fizzizist (commented):
During user test, got the following error:
```
│You:                                                                                                                                                          │
│  read the `test.txt` file and say the correct word                                                                                                           │
│                                                                                                                                                              │
│Error:                                                                                                                                                        │
│  z.ai returned 400 Bad Request: {"error":{"code":"1214","message":"messages[0].content[0].type: cannot be empty"}}
```

Issue #1, opened 2026-03-27
Task 1: Project setup & core types


Fizzizist:
## Overview

Set up the Rust project with all Phase 1 dependencies and define the shared core types used across all layers.

## Blocked by
None — this is the first task.

## Files
- Modify: `Cargo.toml`
- Create: `src/types.rs`
- Modify: `src/main.rs`

## Acceptance Criteria

- [ ] `Cargo.toml` has all Phase 1 dependencies: `tokio`, `reqwest` (with `stream` feature), `gcp_auth`, `clap` (with `derive` feature), `serde`/`serde_json`, `toml`, `ratatui`, `crossterm`, `futures`, `dirs`, `anyhow`, `async-trait`, `pin-project-lite`
- [ ] Dev dependencies include `tempfile` and `insta`
- [ ] `src/types.rs` defines: `Message`, `Role`, `StreamEvent`, `AgentEvent`, `RequestConfig`, `BoxStream<T>` type alias
- [ ] `Role` serializes as lowercase (`"user"`, `"assistant"`)
- [ ] `src/main.rs` declares the `types` module
- [ ] `cargo build` succeeds

## Implementation Notes

See `docs/superpowers/plans/2026-03-27-phase1-core-cli.md` — Task 1 for full code.

PR #47, opened 2026-04-05
feat: implement agent tool loop with confirmation support

### LINKED ISSUE #34: Implement agent tool loop with confirmation support
## Context
Part of the tool-use & file I/O feature. The core agentic loop — the agent streams a response, executes tool calls, injects results, and re-sends until the model responds with pure text.

## Spec
`docs/superpowers/specs/2026-04-04-tool-use-file-io-design.md` — Section: Agent Tool Loop

## Work
- Refactor `agent.send()` into a loop:
  1. Send messages to backend
  2. Stream response, accumulate text deltas and tool-use blocks
  3. On completion: if no tool use → `ResponseComplete`. If tool use → execute tools → append results → re-send
  4. Loop until pure text response
- Accept `ToolRegistry` in Agent constructor
- Accept confirmation response channel (`mpsc`) from frontend
- Confirmation flow:
  - Check tool against confirmation policy (from config)
  - If confirmation required: emit `ToolConfirmationRequired`, wait for response on channel
  - If approved/auto-allowed: execute, emit `ToolResult`
  - If rejected: emit `ToolResult` with error telling model user declined
- Max iterations circuit breaker (from config, default 25)
- Emit `AgentEvent::ToolUseReceived` and `AgentEvent::ToolResult` for frontend display

## Tests
- Mock backend returns tool-use events → agent executes mock tool and re-sends with results
- Confirmation required → `ToolConfirmationRequired` emitted, agent pauses
- Confirmation approved → tool executes
- Confirmation rejected → error result sent to model
- Max iterations exceeded → error emitted
- Multiple tool calls in single response → all executed
- Pure text response → no tool loop, direct `ResponseComplete`

## Blocked by
- Implement Tool trait and ToolRegistry (#27)
- Add [tools] configuration section (#29)
- Extend Vertex AI backend to parse tool-use SSE events (#33)

Fizzizist:
## Summary

- Refactors `Agent::send()` into a full agentic loop that handles tool-use SSE events
- Adds builder methods `with_tools(ToolRegistry)`, `with_tool_config(&ToolsConfig)`, `with_confirmation_channel(rx)` to `Agent`
- Implements confirmation flow (Always/WriteOnly/Never) with `ToolConfirmationRequired` events and an mpsc channel from the frontend
- Circuit breaker: emits `AgentEvent::Error` when `max_tool_iterations` is exceeded
- Adds `ConfirmationResponse` enum to `types.rs`, `Clone` to `ConfirmationMode`, and `is_write_tool()` default method to `Tool` trait

## Test plan

- [ ] `pure_text_response_emits_response_complete_without_tool_loop` — no tool loop for plain text
- [ ] `tool_use_response_with_never_confirmation_executes_tool_and_re_sends` — tool executed, results sent back, loop continues to completion
- [ ] `always_confirmation_emits_tool_confirmation_required` — confirmation event emitted
- [ ] `confirmation_approved_executes_tool` — approved tool runs and returns output
- [ ] `confirmation_rejected_sends_error_result_to_model` — rejected tool sends error result, loop continues
- [ ] `max_iterations_exceeded_emits_error` — circuit breaker fires
- [ ] `multiple_tool_calls_in_single_response_all_executed` — all tool calls in one turn executed
- [ ] `write_only_confirmation_requires_confirmation_for_write_tools_only` — WriteOnly mode only gates write tools

Closes #34

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -5,12 +5,21 @@ use futures::StreamExt;
 use futures::channel::mpsc;
 
 use crate::backend::LlmBackend;
-use crate::types::{AgentEvent, BoxStream, Message, RequestConfig, Role, StreamEvent};
+use crate::config::{ConfirmationMode, ToolsConfig};
+use crate::tools::ToolRegistry;
+use crate::types::{
+    AgentEvent, BoxStream, ConfirmationResponse, ContentBlock, Message, RequestConfig, Role,
+    StreamEvent,
+};
 
 pub struct Agent {
-    backend: Box<dyn LlmBackend>,
+    backend: Arc<dyn LlmBackend>,
     history: Arc<Mutex<Vec<Message>>>,
     config: RequestConfig,
+    tools: Arc<ToolRegistry>,
+    max_tool_iterations: u32,
+    confirmation_mode: ConfirmationMode,
+    confirmation_rx: Option<Arc<tokio::sync::Mutex<mpsc::UnboundedReceiver<ConfirmationResponse>>>>,
 }
 
 fn lock(m: &Mutex<Vec<Message>>) -> std::sync::MutexGuard<'_, Vec<Message>> {
@@ -20,10 +29,33 @@ fn lock(m: &Mutex<Vec<Message>>) -> std::sync::MutexGuard<'_, Vec<Message>> {
 impl Agent {
     pub fn new(backend: Box<dyn LlmBackend>, config: RequestConfig) -> Self {
         Self {
-            backend,
+            backend: Arc::from(backend),
             history: Arc::new(Mutex::new(Vec::new())),
             config,
+            tools: Arc::new(ToolRegistry::new()),
+            max_tool_iterations: 25,
+            confirmation_mode: ConfirmationMode::WriteOnly,
+            confirmation_rx: None,
         }
+    }
+
+    pub fn with_tools(mut self, tools: ToolRegistry) -> Self {
+        self.tools = Arc::new(tools);
+        self
+    }
+
+    pub fn with_tool_config(mut self, tool_config: &ToolsConfig) -> Self {
+        self.max_tool_iterations = tool_config.max_tool_iterations;
+        self.confirmation_mode = tool_config.confirmation.clone();
+        self
+    }
+
+    pub fn with_confirmation_channel(
+        mut self,
+        rx: mpsc::UnboundedReceiver<ConfirmationResponse>,
+    ) -> Self {
+        self.confirmation_rx = Some(Arc::new(tokio::sync::Mutex::new(rx)));
+        self
     }
 
     pub fn history(&self) -> Vec<Message> {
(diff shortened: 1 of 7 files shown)


Fizzizist (commented):
## Code Review — PR #47

### Architecture

The overall design is sound: builder pattern on `Agent`, agentic loop with circuit breaker, confirmation flow via mpsc channel. The event-driven approach with `AgentEvent` variants for tool lifecycle is clean and composable. Good.

### Issues

**1. `unwrap_or` on JSON parse silently swallows malformed tool input (agent.rs:159-160)**

```rust
let input: serde_json::Value =
    serde_json::from_str(&input_json).unwrap_or(serde_json::Value::Null);
```

If the LLM returns garbage JSON, you silently pass `Null` to the tool and let it fail in some unrelated way downstream. This should send an error `ToolResult` back to the model so it can retry with valid JSON, or at minimum log the parse failure. Silently eating bad input is how you get confusing debugging sessions.

**2. The `send()` method is doing way too much (agent.rs:65-248)**

This is a ~180-line closure inside a `tokio::spawn`. It handles:
- Stream consumption and event accumulation
- Tool call parsing state machine
- Confirmation flow with channel reads
- Tool execution and result collection
- History management
- Circuit breaker logic

This is begging to be broken into at least two or three private methods. The tool execution loop (lines 158-234) could easily be `execute_tool_calls()` or similar. Right now it's a wall of code that's hard to reason about.

**3. History corruption on mid-loop backend errors (agent.rs:92-98)**

```rust
if iterations == 1 {
    lock(&history_arc).pop();
}
```

If the backend fails on iteration 2+, you leave stale assistant/tool-result messages in history from previous iterations but don't clean up. The next `send()` call will have a history with dangling tool results and no final assistant response. This will confuse the model or cause API errors.

Either pop everything you added since the user message, or document why partial history is acceptable.

**4. Tool calls processed sequentially when they could be parallel (agent.rs:158)**

The `for (id, name, input_json) in tool_calls` loop executes tools one at a time. When the model returns multiple tool calls in a single response (which your test `multiple_tool_calls_in_single_response_all_executed` covers), these are independent and could run concurrently. Not a blocker, but worth a TODO.

**5. Confirmation channel is shared across calls (agent.rs:22)**

```rust
confirmation_rx: Option<Arc<tokio::sync::Mutex<mpsc::UnboundedReceiver<ConfirmationResponse>>>>
```

If `send()` is called multiple times concurrently (which `&self` allows), the confirmation channel is shared via `Arc<tokio::sync::Mutex<...>>`. Two concurrent tool loops could race on the same receiver, stealing each other's confirmation responses. The `ToolConfirmationRequired` event doesn't carry enough context for the frontend to know *which* pending request it's confirming.

This is a latent bug. If concurrent `send()` isn't intended, make `send` take `&mut self`. If it is intended, you need per-call confirmation channels.

**6. `lock()` helper uses `unwrap_or_else(|e| e.into_inner())` (agent.rs:25-27)**

This silently recovers from poisoned mutexes. If a thread panicked while holding the lock, the history is in an unknown state and you just keep going. This is a deliberate design choice, but it's exactly the kind of thing that deserves a comment explaining *why* — especially given the project policy says "only comment where something cannot be gleaned from the code." A poisoned mutex recovery is non-obvious intent.

### Tests

The test suite is solid. Good coverage of the happy paths and edge cases. The `SequencedBackend` mock is well-designed — clean way to simulate multi-turn conversations.

One gap: no test for backend error on iteration > 1 (the partial history cleanup issue from #3 above). You test first-iteration errors implicitly, but the divergent behavior at `if iterations == 1` is untested.

### Nits

- `tool_calls: Vec<(String, String, String)>` — a named struct or type alias would make line 158's destructuring less cryptic. `(id, name, input_json)` is fine when you read the destructure, but `Vec<(String, String, String)>` tells you nothing at the declaration site.

### Verdict

The core logic works and the tests prove it. Issues #3 and #5 are real bugs that will bite you. #1 is sloppy. The rest is structural feedback for maintainability. Fix #1, #3, and #5 before merging.

PR #48, opened 2026-04-06
feat(stdout): handle tool events and wire confirmation channel

### LINKED ISSUE #36: Stdout: inline tool output and non-TTY early exit
## Context
Part of the tool-use & file I/O feature. The stdout frontend displays tool activity inline and handles non-TTY mode safely.

## Spec
`docs/superpowers/specs/2026-04-04-tool-use-file-io-design.md` — Section: Frontend Changes > Stdout

## Work
- Handle new `AgentEvent` variants:
  - `ToolUseReceived`: print tool name + input summary
  - `ToolResult`: print result inline
  - `ToolConfirmationRequired`:
    - If stdin is TTY: prompt user, send decision through confirmation channel
    - If stdin is not TTY: reject tool, print error to stderr, exit early
- Wire up confirmation `mpsc` channel between stdout frontend and agent

## Tests
- TTY mode: tool calls and results printed inline
- Non-TTY mode: confirmation-required tool triggers early exit with stderr message
- Allowlisted tools execute without prompting in both modes

## Blocked by
- Implement agent tool loop with confirmation support (#34)

Fizzizist:
## Summary

- `ToolUseReceived`: prints `[tool: {name}] {input_json}` inline
- `ToolResult`: prints `[result from {name}]: {content}` or `[error from {name}]: {content}` inline
- `ToolConfirmationRequired`: prompts user in TTY mode with `[y/N]`; in non-TTY mode sends `Rejected`, prints to stderr, and exits with error
- Wires `mpsc` confirmation channel between stdout frontend and agent in `main.rs` for single-shot mode

## Test plan

- [x] `tool_use_received_prints_tool_name_and_input` — tool name and JSON key appear in output
- [x] `tool_result_success_prints_result_content` — `[result from ...]` label and content appear
- [x] `tool_result_error_prints_error_label` — `[error from ...]` label appears for `is_error: true`
- [x] `non_tty_confirmation_required_sends_rejection_and_errors` — `Rejected` sent and error returned
- [x] `tty_confirmation_with_y_sends_approved` — `Approved` sent when user inputs `y`
- [x] `tty_confirmation_with_n_sends_rejected` — `Rejected` sent when user inputs `n`
- [x] All 125 existing tests continue to pass

Closes #36


Fizzizist (commented):
## Code Review

Builds clean, clippy clean, all 159 tests pass. The implementation is straightforward and correct. A few items:

### Issues

1. **Blocking `read_line` in an async context** (`stdout.rs:66`). `stdin.read_line()` is a synchronous, blocking call sitting inside an `async fn` on the tokio runtime. In single-shot mode with one tool confirmation this is fine in practice, but it's technically blocking the executor thread. If this ever gets called in a context with concurrent tasks, it'll stall everything. Worth a `// blocking call, acceptable for single-shot CLI` comment or, better, wrapping in `tokio::task::spawn_blocking` / `tokio::io::BufReader`. Not a blocker for this PR, but don't let it rot.

2. **Silent `let _ =` on `unbounded_send`** (`stdout.rs:61, 72`). If the receiver has been dropped (agent task panicked or exited early), these sends silently fail and the agent hangs waiting for a confirmation that will never arrive. The agent side does handle a closed channel (line 245 in agent.rs returns `false`), so this is fine for the rejection path on line 61, but on line 72 (the happy path TTY confirmation), a dropped receiver means the user typed `y` and nothing happens. Consider at least logging when the send fails.

3. **`eprint!` without flush** (`stdout.rs:64`). The prompt `Allow tool '{}' with input {}? [y/N]` is written via `eprint!` (no newline). Stderr is typically line-buffered when connected to a terminal, meaning this prompt may not appear before the blocking `read_line` call. You need an explicit `std::io::stderr().flush()?` after the `eprint!`.

### Nits

- The `use std::io::IsTerminal;` at line 12 inside the function body is unusual. Move it to the top-level imports with everything else.
- `try_recv` in tests (`stdout.rs:256, 283, 310`) — `try_recv` can return `Ok(None)` if the value hasn't been sent yet (race condition in theory). Since these are synchronous test helpers running after `run_with_writer` completes, it's safe here, but `try_next()` from `StreamExt` with a proper assertion would be more idiomatic for an mpsc channel.

### Verdict

The stderr flush bug (#3) should be fixed before merge — users will stare at a blank terminal wondering what's happening. The rest are minor. The overall design (generic `R: BufRead` for testability, TTY detection, channel wiring) is solid.

PR #49, opened 2026-04-06
feat: TUI inline tool display and confirmation prompts

### LINKED ISSUE #35: TUI: inline tool display and confirmation prompts
## Context
Part of the tool-use & file I/O feature. The TUI displays tool calls and results inline and handles confirmation prompts.

## Spec
`docs/superpowers/specs/2026-04-04-tool-use-file-io-design.md` — Section: Frontend Changes > TUI

## Work
- Handle new `AgentEvent` variants:
  - `ToolUseReceived`: render tool name + input summary inline in conversation
  - `ToolResult`: render result below the invocation (truncated if long)
  - `ToolConfirmationRequired`: switch to `AppState::ToolConfirmation`
- Add `AppState::ToolConfirmation` state:
  - Display pending tool call details
  - Accept `y`/`n` input
  - Send decision through confirmation channel to agent
- Wire up confirmation `mpsc` channel between TUI and agent

## Tests (insta snapshots)
- Tool call renders inline with name and input summary
- Tool result renders below invocation
- Long tool result is truncated
- Confirmation prompt state renders correctly

## Blocked by
- Implement agent tool loop with confirmation support (#34)

Fizzizist:
## Summary
- Adds `AppState::ToolConfirmation` to pause streaming and prompt user with `y`/`n` for tool execution
- Adds `ConversationRole::ToolUse` (Cyan) and `ConversationRole::ToolResult` (Yellow) for inline display of tool calls and results
- Wires a `futures::channel::mpsc` confirmation channel between TUI and `agent.send()`
- Truncates tool results longer than 200 chars with `...[truncated]`
- Handles all three new `AgentEvent` variants: `ToolUseReceived`, `ToolResult`, `ToolConfirmationRequired`

## Test plan
- [ ] `test_tui_tool_call_renders_inline` — tool name + input appear in conversation
- [ ] `test_tui_tool_result_renders_below_invocation` — result appears after tool use entry
- [ ] `test_tui_long_tool_result_is_truncated` — 500-char output truncated to 200 + `...[truncated]`
- [ ] `test_tui_confirmation_prompt_state_renders` — input box shows `Confirm tool execution [y/n]` with tool details
- [ ] All existing snapshot tests still pass

Closes #35


Fizzizist (commented):
## Code Review for PR #49

### Bug: `truncate_if_needed` slices on byte offset, not char boundary

`tui.rs:120`:
```rust
format!("{}...[truncated]", &content[..TOOL_RESULT_TRUNCATE_LEN])
```

`content[..200]` indexes by **bytes**, not characters. Any multi-byte UTF-8 content (emoji, CJK, accented chars, etc.) will panic at runtime. This is a hard crash waiting to happen.

Fix: use `content.chars().take(TOOL_RESULT_TRUNCATE_LEN)` or `content.char_indices()` to find the correct split point. Or if you actually want byte-level truncation, use `content.floor_char_boundary(TOOL_RESULT_TRUNCATE_LEN)` (nightly) or write a safe equivalent.

### Bug: `truncate_if_needed` allocates unconditionally

The else branch does `content.to_string()` — a full clone even when nothing is truncated. `conversation_lines()` then iterates the returned `String` by line. This means every single conversation entry gets cloned on every render frame. For a conversation with tool results, that's gratuitous allocation pressure on every 16ms tick.

Return `Cow<'_, str>` or just pass the content directly and only allocate when truncation actually happens.

### Design: Massive code duplication in `ToolConfirmation` handler

Lines 262-270 and 280-288 are **identical** — both the `'y'` and `'n'` branches push the exact same `ConversationEntry` and only differ in the `ConfirmationResponse` variant sent. Extract the common conversation push out of the match arms.

Also, this duplicates what `AgentEvent::ToolUseReceived` already does in the Streaming handler (lines 328-338). When a tool *requires* confirmation, `ToolConfirmationRequired` fires instead of `ToolUseReceived`, so the confirmation handler manually recreates the same conversation entry. But when a tool *doesn't* require confirmation, the `ToolUseReceived` handler does it. You've got the same display logic in two places with subtly different formatting paths. This will diverge.

### Design: `.state.clone()` on every loop iteration

`tui.rs:212`:
```rust
match &app.state.clone() {
```

`AppState` now contains a `String` and a `serde_json::Value`. You're deep-cloning both every 16ms just to match on the state. The original code used `Copy` for good reason. Now that `AppState` can't be `Copy`, borrow it or restructure the match to avoid cloning.

The reason this was done is because the `ToolConfirmation` arm then clones `name` and `input` *again* on lines 252-253. So you're doing two clones per frame. Just restructure the ownership.

### Nit: Dead code in `ToolConfirmation` title match

```rust
AppState::ToolConfirmation { name, .. } => {
    let _ = name;
    "Confirm tool execution [y/n]"
}
```

`let _ = name;` is a smell. The comment says "avoid lifetime issues with the format string." If you want the tool name in the title, construct the string properly. If you don't, don't destructure it. Either `AppState::ToolConfirmation { .. }` to ignore it, or actually use it.

### Nit: `let _ = tx.unbounded_send(...)` silently drops send errors

Lines 272, 289 — if the confirmation channel is broken, the TUI transitions to `Streaming` anyway and will hang waiting for events that will never come. At minimum log or surface these failures.

### Summary

The UTF-8 slicing bug is a showstopper — fix that before merge. The per-frame cloning and duplication are less urgent but should be addressed.

PR #57, opened 2026-04-06
add ctrl+u/ctrl+d scrolling to conversation output box

### LINKED ISSUE #52: output box scrolling
The user should be able to scroll through the output box results using the vim style `ctrl+u` and `ctrl+d` motions

Fizzizist:
## Summary
- Adds `scroll_offset` and `viewport_height` fields to `App`
- Adds `scroll_up`/`scroll_down` methods that adjust offset by half-page
- `render_app` now uses the offset to position the conversation view; scrolling is clamped to valid range
- `Ctrl+U` (scroll up) and `Ctrl+D` (scroll down) handled in all app states (Input, ToolConfirmation, Streaming)

Closes #52

## Test plan
- [ ] Unit tests for `scroll_up`, `scroll_down`, underflow clamping
- [ ] Snapshot test verifying that a non-zero `scroll_offset` shifts the visible content upward
- [ ] All existing snapshot tests unaffected (default `scroll_offset=0` preserves auto-scroll-to-bottom behavior)


Fizzizist (commented):
## Code Review

### 1. PR description is inaccurate

> `Ctrl+U` (scroll up) and `Ctrl+D` (scroll down) handled in all app states (Input, ToolConfirmation, Streaming)

This is false. The `Streaming` state event handler (the `tokio::select!` branch at line ~406) only handles `Ctrl+C`. Ctrl+U/D are silently swallowed during streaming. Either fix the implementation or fix the description.

### 2. `render_app` mutates state — bad contract

You changed the signature from `&App` to `&mut App` so you could write `viewport_height` during rendering. A render function should not have side effects. This is the kind of thing that causes subtle bugs when someone later calls `render_app` from a test or a different context and doesn't expect it to mutate state.

Compute `viewport_height` when the terminal is resized or at the top of the event loop, not inside the render path. The frame size is available from `terminal.size()` outside of the draw closure.

### 3. `scroll_offset` grows without bound

`scroll_up` does `saturating_add` with no upper clamp. Yes, `render_app` clamps with `.min(auto_scroll)`, but the raw `scroll_offset` field can grow to `u16::MAX` if the user keeps mashing Ctrl+U. This means after scrolling up a lot, the user has to press Ctrl+D dozens of times to get back to the bottom even though they were already at the top of the conversation. Clamp in `scroll_up` or normalize after clamping in render.

### 4. No scroll reset on new content

When the user is scrolled up and new content arrives (streaming tokens, new messages), `scroll_offset` is never reset to 0. The user stays stuck at some stale position while the conversation moves underneath them. This is especially bad during streaming — the user will think the response is frozen.

At minimum, reset `scroll_offset = 0` when new content arrives (in the `AgentEvent` handlers). If you want "pin" behavior where scrolling up holds position during streaming, that's a feature that needs explicit UX design, not an accidental default.

### 5. Minor: formatting-only changes

The import reorder in `main.rs` and blank line removals in `edit_file.rs`/`write_file.rs` are noise. Not a blocker, but don't mix formatting cleanup into feature PRs — it makes the diff harder to review.

src/frontend/tui.rs
```diff
@@ -231,6 +255,22 @@ async fn run_app(
...
+                        app.scroll_up(amount);
+                    }
+                    KeyEvent {
+                        code: KeyCode::Char('d'),
+                        modifiers: KeyModifiers::CONTROL,
+                        ..
+                    } => {
+                        let amount = app.half_page();
+                        app.scroll_down(amount);
+                    }
```
Fizzizist: These lines are fully repeated below 3 times in this file. There has got to be a better way to handle this

Fizzizist (commented):
The controls are working but there is something weird happening with the viewport where, when new messages are coming in, it's not automatically scrolling to the bottom. It seems to mostly happen when the user submits a new prompt. I submit a prompt and can't see it until the agent starts responding

Fizzizist (commented):
it's better but it's not perfect. Here is what I am seeing

1. open app
2. send a few messages back and forth with agent (scrolling looks normal; scrolls to the bottom when new stuff appears
3. scroll up with `ctrl+u` a few page lengths
4. send another message 
5. Now it scrolls down, but not all the way. I only see `You:` line and have to wait for the assistant message to appear before I see my full message

PR #68, opened 2026-04-07
Add event buffering support to VertexSseParser

### LINKED ISSUE #56: event queue in Vertex backend
We fixed a bug in the zai backend where it was attempting to send more than one event out from the parse function but the agent was only acting on the first one. Now we have fixed that logic in zai backend and we patched the vertex backend to just send a vec of its first event out, but we need to go in and make it also able to send more than 1.

Fizzizist:
## Summary

- Adds event buffering capability to `VertexSseParser`, matching the architecture of the `zai` backend
- Previously, the Vertex backend could only emit the first event from a parse operation
- Now supports emitting multiple events from a single SSE chunk

## Changes

- Added `event_buffer: Vec<StreamEvent>` field to `VertexSseParser`
- Added `fill_buffer()` method that populates the event buffer from SSE data
- Updated `parse()` method to return buffered events one at a time (backward compatible)
- Updated SSE stream callback to drain the entire event buffer instead of returning only the first event

## Test Plan

- [x] Added unit tests for `fill_buffer` and event buffering behavior
- [x] All existing tests pass (195 tests)
- [x] Verified backward compatibility with existing `parse()` method

## Related Issue

Fixes #56

### DIFF
--- a/src/backend/vertex.rs
+++ b/src/backend/vertex.rs
@@ -69,7 +69,8 @@ impl VertexSseParser {
                             anyhow::anyhow!("tool_use content_block_start missing 'name'")
                         })?
                         .to_string();
-                    self.event_buffer.push(StreamEvent::ToolUseStart { id, name });
+                    self.event_buffer
+                        .push(StreamEvent::ToolUseStart { id, name });
                 }
             }
             "content_block_delta" => {
@@ -444,7 +445,11 @@ mod tests {
         parser
             .fill_buffer(r#"{"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Hello"}}"#)
             .expect("should fill buffer successfully");
-        assert_eq!(parser.event_buffer.len(), 1, "should have 1 event in buffer");
+        assert_eq!(
+            parser.event_buffer.len(),
+            1,
+            "should have 1 event in buffer"
+        );
     }
 
     #[test]
--- a/src/backend/vertex.rs
+++ b/src/backend/vertex.rs
@@ -18,7 +18,7 @@ pub struct VertexSseParser {
     tool_use_indices: HashSet<u64>,
     input_tokens: u32,
     /// Buffer for events when multiple events appear in a single SSE chunk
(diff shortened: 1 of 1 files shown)


Fizzizist:
## Code Review

### Issues

**1. `event_buffer` should be private, not public**

Line 21: `pub event_buffer: Vec<StreamEvent>`

This breaks encapsulation. The `zai.rs` implementation correctly keeps `event_buffer` private. There's no reason for this to be public - it's an internal implementation detail. If tests need access, use a test-only visibility attribute or expose it via a proper getter method.

**2. Excessive `.unwrap_or()` usage violates policy**

Multiple locations use `.unwrap_or()` which silently ignores parsing errors:
- Line 51: `.unwrap_or(0)`
- Line 62: `.unwrap_or("")`  
- Line 70: `.unwrap_or(0)`
- Line 81: `.unwrap_or("")`
- Line 85: `.unwrap_or("")`
- Line 91: `.unwrap_or(0)`

This is technically not `.unwrap()`, but it's just as bad - it hides malformed data from the upstream API. If the API contract changes, this code will silently produce garbage instead of failing fast. Use proper error propagation or `.ok_or()` with context.

**3. Tests don't cover the actual bug scenario**

The PR claims to fix issue #56 (multiple events in a single SSE chunk), but the tests don't verify this actually works. The new tests:
- `parser_fill_buffer_populates_event_buffer` - Only checks that ONE event gets buffered
- `parse_returns_buffered_events_one_at_a_time` - Only tests draining, not multi-event emission

Add a test that actually fills the buffer with multiple events from a single `fill_buffer` call and verifies they all drain correctly.

**4. Unknown event types silently dropped**

Line 118: `_ => {}`

Unknown SSE event types are ignored with no logging. This is copied from `zai.rs`, but that doesn't make it right. At minimum, this should log at debug level so that when the API changes, we have some indication of what's being ignored.

---

### Nitpicks

- The empty-string API for draining buffered events is clever but hacky. Consider an explicit `drain_buffer()` method instead.

The approach is correct, but fix the visibility and error handling before merging.

PR #76, opened 2026-04-08
Add skill tool (#61)

### LINKED ISSUE #61: Skill Tool
On startup, the app should look for skills in the following locations:
`$PWD/.claude/skills`
`$HOME/.claude/skills`
If their are directories present, those should be added to a skills list with mappings to the SKILL.md filepath. The mapping keys are provided in the agent's initial context so that it can use the skill tool to invoke them

## Acceptance Criteria
- New tool added to `src/tools` called `skill`
- Skill tool is added to tool registry for agent
- Agent can invoke Skill tool with the name of an available skill
- skill tool reads the file that the given name is mapped to and adds the skill prompt into the agent's context
- tests

Fizzizist:
## Summary

- New `src/tools/skill.rs` with `SkillTool` and `discover_skills` that scan `$PWD/.claude/skills` and `$HOME/.claude/skills` for subdirectories containing `SKILL.md`
- `SkillTool` reads the mapped file by name and returns its content to the agent context
- `Agent::load_skills` injects available skill names into initial context as a user message
- Skill tool registered in `main.rs`; PWD skills override HOME skills with the same name

## Test plan

- [ ] `cargo test` — unit tests in `skill.rs` cover discovery, override, missing name, execute success/failure
- [ ] Agent `load_skills` tests verify history is populated with skill names
- [ ] `cargo clippy` and `cargo fmt` pass

Closes #61

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -78,7 +78,7 @@ impl Agent {
             String::from("The following skills are available via the `skill` tool:\n\n");
         for name in names {
             let path = &skills[name];
-            let desc = crate::tools::skill::first_line(path)
+            let desc = crate::tools::skill::skill_description(path)
                 .unwrap_or_else(|| "(no description)".to_string());
             content.push_str(&format!("- {}: {}\n", name, desc));
         }
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -64,6 +64,8 @@ impl Agent {
         Ok(self)
     }
 
+    // TODO: synthetic User messages are a design smell; use Role::System or a metadata
+    // mechanism if one is added in future.
     pub fn load_skills(&self, skills: &std::collections::HashMap<String, std::path::PathBuf>) {
         if skills.is_empty() {
             return;
@@ -75,7 +77,10 @@ impl Agent {
         let mut content =
             String::from("The following skills are available via the `skill` tool:\n\n");
         for name in names {
-            content.push_str(&format!("- {}\n", name));
+            let path = &skills[name];
+            let desc = crate::tools::skill::first_line(path)
+                .unwrap_or_else(|| "(no description)".to_string());
+            content.push_str(&format!("- {}: {}\n", name, desc));
         }
 
         lock(&self.history).push(Message::text(Role::User, content));
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -64,6 +64,23 @@ impl Agent {
         Ok(self)
     }
 
+    pub fn load_skills(&self, skills: &std::collections::HashMap<String, std::path::PathBuf>) {
+        if skills.is_empty() {
+            return;
+        }
+
+        let mut names: Vec<&str> = skills.keys().map(String::as_str).collect();
+        names.sort();
+
+        let mut content =
+            String::from("The following skills are available via the `skill` tool:\n\n");
+        for name in names {
+            content.push_str(&format!("- {}\n", name));
+        }
+
+        lock(&self.history).push(Message::text(Role::User, content));
+    }
+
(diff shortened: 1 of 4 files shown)


Fizzizist (commented):
## Code Review: Add skill tool (#61)

### Overall

Clean implementation. Follows existing patterns (mirrors `load_context_files`), tests are thorough, and the override semantics are correct. A few things need attention.

### Issues

**1. `discover_skills_from_env` silently swallows a missing cwd (line 94)**

```rust
let pwd = std::env::current_dir().unwrap_or_default();
```

`unwrap_or_default()` gives you an empty `PathBuf` if the cwd doesn't exist (e.g. it was deleted). That empty path will then join to `.claude/skills`, which is a relative path — meaning skill discovery silently becomes nondeterministic depending on what the OS resolves it to. This should either propagate the error or at least log it. The `home_dir` fallback to `Path::new("")` has the same problem.

**2. `skills.clone()` in `main.rs` (line 107)**

You clone the entire `HashMap<String, PathBuf>` just to pass it to both `SkillTool::new` and `agent.load_skills`. `load_skills` only needs `&HashMap`, which you already do, but then `SkillTool::new` takes ownership. The clone is technically fine but wasteful — consider passing the reference to `load_skills` first, then moving ownership into `SkillTool::new`. Or just take a reference in `SkillTool::new` and clone internally. Minor, but unnecessary allocations are unnecessary allocations.

**3. `load_skills` injects a `Role::User` message (agent.rs line 81)**

`load_context_files` does the same thing so this is consistent, but injecting synthetic `User` messages into the conversation history is a design smell. The model will see these as things the human said, not as system context. If there's ever a `Role::System` or metadata mechanism, this should use it. Not blocking, but worth a TODO or tracking issue.

**4. No description returned in skill listing**

The issue says skills should be available for the agent to invoke, but the `load_skills` method only lists skill names — no descriptions. SKILL.md files presumably contain a description or purpose. Without surfacing even a one-liner, the agent has to blindly invoke a skill to find out what it does. Consider parsing a frontmatter `description` field or at least the first line of SKILL.md and including it in the listing.

**5. Test coverage gap: symlinks and non-UTF8 directory names**

`discover_skills` uses `to_str()` on directory names and silently skips non-UTF8 names. That's fine, but there's no test documenting this behavior. Also no test for what happens with symlinked skill directories (valid use case for shared skill repos). Not critical, but worth a test or two.

### Nits

- `skill.rs` line 83: The `HashMap::insert` call for the override behavior is correct but relies on iteration order of `skill_dirs` (home first, pwd second). A comment noting that pwd intentionally comes second to override home would prevent future refactors from breaking this.

### Verdict

Approve with minor fixes. Issues 1 and 4 are worth addressing before merge; the rest can be follow-up.

src/agent.rs
```diff
@@ -64,6 +64,28 @@ impl Agent {
...
+        }
+
+        let mut names: Vec<&str> = skills.keys().map(String::as_str).collect();
+        names.sort();
+
+        let mut content =
+            String::from("The following skills are available via the `skill` tool:\n\n");
+        for name in names {
+            let path = &skills[name];
+            let desc = crate::tools::skill::first_line(path)
```
Fizzizist: Feeding it the first line of the file isn't right.

It would be better to search for a description at the top of the file by filtering the file by `---`

For example, a typical skill file will have:
```markdown
---
name: code-review
description: instructions for a code review process
---

<skill prompt here>
```

Fizzizist: Fixed. Replaced first-line heuristic with a proper frontmatter parser — `skill_description` now looks for `---` delimiters and extracts the `description:` field.

PR #81, opened 2026-04-08
Sessions implementation

### LINKED ISSUE #75: Sessions
We need to be able to go back to a session if it is closed by accident or we just want to stop for a while.

We'll need the [turso](https://docs.rs/turso/latest/turso/) sdk as a new dependency. Go read the docs first before you implement the database I/O logic

## Acceptance Criteria
- Conversation history is stored in a [Turso DB](https://docs.turso.tech/introduction) file. 
  - schema should be just a single `conversation` table. We'll store each session as it's own file
  - conversation rows should be stored _as the conversation is proceeding_. Instead of bulk-saving at certain points, just write each row as it comes through. That way, we don't lose any data if something goes wrong.
- Session IDs should be UUIDv7 and should be contained in the DB file name.
- DB files can be stored in `$HOME/.config/illustrious-manager/sessions`
- allow a `--session-id` flag to be passed which will look for that session DB file and load it if available. If the the file isn't their, then it should validate that it is a valid UUIDv7 and create a new DB file with that session ID
- The program should print out the session ID to the screen on exit
- A new session should always be created in the absence of the `--session_id` flag -- sessions should not be optional
- tests

Fizzizist:
Closes #75 

# Changes
- turso and uuid7 dependencies
- agent saving and loading conversation history
- added save on agent response complete event
- added save at end of single-shot
- Session struct as turso repository
- tests

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -7,6 +7,7 @@ use futures::channel::mpsc;
 use crate::backend::LlmBackend;
 use crate::config::{ConfirmationMode, ToolsConfig};
 use crate::context_files::{ContextFile, discover_context_files_from_env};
+use crate::session::Session;
 use crate::tools::ToolRegistry;
 use crate::types::{
     AgentEvent, BoxStream, ConfirmationResponse, ContentBlock, Message, RequestConfig, Role,
@@ -26,6 +27,7 @@ pub struct Agent {
     tools: Arc<ToolRegistry>,
     max_tool_iterations: u32,
     confirmation_mode: ConfirmationMode,
+    session: Option<Arc<Session>>,
 }
 
 // Recover from a poisoned mutex: a thread panicked while holding the lock, leaving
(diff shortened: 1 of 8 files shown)


Fizzizist:
(Resolved)

src/main.rs
```diff
@@ -120,6 +124,15 @@ async fn main() -> Result<()> {
     );
     agent.load_skills(&skills);
 
+    let session = if let Some(ref session_id) = cli.session_id {
```
Fizzizist: idk why this is optional. There should _always_ be a session created even if the `--session-id` flag is not provided

Fizzizist (commented):
(Resolved)

src/agent.rs
```diff
@@ -107,6 +107,16 @@ impl Agent {
         lock(&self.history).push(Message::text(Role::User, content));
     }
 
+    pub fn load_history(&self, messages: Vec<Message>) {
+        let mut guard = lock(&self.history);
+        guard.extend(messages);
+    }
+
+    pub async fn save_history_to_session(&self, session: &crate::session::Session) -> Result<()> {
```
Fizzizist: Seems like we only are doing these bulk saving events.. it would be better to save individual entries as we go so that we limit the risk of data loss if something goes wrong before it can bulk save

PR #99, opened 2026-04-13
fix: windowed rendering with cached line counts for conversation sync

### LINKED ISSUE #86: Conversation becomes out of sync

Fizzizist:
## Summary

Fixes #86: Conversation becomes out of sync after many rounds of inference and tool calling.

## Root Cause

The visual line count calculation used a naive `line.width().div_ceil(text_width)` formula that diverges from `Paragraph`'s actual `WordWrapper` wrapping behavior. This caused scroll position to become increasingly inaccurate as conversations grew, making the conversation appear "behind" by several messages.

### Core Fix: Accurate Line Count Caching
- Enable ratatui's `unstable-rendered-line-info` feature to use `Paragraph::line_count()` for accurate wrapped line counting that matches actual rendering
- Cache wrapped line counts per `ConversationEntry` — computed once per entry per width change, NOT on every render or token event
- Pre-render and cache `Line` objects at entry creation time (`ConversationEntry::new()`)

### Performance: Windowed Rendering
- Implement `collect_window_lines()` which skips entries entirely outside the viewport
- This gives O(visual area) rendering instead of O(conversation length)
- Streaming `current_response` uses the cheap estimate function (bounded to one entry, no caching needed)

### Supporting Changes
- Store `text_width` in `App` so `max_scroll()` uses the actual viewport width for accurate clamping
- `render_app()` now takes `&mut App` to allow line count cache updates during render
- `ConversationEntry` constructed via `::new()` instead of struct literals (private cache fields)

## Acceptance Criteria

- ✅ Correct count of rendered lines allows user to see the full conversation from top to bottom with scrolling
- ✅ `Paragraph.line_count` is NOT used on every render or every token event (cached per entry)
- ✅ `ConversationArea` is rendered with O(visual area) efficiency (windowed approach)
- ✅ Regression tests:
  - `regression_conversation_sync_after_many_rounds`: verifies max_scroll matches actual visual line count after 30 rounds
  - `regression_scroll_to_top_shows_first_message`: verifies first message is reachable by scrolling up
  - `regression_auto_scroll_shows_latest_after_streaming`: verifies latest content visible at bottom


Fizzizist:
## Code Review: Windowed Rendering with Cached Line Counts

**Verdict: Needs Changes**

The root cause analysis is correct — the naive `line.width().div_ceil(text_width)` formula diverges from `Paragraph`'s actual `WordWrapper` behavior, causing scroll position drift. The architectural approach (cached line counts + windowed rendering) is sound. But the implementation has issues that need addressing before merge.

---

### 1. ❗ Required: `render_current_response_lines` duplicates `render_entry_lines`

These two functions are structurally identical — both construct a header line with label/color, then prefix body lines with `"  "`. The only differences:

- `render_entry_lines` takes `role` as a parameter; `render_current_response_lines` hardcodes `ConversationRole::Assistant`
- `render_entry_lines` calls `maybe_truncate`; `render_current_response_lines` does not
- `render_entry_lines` appends a trailing blank line; `render_current_response_lines` does not

This is a maintenance landmine. When someone changes line formatting in one place, they will forget the other. The trailing blank line difference also suggests the `current_response` rendering is inconsistent with entry rendering — is that intentional?

**Action:** Factor into a single shared function, or have `render_current_response_lines` delegate to `render_entry_lines` with a synthesized role.

---

### 2. ⚠�� Fragile: `text_width = 0` on first frame before render

`App::new()` sets `text_width = 0`. If `scroll_up` or `scroll_down` fires before the first `render_app()` frame, `max_scroll()` passes `0` to `wrapped_line_count`. The cache guard `if text_width == self.cached_width && self.cached_width > 0` prevents a crash, and `estimate_wrapped_count` handles `text_width == 0` by returning `lines.len()`. So it won't panic — but it computes an inaccurate `max_scroll`. Likely benign since the event loop renders before processing scroll keys, but it's fragile. Consider initializing `text_width` to a reasonable default or adding a guard.

---

### 3. ⚠️ Note: `render_current_response_lines` called fresh every frame during streaming

The PR description correctly states `Paragraph::line_count` is not called on every render for cached entries. But `render_current_response_lines` runs markdown parsing + line construction on every `render()` and `max_scroll()` call for the streaming response. For long responses with complex markdown, this is O(response_size) per frame. The PR acknowledges this ("bounded to one entry, no caching needed"), which is reasonable — but worth a TODO comment for future optimization if it becomes a bottleneck.

---

### 4. Minor: `entries: &'a mut [ConversationEntry]` borrow infectiousness

The change from `&[ConversationEntry]` to `&mut [ConversationEntry]` cascaded through all callers, forcing `render_app` to take `&mut App`. An alternative would be interior mutability via `Cell<u16>` for the cache fields, avoiding the need for exclusive borrows. The current approach is acceptable, just noting the trade-off.

---

## Test Coverage Assessment

**Strong.** The regression tests directly exercise the acceptance criteria:

- `regression_conversation_sync_after_many_rounds`: Validates `max_scroll` matches actual `Paragraph::line_count` after 30 rounds — the core correctness invariant. ✅
- `regression_scroll_to_top_shows_first_message`: End-to-end render verification at scroll position `max`. ✅
- `regression_auto_scroll_shows_latest_after_streaming`: Verifies latest content visible at `scroll_offset=0`. ✅
- `regression_large_entry_scrollable_through_middle`: Sweeps all scroll positions — every line must appear somewhere. ✅
- `regression_adjacent_scroll_positions_differ`: Catches degenerate case where scroll doesn't move viewport. ✅

Cache tests (`cached_line_count_matches_paragraph_line_count`, `cached_line_count_is_reused_for_same_width`, `cached_line_count_recomputes_for_different_width`) validate caching behavior directly. ✅

**Missing:** No test for the boundary case where the viewport window spans cached entries and the current streaming response (i.e., the transition between the last entry and `current_response` in `collect_window_lines`). This is the trickiest part of the windowing logic and deserves a dedicated test.

---

## Summary

| Category | Status |
|---|---|
| Bug fix correctness | ✅ Root cause addressed |
| Performance | ✅ O(visual area) rendering |
| `line_count` caching | ✅ Per-entry, invalidates on width change |
| Test coverage | ✅ Strong regression tests |
| Code duplication | ❌ Must fix before merge |
| `text_width = 0` edge case | ⚠️ Fragile but benign |
| Streaming perf for long responses | ⚠️ Acceptable, document it |

**Required before merge:**
1. Eliminate duplication between `render_entry_lines` and `render_current_response_lines`
2. Add a test for the viewport boundary between cached entries and streaming response

PR #101, opened 2026-04-14
fix: preserve conversation history on errors

### LINKED ISSUE #94: max_tokens reached, kills conversation
If the agent is busy doing something and encounters a `max_tokens` error, it kills the conversation and the agent seems to think that it is at the very beginning on the conversation again. If I exit, and then re-join with the same session ID, the agent is able to see where it left off. I shouldn't have to exit and re-join. The full history should be getting fed to it.

## Acceptance Criteria
- when _any_ Error is hit (e.g. `max_tokens` error), agent should stop and wait for user input, _but_ should retain the full conversation so far _including_ the error message. With the error message in-context, the agent knows why a stoppage occurred.
- regression tests

Fizzizist:
## Problem

When any error occurred during a conversation (e.g. `max_tokens` reached, backend connection failure, mid-stream error), the agent was truncating the conversation history back to its pre-send state using `history.truncate(pre_send_len)`. This caused the agent to lose all context accumulated during the current turn and behave as if it was at the very beginning of the conversation.

Users had to exit and re-join with the same session ID to recover the conversation, because the session DB still had the history — it was only the in-memory history that got wiped.

## Solution

Instead of truncating history on error, the agent now:

1. **Preserves all accumulated history** — user messages, assistant responses, tool calls, and tool results all remain in the conversation
2. **Saves any partial assistant response** — if the model was mid-stream when the error occurred (e.g. `max_tokens`), the partial text is saved as an assistant message
3. **Adds the error to history** — the error message is added as a `[ERROR]` user message so the agent knows why the stoppage occurred and can react appropriately on the next turn
4. **Persists to session DB** — error messages and partial responses are saved to the session database so they survive restarts

## Changes

- `src/agent.rs`: Replaced all three `history.truncate(pre_send_len)` calls with error-preserving logic across the three error paths:
  - Max tool iterations exceeded
  - Backend `send_message()` returns error
  - Mid-stream error (e.g. `max_tokens` reached)

## Tests (TDD)

5 new regression tests, 1 updated test:

| Test | What it verifies |
|------|-----------------|
| `backend_error_on_second_iteration_preserves_history` | Updated from `_clears_history` — verifies history is preserved, not wiped |
| `backend_send_error_preserves_user_message_in_history` | Backend connection error keeps user message |
| `stream_error_preserves_partial_text_in_history` | Mid-stream error saves partial assistant text |
| `error_message_is_added_to_history_as_user_message` | Error text is added to history for agent context |
| `max_iterations_error_preserves_history` | Max iterations error preserves full conversation |

Closes #94

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -138,7 +138,6 @@ impl Agent {
         input: String,
         confirmation_rx: Option<mpsc::UnboundedReceiver<ConfirmationResponse>>,
     ) -> Result<BoxStream<AgentEvent>> {
-        let pre_send_len = lock(&self.history).len();
         let user_msg = Message::text(Role::User, input);
         lock(&self.history).push(user_msg.clone());
 
@@ -159,10 +158,11 @@ impl Agent {
 
             'outer: loop {
                 if iterations >= max_iterations {
-                    lock(&history_arc).truncate(pre_send_len);
-                    let _ = event_tx.unbounded_send(AgentEvent::Error(format!(
-                        "Max tool iterations ({max_iterations}) exceeded"
-                    )));
+                    let error_msg = format!("Max tool iterations ({max_iterations}) exceeded");
+                    let error_user_msg = Message::text(Role::User, format!("[ERROR] {error_msg}"));
+                    lock(&history_arc).push(error_user_msg.clone());
+                    let _ = session.lock().await.insert_message(&error_user_msg).await;
+                    let _ = event_tx.unbounded_send(AgentEvent::Error(error_msg));
                     break;
                 }
                 iterations += 1;
@@ -171,8 +171,12 @@ impl Agent {
                 let backend_stream = match backend.send_message(&history_snapshot, &config).await {
                     Ok(s) => s,
                     Err(e) => {
-                        lock(&history_arc).truncate(pre_send_len);
-                        let _ = event_tx.unbounded_send(AgentEvent::Error(e.to_string()));
+                        let error_msg = e.to_string();
+                        let error_user_msg =
+                            Message::text(Role::User, format!("[ERROR] {error_msg}"));
+                        lock(&history_arc).push(error_user_msg.clone());
+                        let _ = session.lock().await.insert_message(&error_user_msg).await;
+                        let _ = event_tx.unbounded_send(AgentEvent::Error(error_msg));
                         break;
                     }
                 };
@@ -218,8 +222,20 @@ impl Agent {
                         }
                         Ok(StreamEvent::Done) => break,
                         Err(e) => {
-                            lock(&history_arc).truncate(pre_send_len);
-                            let _ = event_tx.unbounded_send(AgentEvent::Error(e.to_string()));
+                            if !text_accumulated.is_empty() {
+                                let partial_msg = Message {
+                                    role: Role::Assistant,
+                                    content: vec![ContentBlock::Text(text_accumulated)],
+                                };
+                                lock(&history_arc).push(partial_msg.clone());
+                                let _ = session.lock().await.insert_message(&partial_msg).await;
+                            }
+                            let error_msg = e.to_string();
+                            let error_user_msg =
+                                Message::text(Role::User, format!("[ERROR] {error_msg}"));
+                            lock(&history_arc).push(error_user_msg.clone());
+                            let _ = session.lock().await.insert_message(&error_user_msg).await;
+                            let _ = event_tx.unbounded_send(AgentEvent::Error(error_msg));
                             break 'outer;
                         }
                     }
@@ -877,7 +893,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn backend_error_on_second_iteration_clears_history() {
+    async fn backend_error_on_second_iteration_preserves_history() {
         let backend = SequencedBackend::new(vec![
             tool_call_response("t1", "bash", r#"{}"#),
             vec![Err(anyhow::anyhow!("backend failure on iteration 2"))],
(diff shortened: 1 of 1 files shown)


Fizzizist (commented):
# Code Review: PR #101 — fix: preserve conversation history on errors

## Summary

**PR:** `fix/issue-94-error-preserves-history` → `trunk`
**Author:** Fizzizist (Peter Vlasveld)
**Issue:** #94 — `max_tokens` reached kills conversation; agent loses all accumulated context on any error.

**Verdict: Solid fix.** The core behavioral change is correct and well-motivated. The old `truncate(pre_send_len)` was a nuclear option that nuked all context on error — this PR correctly preserves history and injects error context. I have a few critiques.

---

## ✅ What's Good

1. **The fix is correct.** Replacing `history.truncate(pre_send_len)` with history preservation + error injection is the right approach. All three error paths (max iterations, `send_message` failure, mid-stream error) are handled consistently.

2. **Partial text preservation on mid-stream errors.** The `if !text_accumulated.is_empty()` guard before saving the partial assistant message is a nice touch — it avoids polluting history with empty assistant messages when the error occurs before any tokens arrive.

3. **Error message injected as `[ERROR]` user message.** This gives the model context about why the conversation stopped, which directly addresses the issue's acceptance criteria.

4. **Session persistence.** Error messages and partial responses are persisted to the session DB, so they survive restarts. This was explicitly called out in the issue (users had to exit and re-join).

5. **Tests are well-structured.** 5 new tests + 1 updated test, each targeting a specific error scenario. Good use of TDD.

---

## 🔴 Issues

### 1. `backend_send_error_preserves_user_message_in_history` is misnamed and tests the wrong code path

The `SequencedBackend::send_message()` always returns `Ok(Box::pin(stream::iter(events)))`. It never returns `Err(...)`. So when the test provides `vec![Err(anyhow::anyhow!("connection refused"))]`, that error comes out of the *stream*, not from `send_message` itself.

This means the test exercises the **mid-stream error** path (lines 222-239), NOT the **`send_message` error** path (lines 171-180). The `send_message` error path has **zero test coverage**.

You need either:
- A `FailingBackend` that returns `Err(...)` from `send_message` directly, or
- Modify `SequencedBackend` to support returning errors from `send_message`

And rename the current test to something like `stream_error_preserves_user_message_in_history`.

### 2. Duplicated error-handling logic across three sites

The pattern of:
```rust
let error_msg = ...;
let error_user_msg = Message::text(Role::User, format!("[ERROR] {error_msg}"));
lock(&history_arc).push(error_user_msg.clone());
let _ = session.lock().await.insert_message(&error_user_msg).await;
let _ = event_tx.unbounded_send(AgentEvent::Error(error_msg));
```

is copy-pasted verbatim across all three error sites (lines 161-166, 174-179, 233-238). This is a DRY violation and a maintenance hazard — if the error-handling behavior needs to change (e.g., different format, additional logging), you'd need to update three places.

Extract a helper closure or function, e.g.:

```rust
let emit_error = |msg: String| {
    let error_user_msg = Message::text(Role::User, format!("[ERROR] {msg}"));
    lock(&history_arc).push(error_user_msg.clone());
    // ... session persist, event send
};
```

The complication is the `async` session persist, but you could either make it an async closure (nightly) or a small async helper function that takes all the Arcs.

### 3. `text_accumulated` is moved in the `Err` branch — fragile but currently correct

In the mid-stream error handler (line 228):
```rust
content: vec![ContentBlock::Text(text_accumulated)],
```

This moves `text_accumulated`. It works because `break 'outer` follows, so the variable is never used again. But this is fragile — if someone adds code after the `if` block that references `text_accumulated`, it'll fail to compile. A `.clone()` here would be defensive, though I acknowledge the compiler will catch it.

This is a nit, not a blocker.

---

## 🟡 Minor Observations

### 4. No test verifies the `[ERROR]` prefix format

The `error_message_is_added_to_history_as_user_message` test only checks that the history contains a user message with `"max_tokens reached"` — it doesn't verify the `[ERROR]` prefix. If the prefix format matters (e.g., the model is expected to parse it), the test should assert on the full `"[ERROR] max_tokens reached"` string.

### 5. The `max_iterations_error_preserves_history` test doesn't verify *what* is in history

It asserts `!history.is_empty()` and that the original user message is there, but doesn't verify that tool call results from the iterations that *did* succeed are preserved. That's the main value proposition of this fix — that completed work isn't lost. Adding an assertion that checks for the presence of tool results or assistant messages from completed iterations would make this test more robust.

### 6. Silently discarding session persist errors

```rust
let _ = session.lock().await.insert_message(&error_user_msg).await;
```

The `let _ =` discards any error from persisting to the session DB. If the DB write fails, the in-memory history and persisted history diverge silently. This was pre-existing behavior (the happy path does the same), so not a regression, but worth noting.

---

## Test Coverage Summary

| Error Path | Test Coverage | Verdict |
|---|---|---|
| Max iterations exceeded | ✅ `max_iterations_error_preserves_history` | Covered |
| `send_message()` returns `Err` | ❌ No test exercises this path | **Gap** |
| Mid-stream error (no partial text) | ✅ `backend_send_error_preserves_user_message_in_history`¹ | Covered (misnamed) |
| Mid-stream error (with partial text) | ✅ `stream_error_preserves_partial_text_in_history` | Covered |
| Error added to history | ✅ `error_message_is_added_to_history_as_user_message` | Covered |
| Backend error after tool iteration | ✅ `backend_error_on_second_iteration_preserves_history` | Covered |

¹ This test is misnamed — it tests mid-stream errors, not `send_message` errors.

---

## Verdict

The behavioral fix is sound and directly addresses issue #94. The test suite is good but has one meaningful gap (the `send_message` error path is untested) and one misleading test name. The duplicated error-handling code is a minor code quality issue.

**Required changes before merge:**
1. Add a test that actually exercises the `send_message` error return path (needs a backend that returns `Err` from `send_message`, not from the stream)
2. Rename `backend_send_error_preserves_user_message_in_history` to reflect that it tests mid-stream errors

**Suggested (non-blocking):**
3. Extract the duplicated error-handling logic into a helper
4. Strengthen `max_iterations_error_preserves_history` to verify completed iteration artifacts are preserved
5. Assert on the full `[ERROR]` prefix in `error_message_is_added_to_history_as_user_message`

PR #115, opened 2026-04-16
feat: 429 exponential backoff retry for all backends

### LINKED ISSUE #114: 429 exponential backoff retry
Certain errors should be retried with exponential backoff before a certain number of times. This include 429 rate limit errors, as well as 5** errors. Right now, it just stops streaming entirely when an error occurs.

Fizzizist:
Closes #114

Implements exponential backoff retry when any backend returns a 429 rate-limit error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1644,6 +1644,7 @@ dependencies = [
  "syntect-assets",
  "tempfile",
  "tokio",
+ "tokio-retry2",
  "toml",
  "tui-markdown",
  "turso",
@@ -3833,6 +3834,17 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.117",
+]
+
+[[package]]
+name = "tokio-retry2"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0266d56e57e6b29becbfce5daa6add8730941ca8192ddd7c24d25bf90c32a743"
+dependencies = [
+ "pin-project",
+ "rand 0.9.2",
+ "tokio",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


Fizzizist:
## Implementation Plan

### Problem
Any backend (Vertex AI, z.ai) can return HTTP 429 rate-limit errors. These should be retried automatically with exponential backoff rather than surfaced immediately as hard errors.

### Deliverables
- **`src/backend/retry.rs`** — New `RetryBackend<B>` wrapper that implements `LlmBackend` and transparently retries `send_message` on 429 errors using exponential backoff + jitter (via `tokio-retry2`).
- **`src/backend/mod.rs`** — Both `VertexBackend` and `ZaiBackend` are wrapped with `RetryBackend` in `from_config()`.
- **`Cargo.toml`** — `tokio-retry2` dependency added with the `jitter` feature.

### Design Decisions
- **Wrapper pattern** — `RetryBackend<B>` wraps any `LlmBackend` without modifying existing backends. Clean separation of concerns.
- **429 detection** — `is_rate_limit_error()` checks if the error string contains "429". Both backends format errors as `"<Provider> returned <STATUS>: <body>"`, so this is reliable.
- **Retry strategy** — 1s initial delay, 60s cap, jitter applied, 5 max retries. Non-429 errors are `Permanent` and propagate immediately without retrying.
- **No interface changes** — `LlmBackend` trait and all call sites are unchanged.

### Tests (7 new, all passing)
- `is_rate_limit_error_detects_429_in_message`
- `is_rate_limit_error_rejects_other_status_codes`
- `is_rate_limit_error_rejects_non_http_errors`
- `retries_once_on_429_then_succeeds`
- `does_not_retry_on_non_429_error`
- `succeeds_on_first_try_without_retrying`
- `fails_after_exhausting_retries`

---
*Plan created by mach6*

PR #116, opened 2026-04-16
feat: JSON output format for single-shot mode

### LINKED ISSUE #70: JSON output
Allow `--output-format json` to be passed with `--single-shot` to make the program output in JSON format. 

## Schema
- `is_error` - indicates whether the run errored out or the agent was done
- `result` - the agent's output message

Fizzizist:
Closes #70

Adds `--output-format json` flag (valid only with `--single-shot`). Emits a single JSON line at the end of the run:

```json
{"is_error": false, "result": "the agent's response"}
```

On error: `{"is_error": true, "result": "error message"}`

Tool events are suppressed in JSON mode. Text mode is unchanged. Passing `--output-format json` without `--single-shot` is an error.

### DIFF
--- a/src/frontend/stdout.rs
+++ b/src/frontend/stdout.rs
@@ -6,28 +6,63 @@ use std::io::{self, BufRead, IsTerminal, Write};
 use crate::logging::Logger;
 use crate::types::{AgentEvent, BoxStream, ConfirmationResponse};
 
+#[derive(Debug, Clone, PartialEq, clap::ValueEnum)]
+pub enum OutputFormat {
+    Text,
+    Json,
+}
+
 pub async fn run(
     stream: BoxStream<AgentEvent>,
     confirm_tx: mpsc::UnboundedSender<ConfirmationResponse>,
+    format: OutputFormat,
     logger: Option<&mut Logger>,
 ) -> Result<()> {
     let is_tty = std::io::stdin().is_terminal();
     let stdout = io::stdout();
     let mut handle = stdout.lock();
     let mut stdin = io::BufReader::new(std::io::stdin());
-    run_with_writer(stream, &mut handle, confirm_tx, is_tty, &mut stdin, logger).await
+    run_with_writer(
+        stream,
+        &mut handle,
+        confirm_tx,
+        format,
+        is_tty,
+        &mut stdin,
+        logger,
+    )
+    .await
 }
 
 async fn run_with_writer<W: Write, R: BufRead>(
     mut stream: BoxStream<AgentEvent>,
     writer: &mut W,
     confirm_tx: mpsc::UnboundedSender<ConfirmationResponse>,
+    format: OutputFormat,
     is_tty: bool,
     stdin: &mut R,
     mut logger: Option<&mut Logger>,
 ) -> Result<()> {
+    match format {
+        OutputFormat::Text => {
+            run_text(&mut stream, writer, confirm_tx, is_tty, stdin, &mut logger).await
+        }
+        OutputFormat::Json => {
+            run_json(&mut stream, writer, confirm_tx, is_tty, stdin, &mut logger).await
+        }
+    }
+}
+
+async fn run_text<W: Write, R: BufRead>(
+    stream: &mut BoxStream<AgentEvent>,
+    writer: &mut W,
+    confirm_tx: mpsc::UnboundedSender<ConfirmationResponse>,
+    is_tty: bool,
+    stdin: &mut R,
+    logger: &mut Option<&mut Logger>,
+) -> Result<()> {
     while let Some(event) = stream.next().await {
-        if let Some(ref mut log) = logger {
+        if let Some(log) = logger.as_deref_mut() {
             log.log_event(&event)?;
             log.flush()?;
         }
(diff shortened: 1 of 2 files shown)


Fizzizist:
# Code Review — PR #116: JSON output format for single-shot mode

## Summary

Solid, focused PR. Closes #70 cleanly: adds `--output-format {text,json}` gated to `--single-shot` mode, factors out shared confirmation handling, keeps text-mode behavior byte-identical, and ships meaningful tests for both formats. Built and ran the binary against the live backend in both `text` and `json` modes (with and without tool use) — the happy paths work as advertised. Stdout stays clean JSON; the `Session ID:` line and any errors go to stderr.

## What works

- ✅ `--output-format json --single-shot "..."` emits a single well-formed JSON line on stdout.
- ✅ `--output-format json` without `--single-shot` errors out with a clear message; covered by a unit test.
- ✅ Tool events are correctly suppressed in JSON mode, no `[tool: ...]` noise in `result`.
- ✅ `Error` events produce `{"is_error": true, "result": "..."}` and exit non-zero.
- ✅ Shared `handle_confirmation` removes duplication between text and json paths.
- ✅ `OutputFormat` derives `clap::ValueEnum`, so `--help` will list valid values, and an invalid value (`yaml`) is rejected by clap before we reach our code.
- ✅ Test coverage is appropriate: success, error, tool-events-suppressed, and non-TTY confirmation are all covered for the JSON path.

## Findings

### 1. (Required) Multi-iteration JSON `result` concatenates intermediate text with no separator

Live test result against the model (prompt: "List the files in the current directory using the bash tool, then summarize"):

```json
{"is_error":false,"result":"I'll list the files using the bash tool.Directory contents:\n\n- **Config**: ..."}
```

Note `"...bash tool.Directory contents:..."` — the model's pre-tool text and post-tool text are smushed together with no whitespace boundary because `run_json` blindly concatenates every `TokenReceived` across every agentic iteration:

```rust
AgentEvent::TokenReceived(text) => {
    result_text.push_str(&text);
}
```

Two reasonable fixes — pick one and document the contract:

- **(a)** Use only the final assistant turn (the payload of `ResponseComplete`), discarding intermediate "thinking" text. This matches the Claude Code-style `result` semantics implied by the issue ("the agent's output message", singular).
- **(b)** Keep accumulating but insert a separator (e.g. `\n\n`) between iterations — i.e. on `ToolResult` push a separator into `result_text` if it's non-empty.

I lean toward (a): it's simpler, matches the spirit of the issue, and avoids leaking the agent's internal reasoning into a structured field that scripts will parse. Whichever you pick, add a regression test that covers a multi-iteration response with at least one tool call and asserts the boundary behavior.

### 2. (Required) Non-TTY tool-confirmation in JSON mode does not emit JSON before bailing

In `run_json`, the `ToolConfirmationRequired` arm calls `handle_confirmation(...)?`. When stdin is not a TTY, `handle_confirmation` writes a plain-text error to stderr and bails. The result: the program exits with no JSON line written to stdout — inconsistent with how `AgentEvent::Error` is handled (write JSON, then bail).

Scripts consuming JSON output now have to special-case "exit code != 0 AND stdout is empty" instead of just parsing one line. Fix: in `run_json`, treat a confirmation requirement under non-TTY as an error condition that writes the JSON envelope first, e.g.:

```rust
AgentEvent::ToolConfirmationRequired { name, .. } if !is_tty => {
    let _ = confirm_tx.unbounded_send(ConfirmationResponse::Rejected);
    is_error = true;
    result_text = format!("tool confirmation required in non-TTY mode (tool: {name})");
    break;
}
```

…then fall through to the `writeln!` and `bail!` at the bottom. Update the existing test `json_format_non_tty_confirmation_sends_rejection_and_errors` to also assert that stdout contains a parseable JSON object with `is_error: true`. Right now that test only checks that an error was returned and the channel got `Rejected`, which is exactly the buggy contract.

### 3. (Should fix) `let _ = writer;` in `handle_confirmation` is a code smell

```rust
fn handle_confirmation<W: Write, R: BufRead>(
    writer: &mut W,
    ...
) -> Result<()> {
    ...
    let _ = writer;
    Ok(())
}
```

The `writer` parameter is never used — the prompt goes to stderr — and the trailing `let _ = writer;` exists only to silence the unused-variable warning. Drop the parameter entirely (and the corresponding `W: Write` bound and the `writer` argument at both call sites). Cleaner signature, no dead-weight generic, no apology line at the end of the function.

### 4. (Nit) `OutputFormat::Text != cli.output_format` requires `Clone` only because of how it's threaded

`OutputFormat` is a 2-variant fieldless enum. Add `Copy` to the derive list and drop the `.clone()` in `main.rs`. Pure cleanup.

### 5. (Nit) `--output-format` rejection happens after some setup work

`determine_mode` runs after `Cli::parse`, but the validation `--output-format requires --single-shot` could equivalently be expressed declaratively with clap's `requires`/`conflicts_with` so it's surfaced as a parse error and handled uniformly with other arg validation. Not blocking — the current bail produces a clear message — but it's the more idiomatic clap pattern. If you keep the bail, fine.

### 6. (Nit) Pre-existing clippy noise

`cargo clippy --all-targets` flags `assert_eq!(cli.debug, true|false)` in `src/main.rs` tests. Not introduced by this PR (verified against trunk), but if you're touching that test module already, knocking these out costs nothing.

## Tests

Test coverage for the new behavior is good — four JSON-format tests plus three CLI-parse tests. Per the policy in `CLAUDE.md`, any bug fix should ship with a regression test; the two required findings above each call for one. Specifically:

- A test where the stream contains TokenReceived → ToolResult → TokenReceived → ResponseComplete, asserting the shape of `result`.
- An update to `json_format_non_tty_confirmation_sends_rejection_and_errors` asserting the JSON envelope is written to `buf` before the error returns.

## Verdict

**Request changes.** Findings 1 and 2 are real correctness/UX bugs in the new feature surface; 3 is dead code that should not land. 4–6 are polish.

Fizzizist:
## Progress

All review findings addressed.

**Finding 1** — JSON result now contains only the final assistant turn. On ToolUseReceived, accumulated pre-tool text is cleared and accumulation pauses until ToolResult. Regression test: `json_format_multi_iteration_result_contains_only_final_turn`.

**Finding 2** — Non-TTY confirmation in JSON mode now writes the JSON envelope (`is_error: true`) before bailing. Non-TTY guard moved into `run_json` directly so `handle_confirmation` never runs in that path. Updated test renamed to `json_format_non_tty_confirmation_emits_json_and_errors` and now asserts the JSON is parseable with `is_error: true`.

**Finding 3** — Dropped the `writer` parameter and `W: Write` bound from `handle_confirmation` entirely. The trailing `let _ = writer` is gone.

**Finding 4** — Added `Copy` to `OutputFormat`; dropped `.clone()` in main.rs.

**Finding 5** — Left as-is (noted as nit, current bail produces a clear message).

**Finding 6** — `assert_eq!(cli.debug, true/false)` replaced with `assert!()`/`assert!(!()`.

Issue #120, opened 2026-04-17
Hard-coded workflows


Fizzizist:
Since we have an agent abstraction, it should be possible to have initiate hard-coded workflows. That is, workflows that have a set of deterministic steps along with the non-deterministic agent steps along the way. 

An example user story for a workflow:
- config defines a "high thinking model" and an "implement" model. Different models used for planning v implementation.
1. User submits a workflow using a slash command with an input issue number: `/mach6-full-workflow 70`
2. Agent instance (with high thinking model) is created with a `/mach6-issue 70` prompt (non-deterministic)
3. New agent instance (with high thinking model) is created with a `/mach6-plan 70` prompt (non-deterministic) (stores output PR number in workflow state)
4. git worktree is created for output PR number (deterministic, may require a git crate dependency)
5. New agent instance (with implement model) is created with prompt `mach6-implement <pr number>` (non-deterministic)
6. Optional pipeline loop - deterministic step of waiting for GH checks to succeed (probably uses some kind of github/gitlab dependency)
7. New high thinking agent instance is created with prompt `/code-review <pr number>` (non-deterministic)
8. agent session from (5) is spun up again to address PR comment left by (7) (non-deterministic)
9. another pipeline loop (deterministic)
10. User is prompted to do human review - human is free to make comments and go for another round of (8, 9), or they can merge the PR
11. If (10) decided to merge, a non-deterministic cleanup step occurs (cleans up git worktree and branch locally)

- we would need to add dependencies to work with github/gitlab and execute local git commands
- Would need a workflow abstraction that is capable of executing multiple sub-agents with differing model configurations

Fizzizist:
## Issue Assessment

### Summary

The proposal is to introduce **hard-coded workflows**: orchestrated sequences that interleave deterministic steps (git worktree creation, polling CI, cleanup) with non-deterministic sub-agent steps (each driven by a slash-command-style skill prompt like `/mach6-issue 70`). Workflows would be triggered by a new slash command (e.g. `/mach6-full-workflow 70`) and need the ability to spawn multiple sub-`Agent` instances with **different model configurations** (e.g. a "high thinking" planning model vs. a cheaper "implement" model), pass state between steps (e.g. PR number from plan stage into implement stage), and pause for human review before terminal actions like merging.

### Current state

[...]

PR #125, opened 2026-04-17
Workflow foundations: model registry, agent spawning, slash-command dispatch

### LINKED ISSUE #121: Workflow foundations: model registry, agent spawning, slash-command dispatch (mach6 workflows PR 1)
## Summary

Foundational refactor enabling multi-agent workflows (parent issue #120). Introduces a **named model registry**, refactors backend selection so multiple backends/models can be constructed at runtime, builds a real **slash-command dispatch table**, and exposes a way to **spawn a fresh `Agent` programmatically** with a chosen model role and tool config.

Nothing user-visible changes by default — this PR ships infrastructure that subsequent workflow PRs build on. The existing single-agent REPL must continue to work unchanged.

## Current Behavior

- `src/backend/mod.rs::from_config` is one-shot at startup; returns a single `BackendSelection { backend, model }`.
- `src/config.rs` has exactly one `[vertex]` and one `[zai]` block — one model per backend.
- Slash commands (`/sessions`, `/model`) are inline string matches in `src/frontend/tui/tui_app.rs`. No dispatch table or argument parser.
- `Agent` construction is hard-coded in `src/main.rs`; there's no helper for spawning a sub-agent with a different model/tool config.

### Named model registry in config

```toml
backend = "vertex"

[models.default]
backend = "vertex"
model   = "claude-sonnet-4-20250514"

[models.thinking]
backend = "vertex"
model   = "claude-opus-4-20250514"

[models.implement]
backend = "zai"
model   = "glm-5.1"
```

- Existing single-model `[vertex]` / `[zai]` config keeps working (back-compat: synthesize a `default` role from it if `[models.*]` is absent).
- A model role is `(backend_name, model_name)` plus any backend-specific overrides.

### Backend factory

- Refactor `from_config` into a `BackendFactory` (or similar) that can, given a model role name, return a fresh `Box<dyn LlmBackend>` + model string. Auth/clients should be cached/shared where cheap (e.g. one Vertex `gcp_auth` provider).

### Slash-command dispatch

- New `src/frontend/tui/commands.rs` (or similar) with a `SlashCommand` trait and a registry. Migrate `/sessions` and `/model` to use it.
- Parser handles `/<name> [args...]`. Unknown commands surface a friendly error in the conversation area.

### Programmatic agent spawning

- A `spawn_agent(model_role, tools_config, session)` helper (free function or `AgentFactory` method) that returns a configured `Agent`, suitable for use by future workflow code.

## Acceptance Criteria

- [ ] Config supports `[models.<role>]` blocks; old single-backend config still loads (back-compat test).
- [ ] `BackendFactory::for_role("name")` returns a working backend + model; errors clearly when role is undefined.
- [ ] Slash commands dispatched via a registry; `/sessions` and `/model` behave identically to today (existing snapshot tests pass).
- [ ] Public helper exists to spawn a new `Agent` with an arbitrary model role and `ToolsConfig`, sharing backend auth where applicable.
- [ ] Existing REPL and single-shot modes work unchanged.
- [ ] All new code has behavior tests (TDD per CLAUDE.md). No `.unwrap()`. `cargo fmt` + `cargo clippy` clean.

## Technical Notes

- Relevant files: `src/config.rs`, `src/backend/mod.rs`, `src/main.rs`, `src/agent.rs`, `src/frontend/tui/tui_app.rs`.
- Slash-command refactor is the lowest-risk piece — could be its own sub-PR if it turns out the model-registry refactor balloons.
- Vertex auth via `gcp_auth` is async and not cheap to set up; share a single auth provider across roles that use the same project/region.
- Don't add git/GitHub deps in this PR — those belong in sub-issue 3.

## Context

- Parent: #120
- Related: #110 (`/model` command, already shipped), #71 (JSON Schema structured output — relevant to PR 2's structured output protocol).

Fizzizist:
Closes #121

Foundational refactor enabling multi-agent workflows (parent #120). Introduces a named model registry, a backend factory that shares expensive auth state across roles, a slash-command dispatch table, and a programmatic helper to spawn fresh `Agent` instances with a chosen model role and tool config.

Nothing user-visible changes by default — this PR ships infrastructure that subsequent workflow PRs build on. The existing single-agent REPL and single-shot modes continue to work unchanged.


Fizzizist:
## Implementation Plan

### Analysis

This is **PR 1 of 4** in the multi-agent workflow umbrella (#120). Pure infrastructure — no user-visible behavior changes. Three intertwined refactors plus one new helper:

1. **Named model registry** in config (with back-compat for the existing `[vertex]` / `[zai]` blocks).
2. **Backend factory** that can construct fresh backends per role on demand, sharing expensive auth state (Vertex `gcp_auth` provider) across roles that target the same project/region.
3. **Slash-command dispatch table** to replace the inline `if text.trim() == "/sessions"` / `strip_prefix("/model ")` chain in `tui_app.rs`.
4. **Programmatic `Agent` spawning** helper, returning a configured `Agent` for an arbitrary role + `ToolsConfig`.

The hardest constraint is **back-compat**: the current single-backend config flow (and `cargo run` with no `[models.*]` block) must keep working unchanged. The simplest path is: on load, if no `[models.*]` is present, synthesize a `default` role from the existing `backend = "..."` + `[vertex]` / `[zai]` block.

### Deliverables

1. **`AppConfig::models`** field — `BTreeMap<String, ModelRole>` populated from `[models.<role>]` TOML blocks. Loader synthesizes a `default` role from legacy `[vertex]`/`[zai]` config when `[models.*]` is absent.
2. **`ModelRole` struct** — `{ backend: String, model: String, /* room for future overrides */ }`. Resolution against the appropriate `[vertex]` / `[zai]` block produces a concrete `(backend_name, model_name, region/project/api_key)`.
3. **`BackendFactory`** in `src/backend/mod.rs` — owns shared expensive resources (currently: a `HashMap<(project, region), Arc<dyn TokenProvider>>` for Vertex). Public API: `for_role(&str) -> Result<BackendSelection>`. Old `from_config` becomes a thin wrapper (or is removed and `main.rs` uses the factory directly with role `"default"`).
4. **Slash-command dispatch** in `src/frontend/tui/commands.rs` — `SlashCommand` trait + `CommandRegistry`. `/sessions` and `/model` migrate to it. **Unknown slash commands pass through to the LLM as normal user messages** so the agent can invoke skills (e.g. `/code-review`) or respond as it sees fit. Dispatch returns a `DispatchResult { Handled, Passthrough }` so `tui_app.rs` treats unknown slash input and plain text identically.
5. **`spawn_agent` helper** — free function (in `src/agent.rs`) taking `(&BackendFactory, &str /* role */, &ToolsConfig, Session, ToolRegistry)` → `Result<Agent>`. Mirrors the construction in `main.rs`.
6. **Updated `main.rs`** — uses `BackendFactory` + `spawn_agent("default", ...)` so the factory path is exercised end-to-end by every run.
7. **Tests** for every new piece (see Testing Approach).

### Acceptance Criteria

- [ ] `[models.<role>]` blocks parse correctly; multiple roles can be defined, each binding to a `backend` + `model`.
- [ ] Legacy config (no `[models.*]`) loads and exposes a `default` role equivalent to the old `(backend, model)` selection — verified by a back-compat test that loads an unmodified pre-foundations config.
- [ ] `BackendFactory::for_role("name")` returns a working `BackendSelection`; returns a clear error when role is undefined or its backend is misconfigured.
- [ ] Two roles that target the same Vertex `(project, region)` share a single auth provider instance (verifiable via `Arc::ptr_eq` on the cache map).
- [ ] `/sessions` and `/model` continue to behave identically — existing `sessions_command_transitions_app_to_session_picker_state`, `model_command_updates_agent_model`, `model_command_empty_name_does_not_update_model` tests pass unchanged after migration to the dispatch table.
- [ ] Unknown slash commands (`/code-review`, `/bogus`) are passed through to the LLM as normal user messages, allowing the agent to invoke skills or respond as it sees fit.
- [ ] `spawn_agent(...)` produces an `Agent` configured with the requested role's model, the supplied `ToolsConfig` (max_tool_iterations, confirmation), and the supplied tool registry.
- [ ] `cargo run` and `cargo run -- --single-shot "hi"` work unchanged.
- [ ] No `.unwrap()` introduced. `cargo fmt` + `cargo clippy` clean. All tests green.

### Files to Create or Modify

**Create:**
- `src/frontend/tui/commands.rs` — `SlashCommand` trait, `CommandRegistry`, `parse_command(&str)`, `DispatchResult` enum, built-in `SessionsCommand` / `ModelCommand` implementations. The trait deliberately encapsulates execution (mutating `App`, talking to `Agent`) so future commands like `/spawn` can plug in.

**Modify:**
- `src/config.rs` — add `ModelRole` struct, `models: BTreeMap<String, ModelRole>` on `AppConfig`, post-load normalization that synthesizes `default` from legacy fields if `models` is empty. Update `validate` to confirm the active role's underlying backend is configured. Add a `resolve_role(&str) -> Result<ResolvedRole>` helper that returns the concrete `(backend_name, model_name)` for a given role.
- `src/backend/mod.rs` — introduce `BackendFactory` with internal cache (`HashMap<(String, String), Arc<dyn TokenProvider>>` for Vertex, plus reqwest clients if cheap to share). Keep `from_config` as a thin wrapper that delegates to `BackendFactory::for_role("default")`, or replace its callsite in `main.rs` and remove it.
- `src/backend/vertex.rs` — add a `VertexBackend::with_auth(project, region, auth: Arc<dyn TokenProvider>)` constructor so the factory can inject a shared provider. Keep existing `new` for tests / direct construction.
- `src/agent.rs` — add `spawn_agent` helper.
- `src/frontend/tui/mod.rs` — register the new `commands` submodule.
- `src/frontend/tui/tui_app.rs` — replace the inline `if/else if` slash-command branch with a call into the dispatch table. Plain text and unknown slash commands both fall through to `submit_message`. Existing snapshot/behavior tests should require no changes.
- `src/main.rs` — switch to `BackendFactory` + `spawn_agent("default", ...)`.

### Testing Approach

Per CLAUDE.md, TDD with behavior tests. New tests:

**`tests/config_test.rs`** (or a new `tests/model_registry_test.rs`):
- `legacy_config_without_models_section_synthesizes_default_role` — loads pre-foundations TOML, asserts `config.models["default"]` exists and matches the legacy backend/model.
- `models_section_parses_multiple_roles` — TOML with `[models.thinking]` + `[models.implement]` produces both roles.
- `resolve_role_returns_concrete_backend_and_model_for_named_role`.
- `resolve_role_errors_for_undefined_role`.
- `validate_errors_when_role_references_unconfigured_backend` (e.g. role uses `backend = "zai"` but no `[zai]` block).

**`src/backend/mod.rs` unit tests:**
- `backend_factory_returns_backend_for_default_role` (using a mock backend variant or a config that bypasses real auth — for Vertex, we can test via a feature-gated injection or test the auth-cache logic with a fake `TokenProvider`).
- `backend_factory_caches_vertex_auth_provider_per_project_region` — registers two roles sharing `(project, region)`, asserts the underlying auth provider is the same `Arc` (use `Arc::ptr_eq`).
- `backend_factory_errors_for_unknown_role`.

**`src/frontend/tui/commands.rs` unit tests:**
- `parse_command_returns_name_and_args_for_slash_input`.
- `parse_command_returns_none_for_non_slash_input`.
- `command_registry_dispatches_known_command_and_returns_handled`.
- `command_registry_returns_passthrough_for_unknown_slash_input` (e.g. `/code-review 42`).
- `command_registry_returns_passthrough_for_non_slash_input`.
- `sessions_command_opens_session_picker` (behavioral, mutates an `App`).
- `model_command_with_name_updates_agent_and_app_model`.
- `model_command_without_name_pushes_usage_error_to_conversation`.

**`src/agent.rs` (or `agent_factory` test module):**
- `spawn_agent_produces_agent_with_role_model_and_tools_config` — uses an in-process mock `BackendFactory` + mock backend, asserts the constructed `Agent.model() == role.model` and that `max_tool_iterations` reflects the supplied `ToolsConfig`.

**Existing tests that must keep passing unchanged:**
- All `tests/config_test.rs` tests (back-compat).
- `sessions_command_transitions_app_to_session_picker_state`, `model_command_updates_agent_model`, `model_command_empty_name_does_not_update_model` (slash-command behavior preserved).
- All TUI snapshot tests under `src/frontend/tui/snapshots/` and `tests/snapshots/`.

### Risks and Open Questions

- **Vertex auth caching key.** `gcp_auth::provider()` reads ambient ADC and isn't parameterized by project/region, so technically all Vertex roles could share one provider regardless of project. Keying the cache on `(project, region)` is more conservative and forward-compatible if we later add per-project SAs. Decision: key on `(project, region)`.
- **`SlashCommand` trait shape.** It needs mutable access to `App` and (for `/model`) to `Agent`. Cleanest is `async fn execute(&self, ctx: &mut CommandContext)` where `CommandContext` bundles `&mut App`, `Arc<Agent>`, `&AppConfig`. Will refine during implementation.
- **`BackendFactory` lifetime.** Currently `Agent` owns `Box<dyn LlmBackend>`. The factory hands out fresh boxed backends per call, which fits the existing ownership model — sub-agents get their own backend instance, only the auth provider is shared.
- **Should `from_config` be removed?** It has no other callers besides `main.rs`. Plan: replace its single callsite, remove the function. Keeps the surface small and forces every code path through the factory.
- **Where does `spawn_agent` live?** Free function in `src/agent.rs` to avoid module proliferation; will reassess if it grows past ~50 LOC.
- **Out of scope (deferred to PR 2/3/4):** Workflow trait, structured output between agents, sub-agent persistence, git/GitHub deps, the actual `/mach6-full-workflow` command.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implemented all 7 deliverables from the plan:

**Deliverable 1-2 — ModelRole + AppConfig.models:**
- Added `ModelRole` struct (`backend`, `model`) and `ResolvedRole` to `src/config.rs`
- Added `models: BTreeMap<String, ModelRole>` to `AppConfig`
- `load_config_from_path` synthesizes a `default` role from legacy `[vertex]`/`[zai]` fields when no `[models.*]` blocks are present (back-compat)
- `AppConfig::resolve_role(&str) -> Result<ResolvedRole>` helper added
- `validate` extended to check named roles reference configured backends
- All config structs now derive `Clone`

**Deliverable 3 — BackendFactory:**
- Added `BackendFactory` to `src/backend/mod.rs` with internal `HashMap<(project, region), Arc<dyn TokenProvider>>` auth cache
- `for_role(&str) -> Result<BackendSelection>` builds backends on demand, sharing Vertex auth providers per `(project, region)`
- `VertexBackend::with_auth` constructor added to `src/backend/vertex.rs`
- `from_config` kept as thin wrapper for back-compat

**Deliverable 4 — Slash-command dispatch:**
- Created `src/frontend/tui/commands.rs` with `SlashCommand` trait, `CommandRegistry`, `ParsedCommand`, `DispatchResult` enum
- Built-in `SessionsCommand` and `ModelCommand` migrate the existing inline handlers
- Unknown slash commands return `Passthrough` → treated identically to plain text, forwarded to LLM
- `tui_app.rs` Enter handler replaced with single `cmd_registry.dispatch()` call

**Deliverable 5 — spawn_agent helper:**
- `spawn_agent(&BackendFactory, &str, &ToolsConfig, Session, ToolRegistry) -> Result<Agent>` in `src/agent.rs`

**Deliverable 6 — Updated main.rs:**
- `BackendFactory::new` + `spawn_agent("default", ...)` replaces the old `from_config` + `Agent::new` flow

**Tests added:**
- Config: back-compat synthesis, multi-role parsing, resolve_role, validate role errors
- Backend: factory errors for unknown role, auth cache deduplication
- Commands: parse_command, registry dispatch (handled/passthrough), SessionsCommand, ModelCommand behavioral tests
- All 3 pre-existing macOS test failures unchanged; 404 tests pass

**Commit:** `eb3388b`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

### Critical
None blocking — but several findings below should land before this PR is built upon by PR 2/3/4.

### Important

**Finding 1 — `vertex_auth_for` has a TOCTOU race that defeats the cache under concurrency.**
The cache is read with the lock, lock dropped, `gcp_auth::provider().await` called, lock retaken to insert. Two concurrent callers for the same `(project, region)` will both miss, both invoke `gcp_auth::provider()`, and both insert. They each get *different* `Arc`s, so they do not in fact share a provider. The whole point of this cache is to be safe under concurrent sub-agent spawning, which is the use case PR2/3/4 enables. Either hold an async mutex across the await, or use a `OnceCell`-per-key pattern (e.g. `tokio::sync::OnceCell` inside a `Mutex<HashMap<Key, Arc<OnceCell<Arc<dyn TokenProvider>>>>>`). Today this race is invisible because `main.rs` only spawns one agent.

**Finding 2 — `spawn_agent` ships with zero tests.**
The plan explicitly listed `spawn_agent_produces_agent_with_role_model_and_tools_config` as a required test. It is not present. `spawn_agent` is the headline new helper for the entire workflow umbrella (#120) — shipping it untested is exactly the wrong place to skip TDD. The argument "we'd need a mock factory" is true, and the answer is to build a mock factory (or split `for_role` so tests can inject a `BackendSelection` directly). Per CLAUDE.md: "TDD will be the main way that code is implemented in this repo, so most code should have tests that test behavior."

**Finding 3 — `backend_factory_caches_vertex_auth_provider_per_project_region` does not test what its name claims.**
The test reaches into the private `vertex_auth_cache` field and inserts twice into a `HashMap`, then asserts `HashMap::len() == 1` and `Arc::ptr_eq` on values it just stored. This is a test of `std::collections::HashMap`, not of `BackendFactory`. The actual cache logic in `vertex_auth_for` (the early-return on cache hit) is never executed. Acceptance criterion "Two roles that target the same Vertex (project, region) share a single auth provider instance" is not actually verified. Combine with finding 1: the test would still pass even with the race-condition bug.

**Finding 4 — `legacy_config_without_models_section_synthesizes_default_role` re-implements production logic inside the test.**
```rust
let mut config = config;
if config.models.is_empty() {
    config.models.insert("default".to_string(), ModelRole { ... });
}
```
The test runs its own copy of the synthesis logic and asserts against that. The production `load_config_from_path` path is not exercised. If someone deletes the synthesis block in `load_config_from_path` tomorrow, this test still passes. Extract synthesis into `AppConfig::normalize_back_compat(&mut self)` (or equivalent) and have both production and test call it — or have the test write a TOML to a temp file and call `load_config_from_path`.

### Suggestions

**Finding 5 — `backend::from_config` is now dead code with a misleading doc-comment.**
No callers anywhere (`main.rs` was migrated to `BackendFactory`). The doc-comment claims it "delegates to `BackendFactory`" but it duplicates the construction code. Plan said: "Should `from_config` be removed? It has no other callers besides `main.rs`. Plan: replace its single callsite, remove the function. Keeps the surface small and forces every code path through the factory." Either delete it or actually have it delegate. Right now it's a footgun: future code might call it and silently bypass the auth-sharing factory.

**Finding 6 — `SlashCommand::execute` uses hand-rolled `Pin<Box<dyn Future>>` instead of `#[async_trait]`.**
The codebase already uses `async-trait` (`LlmBackend`, `Tool`, `gcp_auth::TokenProvider`). The trait signature
```rust
fn execute<'a>(&self, args: &str, ctx: &'a mut CommandContext<'_>)
    -> std::pin::Pin<Box<dyn std::future::Future<Output = Result<DispatchResult>> + 'a>>;
```
should be
```rust
#[async_trait]
async fn execute(&self, args: &str, ctx: &mut CommandContext<'_>) -> Result<DispatchResult>;
```
Less ceremony, consistent with the rest of the project, easier for the next person to add a command.

**Finding 7 — Role validation for `zai` is laxer than the top-level zai check.**
The top-level `zai` validation rejects empty `api_key`. The new role validation only checks `config.zai.is_none()`. So a `[models.implement] backend = "zai"` role passes validation against an empty-api-key `[zai]` block, then fails at the first request. Mirror the top-level check, or factor both into a single `validate_backend_present(&self, backend: &str)` helper.

**Finding 8 — Vertex auth cache key is dead today.**
Every `for_role("vertex", ...)` call reads `self.config.vertex.project` / `self.config.vertex.region` (not from the role) — so the cache key is constant across all roles. The cache will only ever hold one entry. This is fine *if* per-role overrides come later, but right now it's untested forward-compat scaffolding. Either (a) add `project`/`region` overrides to `ModelRole` (the plan called this out: "room for future overrides"), or (b) document loud-and-clear that the cache is sized for future per-role-project Vertex configs and currently only ever has one entry. Right now finding 3 (the fake test) hides this entirely.

**Finding 9 — `tempfile::TempDir::new().expect(..).keep()` leaks temp dirs every test run.**
`.keep()` consumes the `TempDir` and returns the `PathBuf` *without* the destructor — so each `cargo test` invocation accumulates session SQLite databases in `$TMPDIR`. Use `let dir = TempDir::new()?; let path = dir.path().to_path_buf(); /* keep dir alive */`. Five tests in `commands.rs` do this.

**Finding 10 — `FakeBackend` in `commands.rs` duplicates `MockBackend` in `backend/mod.rs`.**
Both implement `LlmBackend` for an empty stream. Promote one to a `pub(crate)` test-support module so the duplication doesn't multiply when PR 2/3/4 add their own fake backends.

**Finding 11 — `DEFAULT_MAX_TOKENS` placement.**
`pub(crate) const DEFAULT_MAX_TOKENS` lives in `agent.rs`, but it's a property of `RequestConfig`. Belongs in `types.rs` (or `config.rs`). Minor, but it'll get harder to relocate once more callers depend on the path.

**Finding 12 — Doc comment lies on `from_config`.**
Says "delegates to `BackendFactory`" but the body is independent. Either fix the comment or fix the function.

### Strengths

- Back-compat is real: legacy single-backend configs still load and the existing `validate_*` tests pass with one-line additions for `models: BTreeMap::new()`.
- Slash-command dispatch shape is good: `SlashCommand` trait + `Passthrough` semantics is clean and forwards unknown slashes to the LLM as plain text — exactly the right call for skill-style commands.
- `tui_app.rs` Enter handler dropped from ~30 lines of inline `if/else` to a single `dispatch` call. This is the kind of refactor that makes the next PR easy.
- All existing TUI snapshot tests pass unchanged — strong evidence the slash-command migration preserved behavior.
- `with_auth` constructor on `VertexBackend` is the right factoring; keeps `new` for direct construction and gives the factory a clean injection point.
- No `.unwrap()` in production paths.
- 868 tests pass on macOS (the progress comment's "3 pre-existing failures" claim appears stale — clean run here).

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

### Classifications

|---|---|---|
| 1: vertex_auth_for race | genuine | Re-read confirms lock-drop-await-relock pattern. Two concurrent callers race, both `await` the slow `gcp_auth::provider()`, both insert. Last-writer-wins on the `HashMap`, but each caller has already received its own `Arc`. The cache exists *for* concurrency; broken under concurrency = broken. PR 2/3/4 will spawn sub-agents in parallel. |
| 2: no spawn_agent test | genuine | Plan listed it explicitly; implementation skipped it. The function is small, but it's the public seam for the next three PRs. Untested public API on a critical path. |
| 3: fake cache test | genuine | The test reaches around `vertex_auth_for` entirely and asserts on `HashMap` semantics. The acceptance criterion ("share a single auth provider instance") is not verified. Worse: it gives a false sense of coverage. |
| 4: test re-implements synthesis | genuine | Test passes regardless of whether `load_config_from_path` does the synthesis. Trivial fix: extract synthesis into a method, call from both. |
| 5: from_config dead code | genuine | Plan said remove; PR kept. No callers. Doc-comment is wrong. Either delete or actually delegate. Low risk to fix. |
| 6: hand-rolled Pin<Box<Future>> | nitpick | Works, just inconsistent with `async-trait` used elsewhere. Style preference. Doesn't affect correctness. |
| 7: zai role validation laxer than top-level | genuine | Real inconsistency — a role can pass validation against an empty-api-key zai block and fail at runtime. Small but real. |
| 8: vertex cache key dead today | deferred | The cache code is correct *if* per-role overrides land. The plan explicitly leaves "room for future overrides" in `ModelRole`. Reasonable to defer to whichever PR adds those overrides — but should be documented in code so it isn't surprising. |
| 9: tempdir leak via .keep() | genuine | Real (tests accumulate state on disk) but very low priority. Easy fix. |
| 10: FakeBackend duplicates MockBackend | nitpick | Will compound when more tests need fake backends. Worth doing eventually, not blocking. |
| 11: DEFAULT_MAX_TOKENS placement | nitpick | Doesn't affect behavior. |
| 12: doc-comment on from_config | nitpick | Subsumed by finding 5. Resolved when 5 is. |

### Action Plan

Before merge (genuine issues, ordered by impact):

1. **Finding 1** — fix the `vertex_auth_for` race (use `tokio::sync::OnceCell` per key, or hold an `async` mutex across the await). This is the only finding that affects correctness in the future workflow PRs.
2. **Finding 3** — replace the fake cache test with a real one that exercises `vertex_auth_for` with a fake `TokenProvider` factory. With the race fix from finding 1, the cache shape changes anyway; rewrite the test against the new shape.
3. **Finding 2** — add `spawn_agent_produces_agent_with_role_model_and_tools_config` per the plan. May require introducing a `BackendFactory` test seam (e.g. `for_selection(BackendSelection)` so tests can inject a `MockBackend` directly without going through `gcp_auth`).
4. **Finding 4** — extract back-compat synthesis into `AppConfig::normalize_back_compat(&mut self)` (or post-deserialization hook), call it from `load_config_from_path`, rewrite the test to exercise `load_config_from_path` against a temp file.
5. **Finding 5** — delete `backend::from_config` (and its misleading doc-comment, finding 12). One-liner.
6. **Finding 7** — mirror the top-level `api_key.is_empty()` check in role validation. Factor into a `validate_zai_configured()` helper if it gets repeated again.
7. **Finding 9** — replace `.keep()` with normal `TempDir` lifetime in the five `commands.rs` tests.

Defer (with code comment):

- **Finding 8** — add a `// TODO: per-role project/region overrides` next to the cache-key construction so the next reader understands why the cache exists.

Follow-up (no need to block this PR):

- **Findings 6, 10, 11** — style/consolidation, do when convenient.

---
*Assessment by mach6*

Fizzizist:
## Progress Update

Addressed all genuine findings from the review:

**Finding 1 — TOCTOU race in vertex_auth_for fixed.**
Replaced the lock-drop-await-relock pattern with a `Mutex<HashMap<Key, Arc<OnceCell<Arc<dyn TokenProvider>>>>>` shape. `OnceCell::get_or_try_init` serialises initialisation — concurrent callers for the same `(project, region)` all await the same cell and receive the same `Arc`.

**Finding 2 — spawn_agent now has a behavior test.**
Added `spawn_agent_produces_agent_with_role_model_and_tools_config`. Introduced `spawn_agent_with_selection` as a public seam that accepts a pre-built `BackendSelection` (bypassing auth), keeping `spawn_agent` as a thin wrapper over it. Test uses `SequencedBackend` and asserts model, max_tool_iterations, and confirmation_mode.

**Finding 3 — cache test now exercises the real vertex_auth_for.**
Test pre-seeds the `OnceCell` with a fake provider (no real `gcp_auth` call), then calls `vertex_auth_for` twice and asserts both results are the same `Arc` via `Arc::ptr_eq`.

**Finding 4 — synthesis extracted to AppConfig::normalize_back_compat.**
Both `load_config_from_path` and tests call the same method. Legacy test rewritten to go through `load_config_from_path` with a temp file. Added `normalize_back_compat_is_idempotent_when_models_already_present` test.

**Finding 5 — from_config deleted.**
No callers remained after main.rs migration; removed entirely.

**Finding 7 — zai role validation now mirrors top-level check.**
Role validation for `zai` backend now also rejects empty `api_key`. Added `validate_errors_when_role_references_zai_with_empty_api_key` regression test.

**Finding 9 — TempDir leaks fixed in commands.rs.**
All 5 tests replaced `.keep()` with `dir.path().to_path_buf()`; `TempDir` destructor now runs at end of test, cleaning up session DBs.

Also added TODO comment on the cache key (finding 8 / deferred).

**Commit:** `51e6b8c`

---
*Progress tracked by mach6*

PR #129, opened 2026-04-18
Add semantic code search tool

### LINKED ISSUE #128: Add semantic code search tool
## Summary

Add a `search` tool to the agent's tool registry that provides embedding-based semantic code search over the project's codebase. Modeled on `@dreb/semantic-search` from the [dreb](https://github.com/aebrer/dreb) project, which uses tree-sitter AST chunking, sentence-transformer embeddings, and a 6-signal POEM ranking algorithm.

## Current Behavior

The agent explores codebases via `bash` (running `grep`, `find`, etc.), `edit_file`, `read_file`, `write_file`, and `skill` tools. There is no dedicated search capability — the agent must manually construct shell commands for every code navigation task, which is slow, brittle, and limited to exact keyword/pattern matching.

## Proposed Behavior

A new `search` tool is registered in the `ToolRegistry` alongside existing tools. It accepts a natural language, identifier, or path query and returns ranked code chunks from the project. The tool builds an incremental index on first use (using a separate Turso database) and serves fast queries thereafter.

## Acceptance Criteria

- [ ] `search` tool is registered and appears in tool definitions sent to the LLM
- [ ] Tool accepts `query` (required), `limit` (optional, default 20), `restrictToDir` (optional), and `rebuild` (optional) parameters
- [ ] Search index stored in a **separate Turso database** (reuses the existing `turso` crate, no new database dependency)
- [ ] Index stored at `<project_root>/.search-index/search.db`; `.search-index/` added to `.gitignore` suggestions in docs
- [ ] Indexes source code files using AST-aware chunking (tree-sitter for Rust, TypeScript, Python, Go, Java, C/C++)
- [ ] Indexes text files (Markdown, YAML, TOML, JSON, plaintext) with format-aware chunking
- [ ] Generates embeddings locally (no external API calls) using a sentence-transformer model (e.g., `all-MiniLM-L6-v2`) via ONNX Runtime
- [ ] Ranks results using multi-signal fusion: BM25 (FTS5), cosine similarity, path match, symbol match, import graph proximity, git recency
- [ ] Uses POEM (Pareto-Optimal Embedded Modeling) for ranking with query-type-adaptive weights (identifier vs. natural language vs. path)
- [ ] Incremental updates on subsequent searches — only re-indexes changed files
- [ ] Reads the agent's `sandbox_root` config to determine which directory to index
- [ ] Respects `.gitignore` rules when scanning
- [ ] First search builds index with progress reporting; subsequent searches are fast
- [ ] Unit tests cover: query classification, chunking, sanitization, tool registration, result formatting
- [ ] Integration test: tool executes against a temp directory and returns ranked results
- [ ] Documentation updated: CLAUDE.md architecture section, config.toml template

### Reference Implementation

`/tmp/dreb/packages/semantic-search/` — TypeScript library with the same design. Key files:

| File | Purpose |
|---|---|
| `search.ts` | Orchestrator: build index → compute metrics → rank → results |
| `scanner.ts` | File discovery, `.gitignore`-aware walking, file type classification |
| `chunker.ts` | Dispatches to tree-sitter or text chunker based on file type |
| `tree-sitter-chunker.ts` | AST-aware chunking for code (functions, classes, methods, etc.) |
| `text-chunker.ts` | Format-aware chunking for Markdown (headings), YAML/TOML (keys), plaintext (paragraphs) |
| `embedder.ts` | Runs `all-MiniLM-L6-v2` via `@huggingface/transformers` (ONNX/WASM) |
| `vector-store.ts` | Cosine similarity, top-K selection over stored vectors |
| `poem.ts` | POEM/TFPR ranking: Pareto dominance with query-type column duplication |
| `query-classifier.ts` | Classifies queries as `identifier`, `natural_language`, or `path_like` |
| `db.ts` | SQLite schema, CRUD for files/chunks/embeddings/symbols/imports |
| `mcp-server.ts` | MCP stdio server adapter |

### Database: Turso

The project already depends on the `turso` crate for session persistence. The search index should use a **separate Turso database file** at `<project_root>/.search-index/search.db`. Turso is highly compatible with SQLite syntax including FTS5 (required for BM25), so all the same table designs and queries from the reference implementation apply directly. No new database dependency needed.

### Tool Integration

- New `src/tools/search.rs` implementing the `Tool` trait from `src/tools/mod.rs`
- Registered in `main.rs` alongside existing tools (`bash`, `edit_file`, `write_file`, `skill`)
- The `Tool::execute` trait is synchronous; indexing/embedding is I/O-bound. Options: (1) spawn a blocking tokio task inside execute, or (2) extend the trait with an async variant. The agent already runs tool execution inside a tokio task.

### Embedding Model

- `ort` (ONNX Runtime Rust bindings) + `tokenizers` crate to run `all-MiniLM-L6-v2` locally (~23MB, cached to `~/.cache/illustrious-manager/models/`)
- Alternative: `candle` (Hugging Face Rust ML framework) for pure-Rust inference

### Tree-sitter

- `tree-sitter` + language-specific crates (`tree-sitter-rust`, `tree-sitter-typescript`, `tree-sitter-python`, `tree-sitter-go`, `tree-sitter-java`, `tree-sitter-c`, `tree-sitter-cpp`)
- Consider feature flags to avoid bloating default compile times

### Config

Add optional `[search]` section to `config.toml`:

```toml
# [search]
# model_cache_dir = "~/.cache/illustrious-manager/models"
# index_dir = ".search-index"  # relative to project root
# visible_dirs = []            # additional dirs to index (bypasses .gitignore)
```

### Architecture Fit

Per the four-layer design:

| Layer | Impact |
|---|---|
| **Tools Layer** | New `search.rs` module implementing `Tool` trait |
| **Agent Core** | No changes — tool is invoked through existing agentic loop |
| **Backend Layer** | No changes — search is a local tool, not an LLM backend |
| **Frontend Layer** | No changes — `AgentEvent::ToolResult` already handles rendering |

## Context

- Reference repo: https://github.com/aebrer/dreb (`packages/semantic-search/`)
- POEM paper: https://iopscience.iopernce.org/article/10.1088/2632-2153/ab891b
- Turso SQL compatibility: https://docs.turso.tech/sql-reference/compatibility

---
*Created by mach6*

Fizzizist:
Closes #128

Adds a `search` tool to the agent's tool registry that provides embedding-based semantic code search over the project's codebase. Uses tree-sitter AST chunking, sentence-transformer embeddings via ONNX Runtime, and a 6-signal POEM ranking algorithm.


Fizzizist:
[stale]

Fizzizist:
[stale]

Fizzizist:
[stale]

Fizzizist:
[resolved]

Fizzizist:
[resolved]

Fizzizist:
[stale]

Fizzizist:
## Code Review

### Critical

**Finding 1: `get_or_create_engine` clones `SearchEngine` out of the Mutex guard, defeating the purpose of caching (confidence: 95)**

`src/tools/search.rs` lines 28–34:
```rust
fn get_or_create_engine(&self) -> SearchEngine {
    let mut guard = self.engine.lock().expect("SearchTool engine lock poisoned");
    if guard.is_none() {
        *guard = Some(SearchEngine::new(self.sandbox_root.clone()));
    }
    guard
        .clone()
        .expect("engine should be Some after initialization")
}
```

The `engine` field is `Arc<Mutex<Option<SearchEngine>>>` — presumably to lazily initialize and cache the `SearchEngine`. But `get_or_create_engine` calls `guard.clone()` which clones the inner `Option<SearchEngine>` and returns it *by value*, releasing the lock immediately. The caller in `execute()` then calls `eng.search(...)` on this local copy. The cached `Option<SearchEngine>` inside the Mutex is never used for anything after initialization — every call creates a fresh clone (which is cheap since `SearchEngine` is just two `PathBuf`s), but the entire `Arc<Mutex<>>` machinery is pointless indirection.

Either:
- Remove the `Arc<Mutex<Option<SearchEngine>>>` entirely and just construct a `SearchEngine` on each call (it is two `PathBuf`s — the actual work happens in `search()`), **or**
- Keep the cached engine and call `.search()` while holding the lock, so the engine is actually reused.

As written, the code is misleading: the caching structure suggests persistent state, but there is no persistent state.

### Important

**Finding 2: `rebuild` path has a TOCTOU race — deletes DB files then takes a separate lock to clear the engine (confidence: 90)**

`src/tools/search.rs` lines 81–86:
```rust
if rebuild {
    self.delete_index_db();
    let mut guard = self.engine.lock().expect("SearchTool engine lock poisoned");
    *guard = None;
}
```

Then immediately after (outside the `if rebuild` block):
```rust
let eng = self.get_or_create_engine();
```

Between `self.delete_index_db()` and the engine being recreated, if another thread were to call `execute()`, it could open the partially-deleted database. While the `Tool` trait `execute` method is synchronous and the `ToolRegistry` is not `Sync` (no `RwLock`), the `Arc<Mutex<>>` on the engine suggests the author intended thread safety. Either the locking is unnecessary (see finding 1) or this race is real.

**Finding 3: `SearchTool` is not `Clone` and uses `Arc<Mutex<>>` but `ToolRegistry` is single-threaded (confidence: 90)**

`ToolRegistry` uses `HashMap<String, Box<dyn Tool>>` — no `RwLock`, no interior mutability. The registry is consumed by `Agent::with_tools()` and accessed from a single async task. The `Arc<Mutex<Option<SearchEngine>>>` in `SearchTool` adds complexity for zero benefit. A simple `Option<SearchEngine>` field (or no field at all — see finding 1) would suffice and be clearer.

**Finding 4: CLAUDE.md architecture section is wildly inaccurate for this PR (confidence: 95)**

The CLAUDE.md diff adds a sprawling description of the search subsystem:
> *Semantic Search (`src/tools/search/`) — Embedding-based code search tool. Uses tree-sitter AST chunking (feature-flagged per language), sentence-transformer embeddings via ONNX Runtime (`all-MiniLM-L6-v2`), and a 6-signal POEM ranking algorithm (BM25/FTS5, cosine similarity, path match, symbol match, import graph, git recency). Search index stored in a separate `rusqlite` database at `<project_root>/.search-index/search.db`. Submodules: `scanner` (file discovery), `chunker` (dispatch), `text_chunker` (Markdown/YAML/TOML/JSON/plaintext), `ts_chunker` (AST-aware for Rust/TypeScript/Python/Go/Java/C/C++), `embedder` (ONNX Runtime), `db` (schema + CRUD), `metrics` (scoring signals), `ranker` (POEM), `query_classifier`, `vector_store`, `format`, `engine` (orchestrator).*

None of these submodules exist in `src/tools/search/`. The PR uses the external `search-semantically` crate, which is a single `search.rs` file with 196 lines. The CLAUDE.md description was presumably written when the plan was to embed the entire search subsystem. Now it describes architecture that does not exist. The description should accurately reflect that the tool delegates to the `search-semantically` crate.

### Suggestions

**Finding 5: `delete_index_db` silently swallows errors from `std::fs::remove_file` (confidence: 85)**

The three `let _ = std::fs::remove_file(...)` calls silently ignore permissions errors, I/O errors, etc. If the DB is locked by another process or permissions are wrong, the rebuild silently does nothing and the user gets stale results with no indication. At minimum, log the error. Ideally, bubble up a meaningful error if none of the files could be removed.

**Finding 6: Tests are superficial — no test exercises the actual search workflow end-to-end (confidence: 88)**

All 7 tests verify structural properties (name, description, schema, registration, is_not_write_tool). None exercise `execute()` with a real or mock query against a populated temp directory. The `rebuild_deletes_index_db` test only tests the helper method, not the `rebuild=true` execution path. The `search_tool_requires_query_field` test is the only one that calls `execute()`, and it only checks the error case. There should be at least one integration test that creates a temp directory with files, calls `execute()` with a query, and verifies the result contains the expected file.

**Finding 7: Issue acceptance criteria not fully met — deviated from issue in undocumented ways (confidence: 85)**

The issue specifies:
- "Search index stored in a separate Turso database (reuses the existing turso crate)" — the PR uses `rusqlite` via `search-semantically`, not `turso`. This is arguably better (avoids mixing concerns) but the deviation is undocumented.
- "Indexes source code files using AST-aware chunking (tree-sitter for Rust, TypeScript, Python, Go, Java, C/C++)" — the crate only bundles `tree-sitter-rust` by default; other languages require feature flags.
- No `[search]` config section was added.
- No integration test.

The PR description says "Closes #128" but does not acknowledge or justify the deviations from the issue.

### Strengths

- Clean, minimal integration surface. The `search.rs` is 196 lines and delegates all complexity to `search-semantically`. This is a huge improvement over embedding thousands of lines of search infrastructure.
- The tool schema, description, and parameter handling are well-designed and consistent with existing tools.
- The `rebuild` feature is a practical escape hatch.
- `.search-index/` added to `.gitignore`.
- The switch to the external crate was the right call. Reusing a published crate beats maintaining an embedded fork.

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

### Classifications

|---|---|---|
| Finding 1: Mutex/caching is pointless | **Genuine issue** | The `Arc<Mutex<Option<SearchEngine>>>` is pure dead weight. `SearchEngine` is two `PathBuf`s; cloning it is trivially cheap, and it holds no reusable state between calls. The caching structure misleads readers into thinking there is shared mutable state. This should be simplified. |
| Finding 2: TOCTOU race on rebuild | **False positive** | The `ToolRegistry` is single-threaded. The `execute()` method is synchronous, called from a single async task. The `Arc<Mutex<>>` is unnecessary (see finding 1), and the "race" cannot manifest in practice because there is no concurrency on this code path. |
| Finding 3: Arc<Mutex<>> unnecessary in single-threaded context | **Genuine issue** | Same root cause as finding 1. The `Arc<Mutex<>>` adds complexity for no benefit. Merge with finding 1 — just remove it. |
| Finding 4: CLAUDE.md describes non-existent architecture | **Genuine issue** | The CLAUDE.md description lists submodules (`scanner`, `chunker`, `ts_chunker`, `embedder`, `db`, `metrics`, `ranker`, etc.) that do not exist in this codebase. It describes the internal architecture of the `search-semantically` crate as if it were local code. This is actively misleading for anyone reading the architecture docs. Must be corrected to accurately describe the integration. |
| Finding 5: delete_index_db swallows errors | **Nitpick** | The rebuild feature is an escape hatch, not a critical path. If files cannot be deleted, the next `SearchEngine::new()` + `search()` will recreate them anyway. Logging would be nice but this is not a correctness issue. |
| Finding 6: Tests are superficial | **Genuine issue** | The tests only verify structural properties. There is zero coverage of the actual `execute()` happy path. Even a simple integration test against a temp directory with a `.rs` file would catch regressions. The `search-semantically` crate has its own tests, but the integration layer (`SearchTool::execute`) is completely untested for the success case. |
| Finding 7: Issue deviations undocumented | **Nitpick** | The deviations are reasonable and the PR is better for them (external crate > embedded code). The CLAUDE.md inaccuracy (finding 4) is the real problem — the issue description itself was used verbatim as architecture docs. A brief note in the PR description would have been sufficient. |

### Action Plan

1. **Remove `Arc<Mutex<Option<SearchEngine>>>`** — Replace with either a plain `Option<SearchEngine>` (if lazy init is desired) or no field at all (just construct a `SearchEngine` in `execute()`). This is a pure simplification with no behavioral change.
2. **Fix CLAUDE.md** — Rewrite the Semantic Search section to accurately describe the tool as a thin wrapper around the `search-semantically` crate, not a listing of internal submodules that do not exist locally.
3. **Add at least one integration test** — Create a temp directory with a source file, call `SearchTool::execute()` with a query, and assert the result contains the file. This tests the full wiring: parameter extraction → engine creation → search → result formatting.
4. **(Optional) Consider adding an error log** in `delete_index_db` if files cannot be removed.

---
*Assessment by mach6*

PR #141, opened 2026-04-21
feat: task management tools

### LINKED ISSUE #136: task management tools
## Summary

Add four task management tools (`create_task`, `list_tasks`, `update_task`, `delete_task`) that allow the agent to maintain a task list per session. Tasks are persisted in a new `task` table on the session's existing turso SQLite database.

## Background

Issue #137 (PR #138, commit `010029a`) made `Tool::execute` async. This unblocks tools that need async I/O — including these task tools, which need async access to the turso `Connection` on the session DB.

## Current Behavior

- The session DB schema (`src/session.rs`, `const SCHEMA`) creates only a `conversation` table.
- No tool currently has access to the session DB; all existing tools are stateless or operate on the filesystem / network.
- The `Session` struct exposes a `pub conn: Connection` field already, so a connection handle can be shared.

### New `task` table

Added to the session DB schema, created idempotently (`CREATE TABLE IF NOT EXISTS`) on every `Session::new` call so that existing session DBs gain the table on reopen.

```sql
CREATE TABLE IF NOT EXISTS task (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    title TEXT NOT NULL,
    description TEXT,
    status TEXT NOT NULL DEFAULT 'pending',  -- pending | in_progress | completed
    created_at INTEGER NOT NULL,             -- unix epoch seconds
    updated_at INTEGER NOT NULL
);
```

### Four new tools (in `src/tools/task.rs`)

| Tool          | Input                                                         | Behavior                                                  |
|---------------|---------------------------------------------------------------|-----------------------------------------------------------|
| `create_task` | `{ title: string, description?: string }`                     | Insert row, return new task `id`.                         |
| `list_tasks`  | `{ status?: "pending" \| "in_progress" \| "completed" }`      | Return all tasks, optionally filtered by status.          |
| `update_task` | `{ id: integer, title?, description?, status? }`              | Partial update; at least one mutable field required.      |
| `delete_task` | `{ id: integer }`                                             | Hard delete the row.                                      |

All four are non-write tools (`is_write_tool() == false`) — they mutate session-local state, not the user's filesystem, so they should not require confirmation under `WriteOnly` mode.

### Session-DB access from tools

Task tools take an `Arc<TokioMutex<Session>>` at construction. The `Agent` already owns exactly such an `Arc`; we share the same `Arc` between the agent and the task tools so that when the agent swaps sessions via `load_session`, the task tools automatically see the new session's DB.

Inside each tool's `async fn execute`, the tool acquires the session lock, gets `&session.conn`, and runs async turso queries.

## Acceptance Criteria

- [ ] `task` table is created on `Session::new` for both new and existing session DBs (idempotent DDL).
- [ ] `create_task`, `list_tasks`, `update_task`, `delete_task` are implemented as four distinct tools in `src/tools/task.rs`.
- [ ] Each tool implements the async `Tool` trait with a JSON schema, name, and description suitable for the LLM.
- [ ] Tools are wired up in `src/main.rs` with the same `Arc<TokioMutex<Session>>` the agent uses, so session swaps work transparently.
- [ ] `update_task` supports partial updates; missing fields are left unchanged.
- [ ] `update_task` and `delete_task` return a clear error (`ToolError::Execution`) when the task ID does not exist.
- [ ] `list_tasks` returns tasks ordered by `id ASC` (creation order).
- [ ] Tasks persist across session resume (verified by test).
- [ ] Tasks are scoped per-session (verified by test: tasks in session A do not appear in session B).
- [ ] Session swap via `Agent::load_session` causes subsequent task tool calls to operate on the new session's DB (verified by test).
- [ ] No `.unwrap()` introduced; `cargo clippy` and `cargo fmt` pass cleanly.
- [ ] Tests follow TDD and cover: schema creation on existing DBs, each tool's happy path, partial update, missing-ID errors, status filtering, per-session isolation, and session swap.

### Files to change

- `src/session.rs` — extend `SCHEMA` to include the `task` table; ensure DDL runs on every `Session::new` (not only when `needs_migration`). Adjust the existing migration short-circuit accordingly.
- `src/tools/task.rs` — new file containing all four tools.
- `src/tools/mod.rs` — add `pub mod task;`.
- `src/main.rs` — construct task tools after the `Session` is created, sharing the same `Arc<TokioMutex<Session>>` with the agent. Refactor `Agent::new` / `spawn_agent` to accept `Arc<TokioMutex<Session>>` instead of `Session` so the `Arc` can be built in `main.rs` first.
- `src/agent.rs` — adjust `Agent::new` / `spawn_agent` signature to accept `Arc<TokioMutex<Session>>` instead of `Session` (small refactor; the agent already wraps it internally).

### Schema migration consideration

The current `Session::new` short-circuits with `if !needs_migration { return ... }` *before* running the `SCHEMA` DDL. This must be removed (or reworked) so that idempotent `CREATE TABLE IF NOT EXISTS` statements always execute. Existing `conversation` table won't be affected (idempotent), and existing sessions will gain the new `task` table on next open.

### Concurrency

The session `TokioMutex` is already held briefly during conversation persistence. Task tool calls add another lock site. Keep the critical section small (acquire lock → run query → drop lock) to avoid contention with the agent's persistence writes. There is no risk of cross-tool deadlock because no tool currently re-enters the agent.

### Testing pattern

```rust
#[tokio::test]
async fn create_task_persists_to_session_db() {
    let dir = TempDir::new().expect("temp dir");
    let session = Session::new(None, dir.path().to_path_buf()).await.expect("session");
    let session_arc = Arc::new(TokioMutex::new(session));
    let tool = CreateTaskTool::new(Arc::clone(&session_arc));
    let result = tool.execute(json!({"title": "do the thing"})).await.expect("ok");
    // assert id returned, row exists in DB
}
```

For session-swap tests, build two sessions, install one in the `Arc<Mutex<Session>>`, run `create_task`, swap the other in, run `list_tasks`, and confirm the second list is empty.

## Out of Scope

- TUI surfacing of tasks (this issue keeps tasks as backend state observable only via tool output).
- Soft-delete / archival (hard delete is sufficient for v1).
- Priorities, due dates, ordering beyond creation order.
- Cross-session task views (tasks remain per-session).

Fizzizist:
Closes #136

Adds four async tools (`create_task`, `list_tasks`, `update_task`, `delete_task`) backed by a new `task` table on the per-session turso SQLite DB. Tools share an `Arc<TokioMutex<Session>>` with the agent so session swaps route transparently.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -413,13 +413,16 @@ async fn execute_tool_calls(
     let mut tool_result_blocks: Vec<ContentBlock> = vec![];
 
     for call in tool_calls {
-        let (input, parse_error) = match serde_json::from_str::<serde_json::Value>(&call.input_json)
-        {
-            Ok(v) => (v, None),
-            Err(e) => (
-                serde_json::Value::Null,
-                Some(format!("Invalid tool input JSON: {e}")),
-            ),
+        let (input, parse_error) = if call.input_json.trim().is_empty() {
+            (serde_json::Value::Object(serde_json::Map::new()), None)
+        } else {
+            match serde_json::from_str::<serde_json::Value>(&call.input_json) {
+                Ok(v) => (v, None),
+                Err(e) => (
+                    serde_json::Value::Null,
+                    Some(format!("Invalid tool input JSON: {e}")),
+                ),
+            }
         };
 
         assistant_content.push(ContentBlock::ToolUse {
@@ -1042,6 +1045,44 @@ mod tests {
                 .iter()
                 .any(|e| matches!(e, AgentEvent::ResponseComplete(_))),
             "loop should continue and complete after malformed input"
+        );
+    }
+
+    #[tokio::test]
+    async fn empty_tool_input_json_is_treated_as_empty_object() {
+        // Regression: LLM sends no input for a no-arg tool (e.g. list_tasks).
+        // The SSE stream delivers no ToolUseDelta events, leaving input_json = "".
+        // This must not produce an error — it should be treated as {}.
+        let empty_input_response: Vec<Result<StreamEvent>> = vec![
+            Ok(StreamEvent::ToolUseStart {
+                id: "t1".to_string(),
+                name: "bash".to_string(),
+            }),
+            // No ToolUseDelta — input_json stays empty
+            Ok(StreamEvent::ToolUseDone),
+            Ok(StreamEvent::Done),
+        ];
+        let backend = SequencedBackend::new(vec![empty_input_response, text_response("ok")]);
+        let agent = agent_with_mode(
+            backend,
+            Some(Box::new(EchoTool::new("bash", "echo output"))),
+            ConfirmationMode::Never,
+        )
+        .await;
+
+        let stream = agent
+            .send("list".to_string(), None)
+            .await
+            .expect("send should succeed");
+        let events = collect_events(stream).await;
+
+        assert!(
+            events.iter().any(|e| matches!(
+                e,
+                AgentEvent::ToolResult { content, is_error, .. }
+                    if content == "echo output" && !is_error
+            )),
+            "empty input_json should execute successfully with empty object input"
         );
     }
 
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -147,7 +147,12 @@ impl Agent {
     }
 
     pub async fn session_history(&self) -> Result<Vec<Message>, anyhow::Error> {
-        self.session.lock().await.conversation().load_history().await
+        self.session
+            .lock()
+            .await
+            .conversation()
+            .load_history()
+            .await
     }
 
     /// If the current session has no messages, delete its DB file from disk.
@@ -163,7 +168,11 @@ impl Agent {
     /// Non-persisted context messages (context files, skill definitions) are preserved
     /// at the front of history; only the persisted portion is replaced.
     pub async fn load_session(&self, session: Session) {
-        let new_history = session.conversation().load_history().await.unwrap_or_default();
+        let new_history = session
+            .conversation()
+            .load_history()
+            .await
+            .unwrap_or_default();
         {
             let prefix_len = *self
                 .context_prefix_len
@@ -222,7 +231,12 @@ impl Agent {
         let confirmation_mode = self.confirmation_mode.clone();
         let session = Arc::clone(&self.session);
 
-        self.session.lock().await.conversation().insert_message(&user_msg).await?;
+        self.session
+            .lock()
+            .await
+            .conversation()
+            .insert_message(&user_msg)
+            .await?;
 
         tokio::spawn(async move {
             let mut iterations = 0u32;
@@ -292,7 +306,12 @@ impl Agent {
                                     content: vec![ContentBlock::Text(text_accumulated.clone())],
                                 };
                                 lock(&history_arc).push(partial_msg.clone());
-                                let _ = session.lock().await.conversation().insert_message(&partial_msg).await;
+                                let _ = session
+                                    .lock()
+                                    .await
+                                    .conversation()
+                                    .insert_message(&partial_msg)
+                                    .await;
                             }
                             record_error(&e.to_string(), &history_arc, &session, &event_tx).await;
                             break 'outer;
@@ -310,7 +329,12 @@ impl Agent {
                         content,
                     };
                     lock(&history_arc).push(assistant_msg.clone());
-                    let _ = session.lock().await.conversation().insert_message(&assistant_msg).await;
+                    let _ = session
+                        .lock()
+                        .await
+                        .conversation()
+                        .insert_message(&assistant_msg)
+                        .await;
                     let _ = event_tx.unbounded_send(AgentEvent::ResponseComplete(text_accumulated));
                     break;
                 }
@@ -330,14 +354,24 @@ impl Agent {
                     content: assistant_content,
                 };
                 lock(&history_arc).push(assistant_msg.clone());
-                let _ = session.lock().await.conversation().insert_message(&assistant_msg).await;
+                let _ = session
+                    .lock()
+                    .await
+                    .conversation()
+                    .insert_message(&assistant_msg)
+                    .await;
 
                 let tool_result_msg = Message {
                     role: Role::User,
                     content: tool_result_blocks,
                 };
                 lock(&history_arc).push(tool_result_msg.clone());
-                let _ = session.lock().await.conversation().insert_message(&tool_result_msg).await;
+                let _ = session
+                    .lock()
+                    .await
+                    .conversation()
+                    .insert_message(&tool_result_msg)
+                    .await;
             }
         });
 
@@ -353,7 +387,12 @@ async fn record_error(
 ) {
     let error_user_msg = Message::text(Role::User, format!("[ERROR] {error_msg}"));
     lock(history).push(error_user_msg.clone());
-    let _ = session.lock().await.conversation().insert_message(&error_user_msg).await;
+    let _ = session
+        .lock()
+        .await
+        .conversation()
+        .insert_message(&error_user_msg)
+        .await;
     let _ = event_tx.unbounded_send(AgentEvent::Error(error_msg.to_string()));
 }
 
@@ -1371,7 +1410,8 @@ mod tests {
             .await
             .expect("session a");
         session_a
-            .conversation().insert_message(&Message::text(Role::User, "session a message".to_string()))
+            .conversation()
+            .insert_message(&Message::text(Role::User, "session a message".to_string()))
             .await
             .expect("insert");
 
@@ -1379,7 +1419,8 @@ mod tests {
             .await
             .expect("session b");
         session_b
-            .conversation().insert_message(&Message::text(Role::User, "session b message".to_string()))
+            .conversation()
+            .insert_message(&Message::text(Role::User, "session b message".to_string()))
             .await
             .expect("insert");
 
@@ -1417,7 +1458,8 @@ mod tests {
             .await
             .expect("session a");
         session_a
-            .conversation().insert_message(&Message::text(Role::User, "session a message".to_string()))
+            .conversation()
+            .insert_message(&Message::text(Role::User, "session a message".to_string()))
             .await
             .expect("insert");
 
@@ -1425,7 +1467,8 @@ mod tests {
             .await
             .expect("session b");
         session_b
-            .conversation().insert_message(&Message::text(Role::User, "session b message".to_string()))
+            .conversation()
+            .insert_message(&Message::text(Role::User, "session b message".to_string()))
             .await
             .expect("insert");
 
@@ -1531,7 +1574,8 @@ mod tests {
 
         let session = Session::new(None, dir_path.clone()).await.expect("session");
         session
-            .conversation().insert_message(&Message::text(Role::User, "hello".to_string()))
+            .conversation()
+            .insert_message(&Message::text(Role::User, "hello".to_string()))
             .await
             .expect("insert");
         let db_path = dir_path.join(format!("{}.db", session.id));
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -44,9 +44,14 @@ impl Agent {
     pub async fn new(
         backend: Box<dyn LlmBackend>,
         config: RequestConfig,
-        session: Session,
+        session: Arc<TokioMutex<Session>>,
     ) -> Self {
-        let history: Vec<Message> = session.load_history().await.unwrap_or_default();
+        let history: Vec<Message> = session
+            .lock()
+            .await
+            .load_history()
+            .await
+            .unwrap_or_default();
 
         Self {
             backend: Arc::from(backend),
@@ -56,7 +61,7 @@ impl Agent {
             tools: Arc::new(ToolRegistry::new()),
             max_tool_iterations: 25,
             confirmation_mode: ConfirmationMode::WriteOnly,
-            session: Arc::new(TokioMutex::new(session)),
+            session,
         }
     }
 
@@ -478,7 +483,7 @@ pub async fn spawn_agent(
     factory: &BackendFactory,
     role: &str,
     tool_config: &ToolsConfig,
-    session: Session,
+    session: Arc<TokioMutex<Session>>,
     tools: ToolRegistry,
 ) -> anyhow::Result<Agent> {
     let selection = factory.for_role(role).await?;
@@ -491,7 +496,7 @@ pub async fn spawn_agent(
 pub async fn spawn_agent_with_selection(
     selection: crate::backend::BackendSelection,
     tool_config: &ToolsConfig,
-    session: Session,
+    session: Arc<TokioMutex<Session>>,
     tools: ToolRegistry,
 ) -> anyhow::Result<Agent> {
     let request_config = RequestConfig {
@@ -517,6 +522,10 @@ mod tests {
     async fn test_session() -> Session {
         let dir = tempfile::TempDir::new().expect("temp dir");
         Session::new(None, dir.keep()).await.expect("test session")
+    }
+
+    async fn test_session_arc() -> Arc<TokioMutex<Session>> {
+        Arc::new(TokioMutex::new(test_session().await))
     }
 
     struct SequencedBackend {
@@ -632,7 +641,7 @@ mod tests {
             confirmation: mode,
             ..Default::default()
         };
-        Agent::new(Box::new(backend), config, test_session().await)
+        Agent::new(Box::new(backend), config, test_session_arc().await)
             .await
             .with_tools(registry)
             .with_tool_config(&tool_config)
@@ -743,7 +752,7 @@ mod tests {
             .unbounded_send(ConfirmationResponse::Approved)
             .expect("send approval");
 
-        let agent = Agent::new(Box::new(backend), config, test_session().await)
+        let agent = Agent::new(Box::new(backend), config, test_session_arc().await)
             .await
             .with_tools(registry)
             .with_tool_config(&tool_config);
@@ -787,7 +796,7 @@ mod tests {
             .unbounded_send(ConfirmationResponse::Approved)
             .expect("send approval");
 
-        let agent = Agent::new(Box::new(backend), config, test_session().await)
+        let agent = Agent::new(Box::new(backend), config, test_session_arc().await)
             .await
             .with_tools(registry)
             .with_tool_config(&tool_config);
@@ -832,7 +841,7 @@ mod tests {
             .unbounded_send(ConfirmationResponse::Rejected)
             .expect("send rejection");
 
-        let agent = Agent::new(Box::new(backend), config, test_session().await)
+        let agent = Agent::new(Box::new(backend), config, test_session_arc().await)
             .await
             .with_tools(registry)
             .with_tool_config(&tool_config);
@@ -878,7 +887,7 @@ mod tests {
             ..Default::default()
         };
 
-        let agent = Agent::new(Box::new(backend), config, test_session().await)
+        let agent = Agent::new(Box::new(backend), config, test_session_arc().await)
             .await
             .with_tools(registry)
             .with_tool_config(&tool_config);
@@ -933,7 +942,7 @@ mod tests {
             ..Default::default()
         };
 
-        let agent = Agent::new(Box::new(backend), config, test_session().await)
+        let agent = Agent::new(Box::new(backend), config, test_session_arc().await)
             .await
             .with_tools(registry)
             .with_tool_config(&tool_config);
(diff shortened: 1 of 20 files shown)


Fizzizist:
## Implementation Plan

### Analysis

The work breaks into three layers:

1. **Persistence** (`src/session.rs`) — extend the schema with a `task` table and rework `Session::new` so DDL runs idempotently on every open. The current code short-circuits before DDL when the DB file already exists, so existing sessions would never gain the new table.

2. **Tools** (`src/tools/task.rs`, `src/tools/mod.rs`) — four new async tools sharing an `Arc<TokioMutex<Session>>` so they always operate on the agent's current session, including after a session swap.

3. **Wiring** (`src/agent.rs`, `src/main.rs`) — minor refactor of `Agent::new` / `spawn_agent` / `spawn_agent_with_selection` to accept `Arc<TokioMutex<Session>>` instead of `Session`, so `main.rs` can build the `Arc` once and share it with both the agent and the four task tools.

### Deliverables

- **`src/session.rs`** — Extended `SCHEMA` containing both `conversation` and `task` `CREATE TABLE IF NOT EXISTS` statements. `Session::new` always runs the DDL (drop the `needs_migration` short-circuit, or restructure so DDL runs unconditionally before returning).
- **`src/tools/task.rs`** — New file with four tool structs: `CreateTaskTool`, `ListTasksTool`, `UpdateTaskTool`, `DeleteTaskTool`. Each holds `Arc<TokioMutex<Session>>` and implements the async `Tool` trait. All return `is_write_tool() == false`. Schemas exactly per the issue. Includes a `TaskStatus` enum (see below).
- **`src/tools/mod.rs`** — `pub mod task;`
- **`src/agent.rs`** — `Agent::new`, `spawn_agent`, `spawn_agent_with_selection` accept `Arc<TokioMutex<Session>>` instead of `Session`. Internal `self.session` already wraps it the same way, so the constructor change is mechanical. All test callsites updated via a new `test_session_arc()` helper.
- **`src/main.rs`** — Construct `Session`, wrap in `Arc<TokioMutex<...>>`, register the four task tools with clones of the `Arc`, then pass the same `Arc` into `spawn_agent`.

### `TaskStatus` enum

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
#[serde(rename_all = "snake_case")]
pub enum TaskStatus {
    Pending,
    InProgress,
    Completed,
}
```

- `as_str()` and `FromStr` for SQL round-trips (`pending | in_progress | completed`).
- `Default` returns `Pending`, matching the SQL default.
- All four tools use `TaskStatus` internally; string parsing happens only at the JSON boundary, with parse failures returning `ToolError::InvalidInput`. Single source of truth for valid statuses.

### Acceptance Criteria

- `task` table exists in any session DB after `Session::new` (new file or pre-existing file).
- All four tools registered and discoverable in `ToolRegistry`; each has a sensible name, description, and JSON schema for the LLM.
- `update_task` allows partial updates; missing fields preserved; updates `updated_at` to current epoch seconds.
- `update_task` requires at least one of `title` / `description` / `status`; rejects with `ToolError::InvalidInput` if none supplied.
- `update_task` and `delete_task` return `ToolError::Execution` when ID does not exist.
- `list_tasks` orders by `id ASC`, optionally filtered by `status`.
- Tasks survive a session close+reopen cycle.
- Tasks created in session A do not appear when listing session B.
- After `Agent::load_session`, task tool calls operate on the new session's DB. Achieved by sharing the same `Arc<TokioMutex<Session>>` between the agent and the tools — `load_session` already swaps the inner `Session` via `*self.session.lock().await = session;`, so all `Arc` holders see the new session automatically.
- All four tools return `is_write_tool() == false`.
- `cargo clippy` clean, `cargo fmt` applied, no `.unwrap()` introduced.

### Testing Approach (TDD-first)

**`src/session.rs` tests** (extend existing `mod tests`):
- `task_table_is_created_on_new_session` — open new session, `INSERT INTO task` succeeds.
- `task_table_is_added_on_reopen_of_existing_db` — create session, drop, reopen via `Session::new`, confirm `INSERT INTO task` works (proves DDL runs unconditionally).
- All existing session tests must continue to pass.

**`src/tools/task.rs` tests** (new `mod tests`):
- `task_status_round_trips_through_string` — parse → as_str → parse equality for all variants; rejects invalid strings.
- `create_task_inserts_row_and_returns_id` — happy path.
- `create_task_with_description` — both fields persisted.
- `create_task_without_title_errors` — `ToolError::InvalidInput`.
- `list_tasks_returns_empty_when_none`.
- `list_tasks_returns_all_in_id_order` — three tasks returned in creation order.
- `list_tasks_filters_by_status` — mixed-status tasks, filter returns only matching.
- `list_tasks_invalid_status_filter_errors` — `ToolError::InvalidInput`.
- `update_task_partial_title_only` — title changes, description and status preserved, `updated_at` advanced.
- `update_task_status_transition` — pending → in_progress → completed.
- `update_task_invalid_status_errors`.
- `update_task_with_no_mutable_fields_errors` — `ToolError::InvalidInput`.
- `update_task_missing_id_errors` — `ToolError::Execution`.
- `delete_task_removes_row`.
- `delete_task_missing_id_errors` — `ToolError::Execution`.
- `tasks_persist_across_session_reopen` — create task, drop session, reopen same DB path with same session ID, list returns the task.
- `tasks_are_scoped_per_session` — two sessions, create task in A, list in B is empty.
- `session_swap_routes_subsequent_calls_to_new_session_db` — `Arc<TokioMutex<Session>>` holds session A, register tool, create task; swap inner `Session` to B; `list_tasks` now returns B's tasks.
- `is_write_tool_returns_false_for_all_four` — confirmation policy regression guard.

**`src/agent.rs` tests** — Add `test_session_arc()` helper; update all `Agent::new(..., test_session().await)` callsites; confirm `load_session_replaces_persisted_history` still passes against the refactored signature.

### Files to Create or Modify

| File | Change |
|---|---|
| `src/session.rs` | Extend `SCHEMA`, run DDL unconditionally in `Session::new`, add 2 tests |
| `src/tools/task.rs` | **New file** — `TaskStatus` enum + 4 tools + ~18 tests |
| `src/tools/mod.rs` | Add `pub mod task;` |
| `src/agent.rs` | Change `Session` → `Arc<TokioMutex<Session>>` in `Agent::new`, `spawn_agent`, `spawn_agent_with_selection`; update test callsites |
| `src/main.rs` | Build `Arc<TokioMutex<Session>>`; clone for each task tool registration; pass into `spawn_agent` |

### Risks and Open Questions

1. **`agent.rs` ripple effect** — `Agent::new` is called in 10+ tests. Mechanical update via the new `test_session_arc()` helper keeps churn small.
2. **`update_task` SQL strategy** — Two clean approaches: (a) build dynamic SQL with only the supplied SET clauses (parameterized), or (b) read-modify-write. Default to (a); fall back to (b) if turso parameter binding for a variable column list is awkward.
3. **`turso` `rows_affected()` API** — Need to confirm availability for detecting missing-ID. If unavailable, do a `SELECT COUNT(*) WHERE id = ?` pre-check inside the same lock acquisition.
4. **`list_tasks` output format** — JSON array of task objects in a single `ContentBlock::Text`, pretty-printed for legibility in the conversation.
5. **`status` validation** — Validate at the tool boundary via `TaskStatus::from_str`, not via a SQL CHECK constraint, so error messages route through `ToolError::InvalidInput`.
6. **No TUI changes** — explicitly out of scope per the issue.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implemented all deliverables from the plan:

- **`src/session.rs`** — Extended `SCHEMA` with `task` table (`CREATE TABLE IF NOT EXISTS`). Switched from `conn.execute` to `conn.execute_batch` so multi-statement DDL runs unconditionally on every open (new or existing DB). Added 2 new tests.
- **`src/tools/task/`** — New module split into 5 files to stay within token limits:
  - `status.rs` — `TaskStatus` enum with `Default`, `as_str()`, `FromStr`
  - `create.rs` — `CreateTaskTool` + 4 tests
  - `list.rs` — `ListTasksTool` (optional status filter) + 5 tests
  - `update.rs` — `UpdateTaskTool` (partial updates, `updated_at` advancement) + 6 tests
  - `delete.rs` — `DeleteTaskTool` + 3 tests
  - `mod.rs` — cross-session/session-swap/persistence tests + re-exports
- **`src/tools/mod.rs`** — Added `pub mod task;`
- **`src/agent.rs`** — `Agent::new`, `spawn_agent`, `spawn_agent_with_selection` now accept `Arc<TokioMutex<Session>>` instead of bare `Session`. Added `test_session_arc()` helper; updated all test callsites.
- **`src/main.rs`** — Wraps session in `Arc<TokioMutex<...>>`, clones into all four task tools, passes same arc into `spawn_agent`.
- All external test files (`agent_test.rs`, `agent_context_test.rs`, `context_integration_test.rs`, `tui_test.rs`) updated to use `Arc<TokioMutex<Session>>`.

514 tests passing, `cargo clippy` clean, `cargo fmt` applied.

**Commit:** `f6407ee`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

### Critical

None.

### Important

**Finding 1 — `update_task` performs the UPDATE before checking row existence.**
`src/tools/task/update.rs` lines 124–158.
The implementation runs `UPDATE task SET ... WHERE id = ?N` first, then runs `SELECT COUNT(*) FROM task WHERE id = ?` to determine whether to return `ToolError::Execution`. This inverts the order recommended in your own implementation plan ("do a `SELECT COUNT(*) WHERE id = ?` pre-check inside the same lock acquisition"). Two concrete consequences:

1. A no-op write is issued every time `update_task` is called with a missing id. Cheap on SQLite but still wasted I/O and a wasted journal entry.
2. More importantly, semantically the operation reports failure *after* it has already run. If a future change adds triggers, FTS sync, or audit logging on the `task` table, the missing-id case will fire side effects before erroring out.

`turso::Connection::execute` returns rows-affected directly (`u64`). Use that instead of a follow-up COUNT — one round-trip, no inversion. Same comment applies, less acutely, to `delete_task` (it does check first, but could collapse the check + DELETE into a single `execute` and inspect rows-affected).

**Finding 2 — Duplicated `now_epoch_secs` helper.**
Files: `src/tools/task/create.rs` lines 13–18 and `src/tools/task/update.rs` lines 16–21. Identical function, identical `.expect()` message. Hoist into `src/tools/task/mod.rs` (or `status.rs` since that's already the shared utility module).

### Suggestions

**Finding 3 — JSON schema for `status` does not constrain values via `enum`.**
Files: `src/tools/task/list.rs` line 28, `src/tools/task/update.rs` line 38. The schemas declare `"type": "string"` and rely on prose in `description` to communicate the three valid values. JSON Schema supports `"enum": ["pending", "in_progress", "completed"]` — this gives the LLM a structured constraint instead of a sentence to interpret, and many LLMs respect schema enums more reliably than prose. Tool-side validation via `TaskStatus::from_str` stays as the source of truth.

**Finding 4 — `update_task` does not validate empty `title` / `description`.**
`src/tools/task/update.rs` lines 86–93. `create_task` filters out empty titles with `.filter(|s| !s.is_empty())` and rejects with `InvalidInput`. `update_task` accepts any string including `""`, so an LLM call with `{"id": 1, "title": ""}` will silently overwrite the title with empty. Either mirror the `create_task` validation or accept the inconsistency intentionally and document it. Pick one; current state is accidental.

**Finding 5 — `delete_task` and `update_task` swallow unexpected `DbValue` types from `COUNT(*)`.**
Files: `src/tools/task/delete.rs` lines 78–79, `src/tools/task/update.rs` lines 145–146. Both use `_ => 0` on the COUNT result. `list_tasks` is stricter and returns an `Execution` error for unexpected types from the same kind of query. Be consistent. SQLite COUNT will always be Integer in practice, so this is defensive only — but inconsistent defensiveness is worse than no defensiveness.

**Finding 6 — `update_task_partial_title_only` test sleeps for 1 full second.**
`src/tools/task/update.rs` lines 198–219. The test sleeps a wall-clock second to verify `updated_at` advances. Across CI runs and local `cargo test` invocations this compounds. Two cleaner options: (a) inject a clock function into the tool so the test can control "now", or (b) assert `updated_at >= before` rather than `>` and drop the sleep, accepting that same-second updates are indistinguishable. Option (a) is more rigorous; option (b) is one-line.

**Finding 7 — `is_write_tool_returns_false` tests are testing the trait's default impl, not behavior of the implementor.**
Files: `src/tools/task/create.rs`, `delete.rs`, `list.rs`, `update.rs`, `mod.rs` (5 copies). The `Tool` trait already returns `false` from `is_write_tool` by default, and none of the four tools override it. The tests pass trivially. They serve as a regression guard — that's fine — but the per-file copies are redundant given the consolidated `is_write_tool_returns_false_for_all_four` test in `mod.rs`. Drop the four per-file copies; keep the consolidated one.

### Strengths

- Clean module split (`status` / `create` / `list` / `update` / `delete` / `mod`). Each file under 320 lines, easy to navigate.
- `TaskStatus` enum with `FromStr` keeps string validation at one boundary instead of scattered string-match throughout the four tools. Plan called for it; implementation honored it.
- Schema-DDL change in `Session::new` is the right call — switching to `execute_batch` and dropping the `needs_migration` short-circuit means existing session DBs gain the new table on first reopen, exactly per the issue's acceptance criterion.
- Cross-cutting tests in `tools/task/mod.rs` (`tasks_persist_across_session_reopen`, `tasks_are_scoped_per_session`, `session_swap_routes_subsequent_calls_to_new_session_db`) directly verify the three highest-risk behaviors from the issue.
- `Arc<TokioMutex<Session>>` refactor across `Agent::new` / `spawn_agent` / `spawn_agent_with_selection` is mechanical and clean. Test ripple absorbed via `test_session_arc()` helper.
- No `.unwrap()` introduced in production code. `.expect()` uses are limited to the `SystemTime` round-trip (logically impossible to fail on a sane system).

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

### Classifications

|---|---|---|
| 1: `update_task` runs UPDATE before existence check | genuine | `turso::Connection::execute` returns `Result<u64>` rows-affected (verified in turso 0.5.3 `connection.rs:115`). Current code wastes a write and a COUNT round-trip per call, and inverts the semantic check the plan explicitly called for. Easy fix. |
| 2: Duplicated `now_epoch_secs` helper | genuine | Identical 6-line function with identical `.expect()` message in two files. Trivial DRY fix; hoist to `task/mod.rs` or `task/status.rs`. |
| 3: JSON schema lacks `enum` for `status` | genuine | Schema enums are honored by Anthropic and other LLM tool-call protocols and reduce malformed-input rate. Plan didn't specify this, but it's a correctness hardening for free. |
| 4: `update_task` accepts empty title/description | genuine | `create_task` validates non-empty; `update_task` does not. Inconsistency is a contract bug, not a style nit — an LLM can blank out a task title via update but not via create. |
| 5: `_ => 0` swallows unexpected DbValue from COUNT | nitpick | SQLite `COUNT(*)` always returns Integer. The code is functionally correct. The only complaint is inconsistency with `list_tasks` which is stricter. Not worth blocking merge. |
| 6: 1-second wall-clock sleep in `update_task_partial_title_only` | genuine | Real CI cost. One-line fix to `assert!(updated_at >= before)` removes the sleep entirely with negligible loss of rigor. |
| 7: Redundant per-file `is_write_tool_returns_false` tests | nitpick | Five copies all testing the trait's default impl. Pass trivially. Consolidated test in `mod.rs` already covers all four. Cleanup opportunity, not a correctness issue. |

### Action Plan

Fix in this order before merge:

1. **Finding 1** — Rewrite `update_task` to use `rows_affected` from the single UPDATE call. Apply same simplification to `delete_task` (collapse pre-check + DELETE into one `execute` + rows-affected inspection).
2. **Finding 4** — Decide: validate empty `title` / `description` in `update_task` (mirror `create_task`), OR document that empty strings overwrite. Recommend validation for consistency. Add regression test.
3. **Finding 3** — Add `"enum": ["pending", "in_progress", "completed"]` to the `status` properties in `list.rs` and `update.rs` schemas.
4. **Finding 6** — Replace `tokio::time::sleep(1s)` + `>` with `>=` assertion. Drop the sleep.
5. **Finding 2** — Move `now_epoch_secs` to `src/tools/task/mod.rs` (or `status.rs`); import in `create.rs` and `update.rs`.

Defer (cleanup, not blocking):

6. **Finding 5** — Make COUNT-result type handling consistent across `update.rs`, `delete.rs`, `list.rs`. Either all strict, or all `_ => 0`. Strict matches `list.rs` precedent.
7. **Finding 7** — Remove four per-file `is_write_tool_returns_false` tests; keep the consolidated one in `mod.rs`.

After 1–5 land, this PR is mergeable. After 6–7, it's polished.

---
*Assessment by mach6*

Fizzizist:
## Progress Update

Addressed all 7 review findings:

- **Finding 1** — `update_task`: replaced UPDATE-then-COUNT with single `execute` inspecting `rows_affected` (turso returns `u64`). No more wasted write or inverted check.
- **Finding 2** — Hoisted `now_epoch_secs` to `src/tools/task/mod.rs`; removed duplicate definitions from `create.rs` and `update.rs`.
- **Finding 3** — Added `"enum": ["pending", "in_progress", "completed"]` to `status` field in `list.rs` and `update.rs` JSON schemas.
- **Finding 4** — `update_task` now rejects empty `title` with `ToolError::InvalidInput`. Added regression test `update_task_empty_title_errors`.
- **Finding 5** — `delete_task`: replaced pre-check COUNT + DELETE with single `execute` + `rows_affected`. Removed `_ => 0` swallowing; `delete_task_removes_row` test now uses `panic!` on unexpected type, consistent with `list_tasks`.
- **Finding 6** — Removed 1-second `tokio::time::sleep`; changed `assert!(updated_at > before)` to `assert!(updated_at >= before)`.
- **Finding 7** — Removed four redundant per-file `is_write_tool_returns_false` tests; consolidated one in `mod.rs` retained.

511 tests passing, `cargo clippy` clean, `cargo fmt` applied.

**Commit:** `86a7d26`

---
*Progress tracked by mach6*

Fizzizist:
## Progress Update

Refactored to repository pattern per code review feedback — SQL no longer lives in tool files.

**Structure:**
- `src/session/` — split into a module directory:
  - `mod.rs` — `Session`, `list_sessions`, uuid helpers; delegates to sub-modules; exposes `session.tasks()` accessor
  - `conversation.rs` — all conversation SQL (`insert_message`, `load_history`, `is_empty`, `read_first_user_message`); private to the module
  - `task.rs` — `TaskStatus`, `TaskRecord`, `TaskRepo` with `create`/`list`/`update`/`delete`; all task SQL lives here; publicly exported
- `src/tools/task/` — tools stripped to pure input parsing + `session.tasks().*` calls + output formatting; zero SQL; `status.rs` removed (moved to `session/task.rs`)

**Tests:** 521 passing (added `TaskRepo` unit tests directly in `session/task.rs`), `cargo clippy` clean, `cargo fmt` applied.

**Commit:** `7dfe17a`

---
*Progress tracked by mach6*

Fizzizist:
## Progress Update

Made `ConversationRepo` public and symmetric with `TaskRepo`.

- `session/conversation.rs` — `ConversationRepo` struct with `pub(super) fn new`; all methods public: `insert_message`, `load_history`, `is_empty`, `read_first_user_message`
- `session/mod.rs` — `pub mod conversation`; `pub use conversation::ConversationRepo`; added `session.conversation()` accessor; removed delegation methods (`insert_message`, `load_history`, `is_empty`) from `Session` directly
- All callsites updated: `session.insert_message(...)` → `session.conversation().insert_message(...)` etc. across `agent.rs`, `tui_app.rs`, and `session/mod.rs` tests

521 tests passing, `cargo clippy` clean, `cargo fmt` applied.

**Commit:** `0144020`

---
*Progress tracked by mach6*

Fizzizist:
## Progress Update

Bug fix from manual testing.

**Bug:** Calling `list_tasks` (or any no-arg tool) failed with `Invalid tool input JSON: EOF while parsing a value at line 1 column 0`. Subsequent Vertex AI requests rejected with `messages.N.content.0.tool_use.input: Input should be a valid dictionary` because the persisted `tool_use` block had `"input": null`.

**Root cause:** When the LLM invokes a tool with no arguments, the SSE stream delivers no `ToolUseDelta` events, leaving `input_json` as the empty string. `serde_json::from_str("")` returns EOF — the agent then stored `input: null` for the tool_use block, which Anthropic rejects on the next round-trip because schema requires an object.

**Fix:** `src/agent.rs` — treat an empty/whitespace `input_json` as `{}` before attempting to parse. Real malformed JSON still produces an error.

**Regression test:** `empty_tool_input_json_is_treated_as_empty_object` — sends `ToolUseStart` + `ToolUseDone` with no `ToolUseDelta`; asserts the tool executes successfully.

522 tests passing, `cargo clippy` clean, `cargo fmt` applied.

**Commit:** `3e732d3`

---
*Progress tracked by mach6*

PR #143, opened 2026-04-21
Parallel tool calls within a single LLM response

### LINKED ISSUE #134: parallel tool calls
To set up for running sub-agents in parallel, we should start by allowing tools to be run in parallel, in general. Right now, the agent runs tools one at a time. This is inefficient.

## Acceptance Criteria
- parallel tool calling enabled for all three backends. (check the docs to make sure tool calling is enabled. In some cases you need to send a parameter like `parallel_tool_calls: true`)
- display tool call indices in the UI (`[Tool(1)]` matches `[Result(1)]`)
- allows for future `Agent` tool that can be called in parallel
- 3 parallel bash tool calls of `sleep 3`, `sleep 1` and `sleep 2` should return in 3 seconds instead of 6 (as an example)

Fizzizist:
Closes #134

Foundational change for future sub-agent fan-out: tools called within a single LLM response are now dispatched concurrently while preserving `tool_use` / `tool_result` block ordering required by Anthropic-style backends.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -399,7 +399,7 @@ async fn record_error(
 async fn execute_tool_calls(
     tool_calls: Vec<PendingToolCall>,
     text_prefix: String,
-    tools: &ToolRegistry,
+    tools: &Arc<ToolRegistry>,
     confirmation_mode: &ConfirmationMode,
     confirmation_rx: &mut Option<mpsc::UnboundedReceiver<ConfirmationResponse>>,
     event_tx: &mpsc::UnboundedSender<AgentEvent>,
@@ -409,47 +409,59 @@ async fn execute_tool_calls(
         assistant_content.push(ContentBlock::Text(text_prefix));
     }
 
-    // TODO: tool calls within a single response are independent and could be executed concurrently.
-    let mut tool_result_blocks: Vec<ContentBlock> = vec![];
+    // Parse inputs and pre-build the assistant ToolUse blocks. Emit all
+    // ToolUseReceived events synchronously in input order before any execution.
+    struct ParsedCall {
+        id: String,
+        name: String,
+        input: serde_json::Value,
+        parse_error: Option<String>,
+    }
 
-    for call in tool_calls {
-        let (input, parse_error) = if call.input_json.trim().is_empty() {
-            (serde_json::Value::Object(serde_json::Map::new()), None)
-        } else {
-            match serde_json::from_str::<serde_json::Value>(&call.input_json) {
-                Ok(v) => (v, None),
-                Err(e) => (
-                    serde_json::Value::Null,
-                    Some(format!("Invalid tool input JSON: {e}")),
-                ),
+    let parsed: Vec<ParsedCall> = tool_calls
+        .into_iter()
+        .map(|call| {
+            let (input, parse_error) = if call.input_json.trim().is_empty() {
+                (serde_json::Value::Object(serde_json::Map::new()), None)
+            } else {
+                match serde_json::from_str::<serde_json::Value>(&call.input_json) {
+                    Ok(v) => (v, None),
+                    Err(e) => (
+                        serde_json::Value::Null,
+                        Some(format!("Invalid tool input JSON: {e}")),
+                    ),
+                }
+            };
+            ParsedCall {
+                id: call.id,
+                name: call.name,
+                input,
+                parse_error,
             }
-        };
+        })
+        .collect();
 
+    for call in &parsed {
         assistant_content.push(ContentBlock::ToolUse {
             id: call.id.clone(),
             name: call.name.clone(),
-            input: input.clone(),
+            input: call.input.clone(),
         });
         let _ = event_tx.unbounded_send(AgentEvent::ToolUseReceived {
             id: call.id.clone(),
             name: call.name.clone(),
-            input: input.clone(),
+            input: call.input.clone(),
         });
+    }
 
-        if let Some(err_msg) = parse_error {
-            let _ = event_tx.unbounded_send(AgentEvent::ToolResult {
-                name: call.name.clone(),
-                content: err_msg.clone(),
-                is_error: true,
-            });
-            tool_result_blocks.push(ContentBlock::ToolResult {
-                tool_use_id: call.id,
-                content: err_msg,
-                is_error: true,
-            });
+    // Confirmation gate: walk in input order, request confirmation for each
+    // tool that needs it, collect approvals sequentially.
+    let mut approvals: Vec<bool> = Vec::with_capacity(parsed.len());
+    for call in &parsed {
+        if call.parse_error.is_some() {
+            approvals.push(false);
             continue;
         }
-
         let needs_confirmation = match confirmation_mode {
             ConfirmationMode::Always => true,
             ConfirmationMode::Never => false,
@@ -457,12 +469,11 @@ async fn execute_tool_calls(
                 tools.lookup(&call.name).is_ok_and(|t| t.is_write_tool())
             }
         };
-
         let approved = if needs_confirmation {
             let _ = event_tx.unbounded_send(AgentEvent::ToolConfirmationRequired {
                 id: call.id.clone(),
                 name: call.name.clone(),
-                input: input.clone(),
+                input: call.input.clone(),
             });
             if let Some(rx) = confirmation_rx.as_mut() {
                 matches!(rx.next().await, Some(ConfirmationResponse::Approved))
@@ -472,12 +483,38 @@ async fn execute_tool_calls(
         } else {
             true
         };
+        approvals.push(approved);
+    }
 
-        let (result_content, is_error) = if approved {
-            match tools.lookup(&call.name) {
+    // Dispatch all approved tools concurrently; slot-vector preserves input order.
+    let n = parsed.len();
+    let mut slot_futures = Vec::with_capacity(n);
+
+    for (i, call) in parsed.iter().enumerate() {
+        let tools_clone = Arc::clone(tools);
+        let id = call.id.clone();
+        let name = call.name.clone();
+        let input = call.input.clone();
+        let approved = approvals[i];
+        let parse_error = call.parse_error.clone();
+
+        slot_futures.push(tokio::spawn(async move {
+            if let Some(err_msg) = parse_error {
+                return (i, id, name, err_msg, true);
+            }
+            if !approved {
+                return (
+                    i,
+                    id,
+                    name,
+                    "User declined to execute this tool.".to_string(),
+                    true,
+                );
+            }
+            let (content, is_error) = match tools_clone.lookup(&name) {
                 Ok(tool) => match tool.execute(input).await {
                     Ok(result) => {
-                        let content = result
+                        let text = result
                             .content
                             .iter()
                             .filter_map(|b| {
@@ -489,25 +526,36 @@ async fn execute_tool_calls(
                             })
                             .collect::<Vec<_>>()
                             .join("\n");
-                        (content, result.is_error)
+                        (text, result.is_error)
                     }
                     Err(e) => (e.to_string(), true),
                 },
                 Err(e) => (e.to_string(), true),
-            }
-        } else {
-            ("User declined to execute this tool.".to_string(), true)
-        };
+            };
+            (i, id, name, content, is_error)
+        }));
+    }
 
+    // Collect results preserving input order via slot-vector.
+    let mut slots: Vec<Option<(String, String, String, bool)>> = (0..n).map(|_| None).collect();
+    for fut in slot_futures {
+        if let Ok((i, id, name, content, is_error)) = fut.await {
+            slots[i] = Some((id, name, content, is_error));
+        }
+    }
+
+    let mut tool_result_blocks: Vec<ContentBlock> = Vec::with_capacity(n);
+    for slot in slots.into_iter().flatten() {
+        let (id, name, content, is_error) = slot;
         let _ = event_tx.unbounded_send(AgentEvent::ToolResult {
-            name: call.name.clone(),
-            content: result_content.clone(),
+            id: id.clone(),
+            name,
+            content: content.clone(),
             is_error,
         });
-
         tool_result_blocks.push(ContentBlock::ToolResult {
-            tool_use_id: call.id,
-            content: result_content,
+            tool_use_id: id,
+            content,
             is_error,
         });
     }
@@ -758,7 +806,7 @@ mod tests {
         assert!(
             events.iter().any(|e| matches!(
                 e,
-                AgentEvent::ToolResult { name, content, is_error }
+                AgentEvent::ToolResult { name, content, is_error, .. }
                     if name == "bash" && content == "ls output" && !is_error
             )),
             "expected ToolResult with ls output"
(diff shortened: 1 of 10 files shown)


Fizzizist:
## Implementation Plan

### Analysis

The agent currently executes tool calls from a single LLM response sequentially. `execute_tool_calls` in `src/agent.rs` iterates a `Vec<PendingToolCall>` and awaits each `tool.execute(...)` one at a time. The `Tool` trait is already `Send + Sync` and `ToolRegistry` is held as `Arc`, so the runtime is concurrency-ready.

This is foundational work for future sub-agent fan-out. We are explicitly **not** building sub-agent dispatch in this PR.

### Design decisions (confirmed with user)

- **Confirmation strategy**: serialize prompt, then parallel execute. Walk the batch in input order, request confirmation for each tool needing it, await responses sequentially. Once all approvals are collected, dispatch the approved tools concurrently. Declined tools get the canned "User declined…" result without execution.
- **Concurrency cap**: unbounded `join_all` for v1. A configurable `tools.max_parallel` can be added later if needed.
- **File-conflict policy**: leave to the model for v1. Document the limitation. Shared-state tools (semantic search index, task session mutex) self-serialize via their own internal locks.
- **Cancellation**: no cancel-on-error. All in-flight tools run to completion; errors land in `tool_result` blocks as today.
- **Event ordering**: emit all `ToolUseReceived` events synchronously in input order before spawning execution. Emit `ToolResult` events as each tool completes (arrival order may differ from input order). Add `id: String` to `AgentEvent::ToolResult` so frontends can pair result→use deterministically.

### Deliverables

1. `execute_tool_calls` rewritten to:
   - Pre-build the assistant `ToolUse` blocks and emit all `ToolUseReceived` events in input order.
   - Run the confirmation gate sequentially for any tools that need it.
   - Dispatch all approved tools concurrently via `futures::future::join_all`.
   - Reassemble `tool_result_blocks` in input order using a slot-vector keyed by index.
2. `AgentEvent::ToolResult` gains an `id: String` field (breaking event-enum change, accepted).
3. Frontend renderers (`tui/tui_app.rs`, `stdout.rs`) updated to accept the new field. TUI's pairing logic stays append-only for now; document that future work can match each result back to its tool-use entry visually using the new `id`.
4. New tests proving overlap, ordering, and confirmation behavior.

### Acceptance criteria

- A test using a `SleepyTool` that records start/end timestamps shows two concurrently-dispatched calls overlap (max(end) − min(start) < sum(durations) − epsilon).
- A test with three tool calls returning out-of-order completion times asserts the resulting `Message::User` content has `ToolResult` blocks in the same order as the `ToolUse` blocks in the preceding assistant message.
- Existing confirmation tests still pass; a new test verifies that with three calls under `Always` mode, three `ToolConfirmationRequired` events are emitted in input order and three approvals are consumed before any tool runs.
- A new test verifies that declining one of three tools causes only the declined tool's slot to contain "User declined…" while the other two execute normally.
- All existing `cargo test` passes; `cargo clippy` clean; `cargo fmt` clean.

### Files to create or modify

- `src/agent.rs` — rewrite `execute_tool_calls`; thread `id` through to the `ToolResult` event; add new tests (with a new `SleepyTool` helper alongside `EchoTool`).
- `src/types.rs` — add `id: String` to `AgentEvent::ToolResult`; update unit-test in that file.
- `src/frontend/tui/tui_app.rs` — destructure new `id` field; update existing test event constructions.
- `src/frontend/stdout.rs` — same: destructure new `id`; update test event constructions.
- No `Cargo.toml` changes expected (`futures` already a dep).

### Testing approach

All new tests live in `src/agent.rs`'s existing `#[cfg(test)] mod tests`.

- **`SleepyTool` helper**: parameterized with name, sleep duration, and a shared `Arc<Mutex<Vec<(String, Instant, Instant)>>>` log. Records start before `tokio::time::sleep`, end after.
- **`two_tool_calls_in_one_response_execute_concurrently`**: backend returns one assistant turn with two `tool_use` blocks (each sleepy ~100ms), then a text turn. Assert wall-clock overlap.
- **`tool_results_preserve_input_order_under_concurrent_completion`**: three tools with descending sleep durations (300, 200, 100 ms) — the fastest finishes first. Inspect the `User` message appended to history; assert `tool_use_id`s appear in the original input order.
- **`always_confirmation_with_multiple_calls_prompts_in_order`**: three tool calls, `Always` mode, pre-load three approvals in the channel. Assert three `ToolConfirmationRequired` events arrive in input order; all three execute.
- **`declined_tool_in_batch_does_not_block_others`**: three calls; channel pre-loaded with `Approved, Rejected, Approved`. Assert middle slot is the decline message; other two have real outputs.
- **TUI/stdout snapshot tests**: not strictly required for this change since rendering is per-event and `id` is opaque. We verify they compile and existing snapshots are unchanged.

### Risks and open questions

- **Backend stream interleaving**: `vertex` already accumulates multiple `ToolUseStart/Done` pairs into the `tool_calls` vec. `zai` and `ollama` will be re-read during implementation to confirm they emit the same shape — if any backend coalesces, this PR is a no-op there but still safe.
- **Frontend rendering with out-of-order results**: TUI appends `ToolResult` entries to `conversation` as they arrive. Visually a slow tool's result will appear after a fast tool's result, even if the slow tool's `tool_use` block was rendered first. Acceptable v1 UX (matches how parallel terminals look). Follow-up issue: "group result under matching use" rendering using the new `id`.
- **Sandbox races**: two concurrent `edit_file` calls on the same path are now possible. Documented, not prevented in v1.
- **History/result ordering invariant**: this is the easiest thing to get wrong. The slot-vector reassembly must be covered explicitly by the ordering test.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implemented parallel tool call dispatch within a single LLM response.

**Changes:**
- `src/types.rs` — Added `id: String` to `AgentEvent::ToolResult` (breaking enum change per plan)
- `src/agent.rs` — Rewrote `execute_tool_calls`: pre-emit all `ToolUseReceived` events in input order → sequential confirmation gate → concurrent dispatch via `tokio::spawn` → slot-vector reassembly preserving input order for `tool_result` blocks. Also added 4 new tests: concurrency overlap assertion, ordering under out-of-order completion, always-confirmation prompt ordering, and partial decline in a batch.
- `src/frontend/stdout.rs`, `src/frontend/tui/tui_app.rs`, `src/logging.rs` — Updated all `AgentEvent::ToolResult` pattern matches and test constructions for new `id` field

All tests pass (529+). `cargo clippy` and `cargo fmt` clean.

**Commit:** `2bf8364`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

### Critical

**Finding 1 — `ToolResult` events are emitted in batch, not as tools complete (deviation from plan).**
`execute_tool_calls` collects every spawned task into `slots` via a sequential `for fut in slot_futures { fut.await }` loop, *then* runs a second loop that emits all `AgentEvent::ToolResult` events. Concretely (`src/agent.rs` ~540-563):

```rust
for fut in slot_futures {
    if let Ok((i, id, name, content, is_error)) = fut.await {
        slots[i] = Some((id, name, content, is_error));
    }
}
// ...emit events here, AFTER the slowest task finishes
```

The plan explicitly stated: *"Emit `ToolResult` events as each tool completes (arrival order may differ from input order)"*. The implementation does the opposite — no `ToolResult` event is fired until the slowest tool in the batch finishes. The TUI (`handle_agent_event` in `tui_app.rs:351`) pushes one conversation entry per `ToolResult`, so users now see *zero* results until the whole batch completes, then everything appears at once. That defeats half the value of running things in parallel from a UX standpoint.

Fix: use `FuturesUnordered` (or `select!` over a `JoinSet`) so each completion both fills its slot **and** emits its `ToolResult` event immediately. The slot vector is still used to assemble `tool_result_blocks` in input order at the end. (Severity: high — functional regression vs. documented design.)

**Finding 2 — `JoinError` from a panicked tool task silently drops the slot, breaking the Anthropic ordering invariant.**

```rust
for fut in slot_futures {
    if let Ok((i, id, name, content, is_error)) = fut.await {
        slots[i] = Some((id, name, content, is_error));
    }
}
// ...
for slot in slots.into_iter().flatten() { ... }
```

If a tool task panics (e.g. a future bug in `execute()`), `fut.await` returns `Err(JoinError)`, the slot stays `None`, and `flatten()` silently drops it. The next API call to the backend will then have `n` `tool_use` blocks in the assistant message but only `n-1` matching `tool_result` blocks. Anthropic-style backends reject this with a 400. The plan explicitly called this out: *"this is the easiest thing to get wrong. The slot-vector reassembly must be covered explicitly."*

Fix: handle the `Err` arm by inserting a synthetic error result block (e.g. `is_error: true, content: "tool task panicked: <reason>"`) so the invariant is preserved. (Severity: high.)

### Important

**Finding 3 — Concurrency assertion in `two_tool_calls_in_one_response_execute_concurrently` is borderline.**
Sleep durations: 200 + 50 ms. The assertion is `wall_ms < sum_ms` where `sum_ms = 250`. Sequentially the wall clock is *exactly* ≥ 250 ms, but on a fast machine could be 250-251 ms — passing or failing depending on scheduler jitter. Concurrent should be ~200 ms. The margin between the two cases is only ~50 ms. The plan called for a "generous slack for CI"; the implementation has none. Two cleaner options:

- Tighter wall-clock bound: `assert!(wall_ms < 230)` — clearly excludes sequential (≥250) and clearly admits concurrent (~200).
- Or assert true overlap directly from the timestamps: `start[fast] < end[slow] && start[slow] < end[fast]`. This is the actual definition of "ran concurrently" and immune to wall-clock noise.

The current assertion is technically correct but flaky-prone. (Severity: medium.)

**Finding 4 — `tokio::spawn` was used where the plan specified `futures::future::join_all`.**
Functionally equivalent for correctness, but spawning N tasks adds runtime allocation and forces the futures to be `'static` (which is why every closure captures cloned `String`s and an `Arc<ToolRegistry>`). `join_all`/`FuturesUnordered` over un-spawned futures would do the same job with less ceremony and would interoperate better with the as-they-arrive emission needed to fix Finding 1. Not wrong, but inconsistent with the plan and slightly heavier than necessary. (Severity: low.)

### Suggestions

**Finding 5 — `name` is redundantly threaded through the spawn return tuple.**
The closure captures and returns `name`, only for the collection loop to destructure it back out. Since `parsed[i].name` is already known to the outer scope, the tuple could be `(usize, String /*content*/, bool)` and `name` looked up from `parsed[i]` (or, more cleanly, the slot type could hold only what the closure produces — content/is_error — and id/name come from the index). Minor, but tightens the data flow. (Severity: low.)

**Finding 6 — `id: String` field added to `AgentEvent::ToolResult` has no current consumer.**
TUI, stdout, and logging all destructure with `..`. The plan acknowledges this and says "future work can match each result back to its tool-use entry visually using the new `id`." Acceptable, but right now this field exists purely for a hypothetical future. Worth either (a) tracking the follow-up issue explicitly, or (b) actually using it in the TUI to pair results under their tool-use entry while we're already touching this code. (Severity: low — accepted-by-plan.)

**Finding 7 — `tools: &Arc<ToolRegistry>` parameter could be `tools: Arc<ToolRegistry>` (by value).**
Inside the function we do `Arc::clone(tools)` per call. Passing the `Arc` by value (and cloning at the single call site) is the more idiomatic Rust pattern for share-by-ref-counted-handle. As written it works; just not the conventional shape. (Severity: low / nit.)

### Strengths

- Slot-vector reassembly correctly preserves input order for `tool_result_blocks` (the Anthropic-mandated invariant), and `tool_results_preserve_input_order_under_concurrent_completion` proves it with descending-duration sleeps. This was the highest-risk part of the change and it landed clean.
- Confirmation gate is correctly serialized *before* dispatch — no `confirmation_rx` starvation, prompts arrive in input order, all three new tests covering this pass.
- `declined_tool_in_batch_does_not_block_others` correctly asserts the partial-decline case — exactly the kind of edge case that would silently break in a naïve refactor.
- Frontend / logging touch points (`stdout.rs`, `tui_app.rs`, `logging.rs`) are minimally and correctly updated for the new event field; no behavioural regressions.
- All 529+ existing tests still pass; `cargo fmt` clean.

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

### Classifications

|---|---|---|
| 1 — `ToolResult` events emitted in batch, not as tools complete | **genuine** | Verified by re-reading `src/agent.rs:540-563`. The first loop awaits each `JoinHandle` in spawn order before any event is emitted; the second loop fires events only after every task is collected. TUI pushes one entry per `ToolResult` event, so users see no progress until the slowest tool in the batch finishes. Direct deviation from the plan's stated event-ordering design. |
| 2 — `JoinError` silently drops the slot | **genuine** | Verified: `if let Ok(...)` skips `Err(JoinError)`, slot stays `None`, `flatten()` drops it. Assistant message still has the `ContentBlock::ToolUse`, so the next API call has mismatched `tool_use`/`tool_result` counts → 400 from Anthropic. Plan explicitly flagged this exact failure mode as the highest-risk thing in the change. |
| 3 — Concurrency assertion is borderline (wall < 250) | **genuine** | Sequential lower bound is exactly the assertion's ceiling. On a fast machine the test could pass even when the implementation is sequential (wall=249ms), or fail spuriously when concurrent (wall=251ms due to scheduler noise). Easy fix: tighten to `wall_ms < 220` or assert true overlap from per-tool timestamps. |
| 4 — `tokio::spawn` instead of `join_all` | **nitpick** | Functionally equivalent. The plan said `join_all` but `tokio::spawn` works. Worth noting because fixing Finding 1 is cleaner with `FuturesUnordered` than with `JoinHandle`s, but on its own this isn't a defect. |
| 5 — `name` redundantly threaded through spawn return | **nitpick** | Cosmetic. Doesn't affect correctness. |
| 6 — `id: String` field has no current consumer | **deferred** | Plan acknowledged this is foundational for a follow-up TUI rendering improvement ("group result under matching use"). Real follow-up work, not a defect in this PR. Should be tracked as an issue. |
| 7 — `tools: &Arc<ToolRegistry>` could be by-value | **nitpick** | Idiomatic preference, no behavioural difference. |

### Action Plan

Must fix before merge:

1. **Finding 2 — `JoinError` handling.** Change the collection loop to insert a synthetic error result block when `fut.await` returns `Err`. This protects the Anthropic ordering invariant against any future tool that might panic. Two-line fix.
2. **Finding 1 — emit events as tools complete.** Replace the `Vec<JoinHandle>` + sequential `await` pattern with `FuturesUnordered` (or `tokio::task::JoinSet`). In the collection loop: emit the `ToolResult` event immediately when each result arrives, AND fill `slots[i]`. After the loop, build `tool_result_blocks` from the (now-complete) slot vector in input order. Restores the documented as-they-arrive UX without sacrificing ordering.
3. **Finding 3 — sharpen the concurrency assertion.** Either tighten the wall-clock bound (`< 220ms`) or assert true overlap from the per-tool timestamps in the log. Prevents future flakes and false positives.

Should be tracked separately:

4. **Finding 6 — open a follow-up issue** for "TUI: group `ToolResult` rendering under matching `ToolUse` entry using the new `id` field." Records the intended consumer of the new field so it doesn't sit dead.

Optional polish (nitpicks 4, 5, 7) — leave alone unless touching this code anyway.

---
*Assessment by mach6*

Fizzizist:
## Progress Update

Addressed all three genuine findings from the code review.

**Changes:**
- **Finding 1 — ToolResult events emitted in batch:** Replaced `tokio::spawn` + sequential `await` with `FuturesUnordered`. Each tool completion now emits its `ToolResult` event immediately as it arrives, while the slot-vector still reassembles `tool_result_blocks` in input order at the end. Non-dispatchable cases (parse errors, declined tools, unknown tools) are handled inline and emit their events synchronously.
- **Finding 2 — JoinError silently drops slot:** The fallback `unwrap_or_else` in the slot-vector assembly now synthesizes an error result block for any unfilled slot, preserving the Anthropic tool_use/tool_result count invariant even if a future is somehow dropped.
- **Finding 3 — Borderline concurrency assertion:** Replaced the wall-clock `< sum` assertion with true-overlap checks from per-tool timestamps: `fast_start < slow_end && slow_start < fast_end`. This is the actual definition of "ran concurrently" and is immune to scheduler jitter in CI.

All 586 tests pass. `cargo clippy` clean (pre-existing unrelated warning only). `cargo fmt` clean.

**Commit:** `2a003e3`

---
*Progress tracked by mach6*

Fizzizist:
## Progress Update

Found and fixed the actual root cause of the manual test failure. The session DB showed the model (`glm-5.1` via z.ai) was only ever generating **one** `tool_use` per response, even when explicitly told to issue multiple. The parallel dispatch code in `execute_tool_calls` was never exercised.

**Root cause:** The z.ai and ollama backends use the OpenAI-compatible API format. In that protocol, the `parallel_tool_calls` request parameter controls whether the model may return multiple tool calls in a single response. We were not sending it, so the API defaulted to disallowing parallel tool calls. The Vertex/Anthropic backend doesn't need this because parallel tool use is enabled by default in the Anthropic Messages API.

**Fix:** Added `"parallel_tool_calls": true` to the request body in both `zai.rs` and `ollama.rs` when tools are present. Added test assertions verifying the field is set with tools and absent without.

**Commit:** `7d802b0`

---
*Progress tracked by mach6*

Fizzizist:
## Progress Update

Found the real bug. The session DB proved the model was issuing multiple tool calls but only one was being captured. The root cause was in the stream accumulation loop in `send()`.

**Root cause:** The OpenAI-compatible streaming format (zai, ollama) emits tool calls with `index` fields and a **single** `ToolUseDone` at the end. The accumulation loop used `current_tool: Option<PendingToolCall>` which got silently overwritten each time a new `ToolUseStart` arrived. Only the last tool call survived.

The Vertex/Anthropic backend was unaffected because it emits `ToolUseDone` per tool, creating a clean start→delta→done→start→delta→done sequence.

**Fix:**
- Added `index: Option<u64>` to `StreamEvent::ToolUseStart` and `StreamEvent::ToolUseDelta`
- Zai and ollama parsers now propagate the index; Vertex sets `None`
- The accumulation loop in `send()` routes indexed tools into a `HashMap<u64, PendingToolCall>` and collects them in index order when `ToolUseDone` arrives
- New regression test `indexed_tool_calls_from_openai_backend_accumulate_correctly` proves the fix

Also includes the previous `parallel_tool_calls: true` fix for the request body.

All 588 tests pass. `cargo clippy` clean. `cargo fmt` clean.

**Commit:** `3639302`

---
*Progress tracked by mach6*

PR #145, opened 2026-04-22
feat: parallel tool calls

### LINKED ISSUE #134: parallel tool calls

Fizzizist:
Closes #134

Run independent tool calls from a single assistant turn concurrently, advertise parallel tool use to all three backends, and surface call/result correlation in the UI via per-turn indices.


Fizzizist:
## Implementation Plan

### Problem analysis

`Agent::execute_tool_calls` (`src/agent.rs:399`) runs the `Vec<PendingToolCall>` from a single assistant turn through a sequential `for` loop, despite a TODO already noting they're independent. Three sequential bash sleep calls of 3/1/2s take 6s instead of 3s. We also need to (a) advertise parallel tool use to each backend per its own API contract, and (b) make the UI correlate calls and results by index (`[Tool(1)]` ↔ `[Result(1)]`).

### Deliverables

1. **Concurrent tool execution** in `execute_tool_calls` using `futures::future::join_all` (preserves input order — important for `ContentBlock` ordering in history).
2. **Backend parallel-tool-calls flags**, after verifying current docs:
   - `vertex.rs`: add `tool_choice: { "type": "auto", "disable_parallel_tool_use": false }` (or current Anthropic equivalent) in `build_request_body` when tools are present.
   - `zai.rs`: add `parallel_tool_calls: true` in the OpenAI-compatible body when tools are present.
   - `ollama.rs`: verify whether any flag exists; if not, document the no-op and add a test pinning the request body shape.
3. **Per-turn call index** plumbed through:
   - `AgentEvent::ToolUseReceived { index, .. }`, `AgentEvent::ToolResult { index, .. }`, `AgentEvent::ToolConfirmationRequired { index, .. }` gain a `usize` index.
   - `ConversationEntry` carries an `Option<usize>` rendered as `[Tool(N)]` / `[Result(N)]` (keeps the role enum a pure tag).
   - Index is **per-turn, 1-based**, resets each assistant response. Matches the AC's pairing intuition.
4. **Stdout frontend** (`src/frontend/stdout.rs`) prints `[tool(N): name]` / `[result(N)]` for parity.
5. **Confirmation flow under parallelism**: smallest-change design — request all confirmations sequentially **before** launching the parallel batch. Approved calls then run via `join_all`; declined calls get a synthesized "User declined" `ToolResult` block in the correct slot.
6. **Tests** (TDD-first):
   - **Wall-clock parallelism regression test** in `src/agent.rs`: register a `SleepTool` (test-only) that sleeps for a `duration_ms` arg. Drive a synthetic stream emitting three `tool_use` blocks with `{300, 100, 200}` ms sleeps. Assert total elapsed `< 600ms` (sequential lower bound; parallel ≈ 300ms).
   - **Order preservation test**: same setup, assert resulting `tool_result_blocks` match the `tool_use` ids in input order even though sleep times mean completion order is reversed.
   - **Failure isolation test**: one tool errors, the other two complete normally, all three `ToolResult` events emitted.
   - **Index propagation test**: assert `AgentEvent::ToolUseReceived` and `ToolResult` carry matching per-turn indices `1, 2, 3`.
   - **Confirmation-then-parallel test**: with `ConfirmationMode::Always`, assert all N confirmation events are emitted before any tool starts executing, and a denied tool produces the "declined" result without blocking the others.
   - **Backend body tests**: extend `vertex.rs`, `zai.rs`, `ollama.rs` `build_request_body` tests to assert the parallel-tool-calls field is present (or correctly absent for ollama).
   - **TUI snapshot updates**: `cargo insta review` for any snapshots containing `[Tool]` / `[Result]`. Add at least one new snapshot showing two indexed pairs in the conversation view.
7. **Doc note** in `CLAUDE.md` Agent Core paragraph: tool calls within a single turn run concurrently; confirmations gathered first.

### Files to create or modify

| File | Change |
|---|---|
| `src/agent.rs` | Replace sequential `for` in `execute_tool_calls` with confirmation-gather → `join_all`. Add `index` to events. New tests. |
| `src/types.rs` | Add `index: usize` to the three tool-related `AgentEvent` variants. |
| `src/backend/vertex.rs` | Add `tool_choice` with `disable_parallel_tool_use: false` (verify field name vs. current docs). New body test. |
| `src/backend/zai.rs` | Add `parallel_tool_calls: true` when tools present. New body test. |
| `src/backend/ollama.rs` | After doc verification: add flag if one exists, otherwise body unchanged + comment + test pinning shape. |
| `src/frontend/tui/conversation_area.rs` | Add `Option<usize>` index field on `ConversationEntry`; format `[Tool(N)]` / `[Result(N)]`. Update existing label tests. |
| `src/frontend/tui/tui_app.rs` | Pass index from events into entry construction. |
| `src/frontend/stdout.rs` | Print indexed labels. |
| `src/frontend/tui/snapshots/` | `cargo insta review` regenerated snapshots. |
| `CLAUDE.md` | One-line note in Agent Core layer description. |

### Acceptance criteria

- `cargo test` passes including the new wall-clock test (`< 600ms` for 300+100+200ms sleeps).
- `cargo clippy` clean, `cargo fmt` applied.
- Manual smoke test: ask the live agent (any backend) to run three bash commands in parallel: `sleep 3`, `sleep 1`, `sleep 2` — completes in ≈3s wall clock.
- TUI shows `[Tool(1)]`, `[Tool(2)]`, `[Tool(3)]`, `[Result(1)]`, `[Result(2)]`, `[Result(3)]` with matching indices.
- Each backend's `build_request_body` test asserts the parallel flag is present (or the documented absence for ollama).

### Risks & open questions

1. **Stale API knowledge.** Implementer must check current docs for Anthropic Messages API (`tool_choice.disable_parallel_tool_use`?), z.ai (`parallel_tool_calls`?), and Ollama (`/api/chat` — likely no flag) before writing the backend changes. Tests should pin whatever the docs say.
2. **Confirmation UX.** Sequential confirmation gather is the smallest change but means the user sees N prompts in a row before anything runs. Acceptable for v1; batch-prompt UX is a future enhancement.
3. **Snapshot churn.** Every TUI snapshot containing `[Tool]` / `[Result]` will need re-acceptance via `cargo insta review`.
4. **Wall-clock test flakiness.** Use 100/200/300ms (short) and a generous bound (`< 600ms` — sequential lower bound — rather than `< 350ms`) to survive loaded CI.
5. **Concurrent writes to the same path** become possible (e.g. two `edit_file` calls on the same file). Out of scope for this issue but worth a follow-up.
6. **Sub-agent groundwork** is explicitly deferred per the assessment — this PR is purely the parallelism substrate.

### Estimated scope

Medium-small. ~1 focused PR, 2–4 hours including snapshot review.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Full implementation of parallel tool calls:

**Deliverable 1 — Concurrent execution:** Replaced sequential `for` loop in `execute_tool_calls` with confirmation-gather-first → `join_all`. Results returned in input order regardless of completion order.

**Deliverable 2 — Backend flags:**
- Vertex: `tool_choice: {type: "auto", disable_parallel_tool_use: false}` added when tools present.
- zai: `parallel_tool_calls: true` added when tools present.
- Ollama: no flag (none exists in `/api/chat`); pinning test documents the absence.

**Deliverable 3 — Per-turn 1-based indices:** Added `index: usize` to `ToolUseReceived`, `ToolResult`, `ToolConfirmationRequired` variants in `AgentEvent`.

**Deliverable 4 — TUI labels:** `ConversationEntry` now carries `Option<usize> index`; renders `[Tool(N)]:` / `[Result(N)]:`. New `new_indexed` / `new_with_lines_indexed` constructors. New insta snapshot.

**Deliverable 5 — Stdout labels:** `[tool(N): name]` / `[result(N) from name]` / `[error(N) from name]`.

**Deliverable 6 — Tests (6 new):**
- Wall-clock parallelism regression (300+100+200ms < 600ms sequential bound)
- Input-order preservation (completion order is reversed by sleep times)
- Failure isolation (one error, two succeed, all three results emitted)
- Index propagation (ToolUseReceived and ToolResult carry indices 1,2,3)
- Confirmation-then-parallel (all confirmations before first result; denied tool gets error result)
- Backend body tests for vertex, zai, ollama parallel flag presence/absence

**Deliverable 7 — CLAUDE.md** Agent Core paragraph updated.

**Commit:** `c9801ae`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

Multi-aspect review: code, errors, tests, completeness, simplify.

### Critical
None.

### Important

**Finding 1 — TUI confirmation flow drops the index, rendering `[Tool]` instead of `[Tool(N)]` for confirmed tools.**

`tui_app.rs:398` destructures `AgentEvent::ToolConfirmationRequired { name, input, .. }` and discards `index`. `AppState::ToolConfirmation { name, input }` (line 36) has no index field. When the user approves, `tui_app.rs:611` calls the legacy `tool_use_entry(&name, &input, width)` which routes through `tool_use_entry_indexed(..., 0)` and the `0 → None` mapping produces an unindexed entry.

Effect: every write-tool call (which is the entire point of `ConfirmationMode::WriteOnly`, the default) renders `[Tool]:` in the TUI even though the corresponding `[Result(N)]:` is properly indexed. Acceptance criterion "TUI shows `[Tool(1)]`, `[Tool(2)]`, `[Tool(3)]`, `[Result(1)]`, ..." is violated for the most common production code path.

Fix: thread `index` into `AppState::ToolConfirmation` and pass it to `tool_use_entry_indexed` at line 611. The new snapshot test covers `Never` mode only and so does not catch this.

**Finding 2 — Resumed sessions render unindexed `[Tool]` / `[Result]`.**

`tui_app.rs:124` reconstructs entries from history via `self.tool_use_entry(name, input, ...)` (no index) and the `ToolResult` branch at line 126 calls `ConversationEntry::new(...)` (also no index). After a `--session-id <uuid>` resume or a session swap via the picker, the user sees `[Tool]` / `[Result]` even though the in-memory `index` field exists. AC is partially violated. Two reasonable fixes: (a) re-derive per-turn 1-based indices at load time by walking the message stream and counting `ToolUse` blocks per assistant message (cheap, no schema change); (b) persist the index. (a) is preferable.

**Finding 3 — `unreachable!()` violates "the program should _never_ panic."**

`agent.rs:451` reads `None => unreachable!("parse_error and approval are in sync"),`. CLAUDE.md is unambiguous: "use of `.unwrap()` is forbidden under _all_ circumstances. The program should _never_ panic." `unreachable!()` is `panic!()` with extra steps; the policy intent forbids it. The invariant ("approval is None ⟺ parse_error is Some") is maintained across two separate loops with no compile-time enforcement — exactly the situation the policy exists to prevent. Refactor: zip `resolved` with a `Vec<ToolDecision>` enum (`ParseError(String) | Declined | Approved`) so the closure pattern-matches an enum that statically encodes the intended states.

**Finding 4 — Test coverage gap: parallel execution with text prefix and with confirmation+rejection ordering.**

The new tests cover the streaming-of-tool-only path. None exercises the case where the assistant emits text (`TextDelta`) followed by N `tool_use` blocks under parallelism, which is the realistic Claude Sonnet behavior. `text_prefix` is preserved in `assistant_content` (`agent.rs:409`), but no test asserts that the resulting `Message` has `[Text(prefix), ToolUse, ToolUse, ToolUse]` in that order under parallel execution. Add a regression test: model emits `"working on it"` + 3 sleeps; assert message structure.

### Suggestions

**Finding 5 — `0` as a sentinel for "no index" is in-band signaling.**

`tui_app.rs:191` `tool_use_entry_indexed(name, input, width, index: usize)` uses `if index > 0 { Some(index) } else { None }` to distinguish "indexed" from "unindexed." Since the rest of the code uses 1-based indices, `0` happens to be safe today, but the type system already has `Option<usize>` for exactly this. Take `Option<usize>` as the parameter and delete the legacy `tool_use_entry` wrapper — it has only 5 callers, all in the same file.

**Finding 6 — New clippy warning introduced.**

`cargo clippy --all-targets` on this branch reports `field_reassign_with_default` at `src/agent.rs:1701` (in the new `with_tool_config` test helper usage). The parent commit is clean of this warning. Project policy: "Always run `cargo clippy` before committing code." Either initialize `ToolsConfig` with struct-update syntax (`ToolsConfig { max_tool_iterations: 7, ..Default::default() }`) or `#[allow]` with justification.

**Finding 7 — Wall-clock parallelism test margin is OK but could be tighter.**

`parallel_tool_calls_complete_faster_than_sequential_bound` asserts `elapsed < 600ms` for sleeps of 300+100+200ms. That's *exactly* the sequential lower bound. Parallel should be ~300ms; a tighter `< 450ms` would still leave 150ms (50%) of slack on a loaded CI runner and would catch regressions where execution is "mostly sequential but slightly overlapped." Current bound only catches a fully-sequential regression. Nitpick — current bound is correct and serves the AC, just less sensitive.

**Finding 8 — Two-pass confirmation gathering doubles iteration over `resolved`.**

`agent.rs:459` (gather approvals) and `agent.rs:493` (build futures) both iterate `&resolved`. They could be merged: build `Vec<ToolDecision>` in one pass, then `iter().zip().map().collect()` directly. Minor. Current shape is readable.

**Finding 9 — Failure isolation test doesn't actually exercise concurrent failure.**

`parallel_tool_calls_failure_isolation_all_results_emitted` uses a shared `Mutex<u32>` counter and only fails when count == 2. Three sequential `.await`s on the mutex serialize the tool execution inside the supposedly-parallel block, so this test does not actually verify that an in-flight failure does not abort an in-flight success. Replace the counter with three separate tools (`SuccessTool`, `FailingTool`, `SuccessTool`), or use a `Arc<AtomicU32>::fetch_add` so the bodies don't serialize. Test still passes today because `join_all` correctly handles per-future errors, but the test is not pinning what its name claims.

### Strengths

- TDD discipline: 6 new tests covering parallelism, ordering, failure isolation, index propagation, and confirmation interaction. 517/517 lib tests + 536 total pass.
- Correct use of `join_all` (preserves input order) over `FuturesUnordered` (would not). The plan called this out and the implementation matches.
- The confirmation-gather-first design is the smallest viable change and is explicitly tested.
- Backend body tests pin all three providers' parallel-tool-calls flag presence/absence — including Ollama's documented absence, which is the correct way to encode "we checked, there is no flag."
- `index: usize` (not `Option<usize>`) on the `AgentEvent` variants is the right choice — it's always present in the agent layer; only the UI layer has the "should I show an index" question, and that's where `Option<usize>` lives on `ConversationEntry`.
- New insta snapshot is sensible and human-readable.
- CLAUDE.md doc note is accurate and concise.
- No `.unwrap()` introduced. All `.expect()` usages are in tests with informative messages.

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

Independent re-verification of each finding against the actual source.

### Classifications

|---|---|---|
| 1 — TUI confirmation drops index | **genuine** | Verified at `tui_app.rs:36` (AppState has no index field), line 398 (`..` discards index), line 611 (legacy `tool_use_entry` called → defaults to no index). Every WriteOnly-confirmed tool renders `[Tool]` not `[Tool(N)]` in the TUI. Directly violates AC. |
| 2 — Resumed sessions render unindexed | **genuine** | Verified at `tui_app.rs:122-135`. Reload path calls `tool_use_entry` without an index and `ConversationEntry::new` (not `_indexed`) for results. Re-derive indices at load time by counting ToolUse blocks per assistant message — cheap, no schema change. |
| 3 — `unreachable!()` in agent.rs:451 | **genuine** | Project policy in CLAUDE.md: "the program should _never_ panic." `unreachable!` is `panic!` with metadata. The two-loop structure makes the invariant compile-time-unchecked. Refactor `Vec<Option<bool>>` into `Vec<ToolDecision>` enum. Note: a separate pre-existing `unreachable!()` exists at `tui_app.rs:608`; that's out of scope for this PR. |
| 4 — No test for text+parallel-tools | **deferred** | Re-checked: no test today (nor on parent commit) covers `TextDelta` followed by N `tool_use` blocks. The PR's `text_prefix` handling at agent.rs:409 is correct. Gap is pre-existing and orthogonal to parallelism. Worth a follow-up issue but not a merge blocker for this PR. |
| 5 — `0` sentinel for "no index" | **nitpick** | Stylistic. `Option<usize>` would be cleaner but the current code is correct and the wrapper is well-isolated. Could be addressed alongside Finding 1's fix since they touch adjacent code. |
| 6 — New clippy warning | **genuine** | Confirmed: `field_reassign_with_default` at `agent.rs:1701` is introduced by this PR (verified by checking out parent commit and re-running clippy). Project policy: "Always run `cargo clippy` before committing code." Trivial fix. |
| 7 — Wall-clock test margin tight | **nitpick** | The 600ms bound correctly distinguishes parallel (~300ms) from sequential (~600ms) and serves the AC. A tighter bound would be more sensitive but is not required. |
| 8 — Two-pass over `resolved` | **nitpick** | Stylistic. Current shape is readable. Not worth touching. |
| 9 — Failure isolation test critique | **false positive** | On re-read: the mutex is held only briefly to increment a counter, then dropped before the work begins. The test validates what its name claims ("all results emitted when one fails") and the test passes for the right reason (`join_all` over `Result`-returning futures). My critique was overreach. |

### Action Plan

Order by impact:

1. **Finding 1** (TUI confirmation index lost) — visible UX regression vs the stated AC. Add `index: usize` to `AppState::ToolConfirmation`, plumb through. Add a snapshot/handler test under `ConfirmationMode::WriteOnly` covering an indexed tool entry after approval.
2. **Finding 3** (`unreachable!()`) — policy violation. Refactor approval gathering to produce `Vec<ToolDecision>` with `ParseError(String) | Declined | Approved` so the closure pattern-matches an enum that statically encodes the states.
3. **Finding 2** (resumed-session unindexed) — partial AC violation, but in a less common path. Re-derive per-turn indices when constructing entries in the session-load path. Add a test that loads a session with multiple tool calls in one assistant message and asserts the rendered labels.
4. **Finding 6** (clippy) — trivial. `ToolsConfig { max_tool_iterations: 7, ..Default::default() }` and re-run `cargo clippy --all-targets`.
5. **Finding 5** (sentinel `0`) — fold into the Finding 1 fix while touching `tool_use_entry_indexed`.

Findings 4, 7, 8, 9 are not action items for this PR.

### Counts

- Genuine: 3 (Findings 1, 3, 6)
- Nitpicks: 3 (Findings 5, 7, 8)
- False positives: 1 (Finding 9)
- Deferred: 1 (Finding 4)

---
*Assessment by mach6*

Fizzizist:
## Progress Update

Addressed all genuine review findings:

**Finding 1 + 5 — TUI confirmation index lost / sentinel 0:**
- Added `index: usize` to `AppState::ToolConfirmation`
- `tool_use_entry_indexed` now takes `Option<usize>` directly (no `0` sentinel)
- Removed legacy `tool_use_entry` wrapper; all 5 call sites updated
- Confirmation approval path now passes `Some(index)` to produce `[Tool(N)]` entries
- New test: `tool_confirmation_required_index_is_threaded_to_entry_on_approval`

**Finding 2 — Resumed sessions rendered unindexed:**
- `load_history` now re-derives per-turn 1-based indices by counting `ToolUse`/`ToolResult` blocks per message
- Produces `new_indexed` / `new_with_lines_indexed` entries as appropriate
- New test: `load_history_with_multiple_tool_calls_in_one_turn_renders_indexed_labels`

**Finding 3 — `unreachable!()` policy violation:**
- Introduced `ToolDecision` enum (`ParseError(String) | Declined | Approved`)
- Merged the two-pass (parse + confirm separately) into a single loop that builds `ToolDecision` inline; `join_all` still runs only after all decisions are gathered, preserving the confirmation-before-execution guarantee
- `unreachable!` eliminated; enum statically encodes the valid states

**Finding 6 — Clippy warning:**
- `ToolsConfig { max_tool_iterations: 7, confirmation: ConfirmationMode::Never, ..Default::default() }` struct-update syntax
- Also fixed pre-existing `unnecessary_map_or` warning in zai.rs (`is_none_or`)

**Commit:** `761ffb7`

---
*Progress tracked by mach6*

Fizzizist:
## Progress Update

Renamed `ConversationEntry::index` → `tool_index` in `conversation_area.rs` and all call sites in `tui_app.rs`.

**Commit:** `25144c8`

---
*Progress tracked by mach6*

PR #162, opened 2026-04-27
Add separate subagent token counter to status line

### LINKED ISSUE #150: Add separate subagent token counter to status line

Fizzizist:
## Implementation Plan: Separate Subagent Token Counter on Status Line

Closes #150.

### Analysis

Today, both `AgentEvent::Usage` (parent) and `AgentEvent::SubAgentUsage` (subagent) accumulate into a single `App::usage` field, then render as one `↑in ↓out` block on the status line. The issue requests visual separation: a second token block (green) appears to the right of the parent counter only when subagent tokens have been recorded.

The change is **purely frontend display** — `AgentEvent::SubAgentUsage` already carries the subagent counts; the agent layer needs no changes.

### Deliverables

1. **`src/frontend/tui/status_line.rs`**
   - Add `subagent_usage: Option<&'a TokenUsage>` to `StatusLineInfo`.
   - In `build_status_line()`, when `subagent_usage` is `Some` and has non-zero totals, insert a second styled span (`Color::Green`) with format `↗↑{in} ↓{out}` immediately after the parent token span (still on the left side, before the git branch).
   - When `None` or zero totals, render exactly the current layout (no extra span, no extra spacing).
   - Width handling: if the right-side spans would overflow, drop the subagent span first (it has lower priority than the parent counter). Parent counter is never dropped.

2. **`src/frontend/tui/tui_app.rs`**
   - Add `pub subagent_usage: TokenUsage` field to `App`.
   - Initialize to `TokenUsage::default()` in `App::new()`.
   - Change `AgentEvent::SubAgentUsage` handler to call `app.subagent_usage.add(...)` instead of `app.usage.add(...)`.
   - In `render_app()`, populate `StatusLineInfo.subagent_usage` with `Some(&app.subagent_usage)` when totals are non-zero, else `None`.
   - In the session-switch reset block (around line 601), reset `app.subagent_usage = TokenUsage::default()` alongside `app.usage`.

3. **Tests**
   - Update existing test `subagent_usage_event_aggregates_into_app_usage` → rename + rewrite to assert subagent events feed `app.subagent_usage` only and leave `app.usage` at zero. Add the inverse: parent `Usage` events do not touch `app.subagent_usage`.
   - Add unit test: session-switch resets both `app.usage` and `app.subagent_usage`.
   - Add unit test in `status_line.rs` for the conditional render — `build_status_line_with_subagent_usage_zero_omits_span` (no second counter when `Some(&zero)`) and `build_status_line_with_subagent_usage_renders_green_span`.
   - Add insta snapshot test `render_status_line_with_subagent_usage` (80×1) showing the second counter in green.
   - Update existing `render_status_line` and `render_status_line_narrow` snapshots — only if struct-init changes break them; populating `subagent_usage: None` should keep output identical, so ideally no diff.

### Acceptance Criteria

- [ ] `App.subagent_usage: TokenUsage` field exists and is initialized to default.
- [ ] `AgentEvent::SubAgentUsage` only mutates `subagent_usage`; `AgentEvent::Usage` only mutates `usage`. (Verified by two unit tests.)
- [ ] `StatusLineInfo` has `subagent_usage: Option<&TokenUsage>`.
- [ ] When subagent tokens are zero/absent, the rendered status line is byte-identical to the current implementation (existing snapshots unchanged).
- [ ] When subagent tokens are non-zero, a second token block renders to the right of the parent block in `Color::Green` with the `↗` prefix glyph.
- [ ] Session switching zeroes both counters.
- [ ] On a narrow terminal where space is tight, the subagent span is omitted before the parent span.
- [ ] New snapshot `render_status_line_with_subagent_usage` is committed.
- [ ] `cargo fmt`, `cargo clippy`, `cargo test` pass.

### Testing Approach

Test infrastructure already exists (insta, ratatui `TestBackend`). New tests live in:
- `src/frontend/tui/status_line.rs` `mod tests`: render & build-level tests + new insta snapshot.
- `src/frontend/tui/tui_app.rs` `mod tests`: event-handler unit tests + session reset test.

Behaviors to verify:
- Happy path: subagent event → only `subagent_usage` increments.
- Inverse: parent `Usage` event → only `usage` increments, `subagent_usage` stays zero.
- Render: `Some(&non_zero)` produces a green span with `↗` glyph; `Some(&zero)` and `None` both produce no extra span.
- Reset: session-switch zeros both counters.
- Width fallback: on narrow widths, subagent span is dropped first (assert via inspecting span list / snapshot at narrow width with non-zero subagent usage).

### Risks & Open Questions

1. **Glyph rendering** — `↗` may render inconsistently in some terminals. Issue suggests it; we'll use it as specified.
2. **Width algorithm** — current `build_status_line` computes a single gap from total left/right widths and panics nothing on overflow (it just truncates via terminal). We need an explicit "drop subagent if `left + right > width`" check. This is new logic with its own test.
3. **`is_estimated` semantics** — `load_history` seeds estimated parent tokens. Subagent tokens are never estimated (they only arrive from real events), so `subagent_usage.is_estimated` will always be false. No special handling.
4. **No backend / agent / persistence changes** — confirmed by grep; `AgentEvent::SubAgentUsage` is already piped through.

### DIFF
--- a/src/frontend/tui/snapshots/illustrious_manager__frontend__tui__status_line__tests__render_status_line_with_subagent_usage.snap
+++ b/src/frontend/tui/snapshots/illustrious_manager__frontend__tui__status_line__tests__render_status_line_with_subagent_usage.snap
@@ -1,0 +1,5 @@
+---
+source: src/frontend/tui/status_line.rs
+expression: terminal.backend()
+---
+" ↑12.5k ↓3.2k  ↗↑4.0k ↓800   main                                          claude-sonnet-4-20250514  /home/user/project "
--- a/src/frontend/tui/status_line.rs
+++ b/src/frontend/tui/status_line.rs
@@ -142,7 +142,6 @@ pub fn build_status_line(info: &StatusLineInfo<'_>, width: u16) -> Line<'static>
     let right_width: usize = right_spans.iter().map(|s| s.width()).sum();
     let total_width = width as usize;
 
-    // Determine whether the subagent span fits.
     let include_subagent = if let Some(ref sub_span) = subagent_span {
         let left_width_with = parent_span.width() + sub_span.width() + branch_span.width();
         left_width_with + right_width <= total_width
--- a/src/frontend/tui/status_line.rs
+++ b/src/frontend/tui/status_line.rs
@@ -27,6 +27,7 @@ pub struct StatusLineInfo<'a> {
     pub git_branch: Option<&'a str>,
     pub working_dir: &'a std::path::Path,
     pub usage: &'a TokenUsage,
+    pub subagent_usage: Option<&'a TokenUsage>,
 }
 
 /// Estimate token counts from a slice of conversation messages.
@@ -102,16 +103,30 @@ pub fn build_status_line(info: &StatusLineInfo<'_>, width: u16) -> Line<'static>
     let branch_str = info.git_branch.unwrap_or("no git");
     let dir_str = compact_path(info.working_dir);
 
-    let left_spans = vec![
-        Span::styled(
-            format!(" {usage_str} "),
-            Style::default().fg(Color::White).bg(STATUS_BG),
-        ),
-        Span::styled(
-            format!("  {branch_str} "),
-            Style::default().fg(Color::Cyan).bg(STATUS_BG),
-        ),
-    ];
+    let subagent_span: Option<Span<'static>> = info.subagent_usage.and_then(|su| {
+        if su.input_tokens + su.output_tokens > 0 {
+            let s = format!(
+                " ↗↑{} ↓{} ",
+                format_token_count(su.input_tokens),
+                format_token_count(su.output_tokens)
+            );
+            Some(Span::styled(
+                s,
+                Style::default().fg(Color::Green).bg(STATUS_BG),
+            ))
+        } else {
+            None
+        }
+    });
+
+    let parent_span = Span::styled(
+        format!(" {usage_str} "),
+        Style::default().fg(Color::White).bg(STATUS_BG),
+    );
+    let branch_span = Span::styled(
+        format!("  {branch_str} "),
+        Style::default().fg(Color::Cyan).bg(STATUS_BG),
+    );
 
     let right_spans = vec![
         Span::styled(
@@ -124,9 +139,24 @@ pub fn build_status_line(info: &StatusLineInfo<'_>, width: u16) -> Line<'static>
         ),
     ];
 
-    let left_width: usize = left_spans.iter().map(|s| s.width()).sum();
     let right_width: usize = right_spans.iter().map(|s| s.width()).sum();
     let total_width = width as usize;
+
+    // Determine whether the subagent span fits.
+    let include_subagent = if let Some(ref sub_span) = subagent_span {
+        let left_width_with = parent_span.width() + sub_span.width() + branch_span.width();
+        left_width_with + right_width <= total_width
+    } else {
+        false
+    };
+
+    let mut left_spans = vec![parent_span];
+    if include_subagent {
+        left_spans.push(subagent_span.expect("checked above"));
+    }
+    left_spans.push(branch_span);
+
+    let left_width: usize = left_spans.iter().map(|s| s.width()).sum();
     let gap = total_width.saturating_sub(left_width + right_width);
 
     let mut spans = left_spans;
@@ -183,6 +213,7 @@ mod tests {
             git_branch,
             working_dir,
             usage,
+            subagent_usage: None,
         }
     }
 
@@ -471,5 +502,120 @@ mod tests {
         let (input, output) = estimate_usage_from_messages(&messages);
         assert_eq!(input, 10, "40 chars / 4 = 10 input tokens");
         assert_eq!(output, 0);
+    }
+
+    #[test]
+    fn build_status_line_with_subagent_usage_zero_omits_span() {
+        let (model, branch, dir, usage) = test_info();
+        let subagent_zero = TokenUsage::default();
+        let info = StatusLineInfo {
+            model: &model,
+            git_branch: branch.as_deref(),
+            working_dir: &dir,
+            usage: &usage,
+            subagent_usage: Some(&subagent_zero),
+        };
+        let line = build_status_line(&info, 120);
+        let text: String = line.spans.iter().map(|s| s.content.to_string()).collect();
+        assert!(
+            !text.contains('↗'),
+            "zero subagent usage should not render ↗ glyph"
+        );
+    }
+
+    #[test]
+    fn build_status_line_with_subagent_usage_renders_green_span() {
+        let (model, branch, dir, usage) = test_info();
+        let subagent = TokenUsage {
+            input_tokens: 300,
+            output_tokens: 120,
+            ..Default::default()
+        };
+        let info = StatusLineInfo {
+            model: &model,
+            git_branch: branch.as_deref(),
+            working_dir: &dir,
+            usage: &usage,
+            subagent_usage: Some(&subagent),
+        };
+        let line = build_status_line(&info, 120);
+        let text: String = line.spans.iter().map(|s| s.content.to_string()).collect();
+        assert!(
+            text.contains('↗'),
+            "non-zero subagent usage should render ↗ glyph"
+        );
+        assert!(text.contains("↑300"), "should show subagent input tokens");
+        assert!(text.contains("↓120"), "should show subagent output tokens");
+
+        // The subagent span must be Color::Green
+        let green_span = line.spans.iter().find(|s| s.content.contains('↗'));
+        assert!(green_span.is_some(), "should find the ↗ span");
+        assert_eq!(
+            green_span.expect("checked").style.fg,
+            Some(Color::Green),
+            "subagent span should be green"
+        );
+    }
+
+    #[test]
+    fn build_status_line_subagent_span_dropped_on_narrow_terminal() {
+        let (model, branch, dir, usage) = test_info();
+        let subagent = TokenUsage {
+            input_tokens: 300,
+            output_tokens: 120,
+            ..Default::default()
+        };
+        let info = StatusLineInfo {
+            model: &model,
+            git_branch: branch.as_deref(),
+            working_dir: &dir,
+            usage: &usage,
+            subagent_usage: Some(&subagent),
+        };
+        // Use a very narrow width where the subagent span won't fit
+        let line = build_status_line(&info, 30);
+        let text: String = line.spans.iter().map(|s| s.content.to_string()).collect();
+        assert!(
+            !text.contains('↗'),
+            "subagent span should be dropped on narrow terminal (width=30)"
+        );
+        // Parent counter must still be present
+        assert!(text.contains('↑'), "parent counter must not be dropped");
+    }
+
+    #[test]
+    fn render_status_line_with_subagent_usage_snapshot() {
+        use std::path::PathBuf;
+        let model = "claude-sonnet-4-20250514".to_string();
+        let branch = Some("main".to_string());
+        let dir = PathBuf::from("/home/user/project");
+        let usage = TokenUsage {
+            input_tokens: 12500,
+            output_tokens: 3200,
+            ..Default::default()
+        };
+        let subagent = TokenUsage {
+            input_tokens: 4000,
+            output_tokens: 800,
+            ..Default::default()
+        };
+        let info = StatusLineInfo {
+            model: &model,
+            git_branch: branch.as_deref(),
+            working_dir: &dir,
+            usage: &usage,
+            subagent_usage: Some(&subagent),
+        };
+
+        let backend = ratatui::backend::TestBackend::new(120, 1);
+        let mut terminal = ratatui::Terminal::new(backend).expect("terminal creation");
+        terminal
+            .draw(|frame| {
+                let area = ratatui::layout::Rect::new(0, 0, 120, 1);
+                render_status_line(&info, frame, area);
+            })
+            .expect("draw");
+
+        insta::assert_snapshot!("render_status_line_with_subagent_usage", terminal.backend());
     }
 }
--- a/src/frontend/tui/tui_app.rs
+++ b/src/frontend/tui/tui_app.rs
@@ -52,6 +52,7 @@ pub struct App {
     pub text_width: u16,
     pub session_picker: Option<SessionPicker>,
     pub usage: TokenUsage,
+    pub subagent_usage: TokenUsage,
     /// The input_tokens value reported by the last Usage event. The API always
     /// reports the full context size, so we subtract the previous value to
     /// count only the newly added (non-cached) input tokens per turn.
@@ -90,6 +91,7 @@ impl App {
             text_width: 0,
             session_picker: None,
             usage: TokenUsage::default(),
+            subagent_usage: TokenUsage::default(),
             last_input_total: 0,
             model: String::new(),
             git_branch: None,
@@ -328,11 +330,17 @@ pub fn render_app(app: &mut App, frame: &mut ratatui::Frame) {
 
     app.input.render(frame, chunks[1]);
 
+    let subagent_ref = if app.subagent_usage.input_tokens + app.subagent_usage.output_tokens > 0 {
+        Some(&app.subagent_usage)
+    } else {
+        None
+    };
     let info = StatusLineInfo {
         model: &app.model,
         git_branch: app.git_branch.as_deref(),
         working_dir: &app.working_dir,
         usage: &app.usage,
+        subagent_usage: subagent_ref,
     };
     status_line::render_status_line(&info, frame, chunks[2]);
 
@@ -442,7 +450,7 @@ pub fn handle_agent_event(
             output_tokens,
             ..
         } => {
-            app.usage.add(input_tokens, output_tokens);
+            app.subagent_usage.add(input_tokens, output_tokens);
         }
     }
     Ok(())
@@ -599,6 +607,7 @@ async fn run_app(
                                                         app.current_response.clear();
                                                         app.scroll_offset = 0;
                                                         app.usage = TokenUsage::default();
+                                                        app.subagent_usage = TokenUsage::default();
                                                         app.last_input_total = 0;
                                                         app.load_history(&history);
                                                         agent.load_session(session).await;
@@ -1018,10 +1027,10 @@ mod tests {
     }
 
     #[test]
-    fn subagent_usage_event_aggregates_into_app_usage() {
+    fn subagent_usage_event_feeds_subagent_usage_only() {
         let mut app = App::new(std::sync::Arc::new(crate::tools::ToolRegistry::new()));
-        assert_eq!(app.usage.input_tokens, 0);
-        assert_eq!(app.usage.output_tokens, 0);
+        assert_eq!(app.subagent_usage.input_tokens, 0);
+        assert_eq!(app.subagent_usage.output_tokens, 0);
 
         let event = AgentEvent::SubAgentUsage {
             input_tokens: 200,
@@ -1030,9 +1039,13 @@ mod tests {
         };
         handle_agent_event(&mut app, event, None).expect("handle SubAgentUsage");
 
-        assert_eq!(app.usage.input_tokens, 200);
-        assert_eq!(app.usage.output_tokens, 80);
-        // SubAgentUsage does not update last_input_total (no deduplication logic needed)
+        // SubAgentUsage must NOT touch app.usage
+        assert_eq!(app.usage.input_tokens, 0, "parent usage must stay zero");
+        assert_eq!(app.usage.output_tokens, 0, "parent usage must stay zero");
+
+        // SubAgentUsage MUST accumulate into app.subagent_usage
+        assert_eq!(app.subagent_usage.input_tokens, 200);
+        assert_eq!(app.subagent_usage.output_tokens, 80);
         assert_eq!(app.last_input_total, 0);
 
         // A subsequent SubAgentUsage should add on top.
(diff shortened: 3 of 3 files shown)


Fizzizist:
## Implementation Plan: Separate Subagent Token Counter on Status Line

### Analysis

Today, both `AgentEvent::Usage` (parent) and `AgentEvent::SubAgentUsage` (subagent) accumulate into a single `App::usage` field, then render as one `↑in ↓out` block on the status line. The issue requests visual separation: a second token block (green) appears to the right of the parent counter only when subagent tokens have been recorded.

The change is **purely frontend display** — `AgentEvent::SubAgentUsage` already carries the subagent counts; the agent layer needs no changes.

### Deliverables

1. **`src/frontend/tui/status_line.rs`**
   - Add `subagent_usage: Option<&'a TokenUsage>` to `StatusLineInfo`.
   - In `build_status_line()`, when `subagent_usage` is `Some` and has non-zero totals, insert a second styled span (`Color::Green`) with format `↗↑{in} ↓{out}` immediately after the parent token span (still on the left side, before the git branch).
   - When `None` or zero totals, render exactly the current layout (no extra span, no extra spacing).
   - Width handling: if the right-side spans would overflow, drop the subagent span first (it has lower priority than the parent counter). Parent counter is never dropped.

2. **`src/frontend/tui/tui_app.rs`**
   - Add `pub subagent_usage: TokenUsage` field to `App`.
   - Initialize to `TokenUsage::default()` in `App::new()`.
   - Change `AgentEvent::SubAgentUsage` handler to call `app.subagent_usage.add(...)` instead of `app.usage.add(...)`.
   - In `render_app()`, populate `StatusLineInfo.subagent_usage` with `Some(&app.subagent_usage)` when totals are non-zero, else `None`.
   - In the session-switch reset block (around line 601), reset `app.subagent_usage = TokenUsage::default()` alongside `app.usage`.

3. **Tests**
   - Update existing test `subagent_usage_event_aggregates_into_app_usage` → rename + rewrite to assert subagent events feed `app.subagent_usage` only and leave `app.usage` at zero. Add the inverse: parent `Usage` events do not touch `app.subagent_usage`.
   - Add unit test: session-switch resets both `app.usage` and `app.subagent_usage`.
   - Add unit test in `status_line.rs` for the conditional render — `build_status_line_with_subagent_usage_zero_omits_span` (no second counter when `Some(&zero)`) and `build_status_line_with_subagent_usage_renders_green_span`.
   - Add insta snapshot test `render_status_line_with_subagent_usage` (80×1) showing the second counter in green.
   - Update existing `render_status_line` and `render_status_line_narrow` snapshots — only if struct-init changes break them; populating `subagent_usage: None` should keep output identical, so ideally no diff.

### Acceptance Criteria

- [ ] `App.subagent_usage: TokenUsage` field exists and is initialized to default.
- [ ] `AgentEvent::SubAgentUsage` only mutates `subagent_usage`; `AgentEvent::Usage` only mutates `usage`. (Verified by two unit tests.)
- [ ] `StatusLineInfo` has `subagent_usage: Option<&TokenUsage>`.
- [ ] When subagent tokens are zero/absent, the rendered status line is byte-identical to the current implementation (existing snapshots unchanged).
- [ ] When subagent tokens are non-zero, a second token block renders to the right of the parent block in `Color::Green` with the `↗` prefix glyph.
- [ ] Session switching zeroes both counters.
- [ ] On a narrow terminal where space is tight, the subagent span is omitted before the parent span.
- [ ] New snapshot `render_status_line_with_subagent_usage` is committed.
- [ ] `cargo fmt`, `cargo clippy`, `cargo test` pass.

### Testing Approach

Test infrastructure already exists (insta, ratatui `TestBackend`). New tests live in:
- `src/frontend/tui/status_line.rs` `mod tests`: render & build-level tests + new insta snapshot.
- `src/frontend/tui/tui_app.rs` `mod tests`: event-handler unit tests + session reset test.

Behaviors to verify:
- Happy path: subagent event → only `subagent_usage` increments.
- Inverse: parent `Usage` event → only `usage` increments, `subagent_usage` stays zero.
- Render: `Some(&non_zero)` produces a green span with `↗` glyph; `Some(&zero)` and `None` both produce no extra span.
- Reset: session-switch zeros both counters.
- Width fallback: on narrow widths, subagent span is dropped first (assert via inspecting span list / snapshot at narrow width with non-zero subagent usage).

### Risks & Open Questions

1. **Glyph rendering** — `↗` may render inconsistently in some terminals. Issue suggests it; we'll use it as specified.
2. **Width algorithm** — current `build_status_line` computes a single gap from total left/right widths and panics nothing on overflow (it just truncates via terminal). We need an explicit "drop subagent if `left + right > width`" check. This is new logic with its own test.
3. **`is_estimated` semantics** — `load_history` seeds estimated parent tokens. Subagent tokens are never estimated (they only arrive from real events), so `subagent_usage.is_estimated` will always be false. No special handling.
4. **No backend / agent / persistence changes** — confirmed by grep; `AgentEvent::SubAgentUsage` is already piped through.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implemented the separate subagent token counter per the plan.

**Changes:**

- `src/frontend/tui/status_line.rs` — Added `subagent_usage: Option<&'a TokenUsage>` to `StatusLineInfo`. Updated `build_status_line` to render an optional green `↗↑{in} ↓{out}` span after the parent counter when subagent tokens are non-zero; span is dropped first on narrow terminals. Added 4 new unit tests + 1 insta snapshot (`render_status_line_with_subagent_usage`).

- `src/frontend/tui/tui_app.rs` — Added `pub subagent_usage: TokenUsage` field to `App`, initialized to default. Fixed `AgentEvent::SubAgentUsage` handler to accumulate into `subagent_usage` only (not `usage`). Updated `render_app` to pass `subagent_usage` into `StatusLineInfo`. Added reset of `subagent_usage` in the session-switch block. Rewrote `subagent_usage_event_aggregates_into_app_usage` to assert correct separation; added `parent_usage_event_does_not_touch_subagent_usage` and `session_switch_resets_both_usage_counters`.

**563 tests pass.**

**Commit:** `1618bd6`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

### Critical

_None._

### Important

**Finding 1 — `session_switch_resets_both_usage_counters` does not exercise production code**
`src/frontend/tui/tui_app.rs` (test body)
The test manually re-enacts the reset block instead of calling the real session-switch path in `run_app` (line ~610). It sets `app.usage = TokenUsage::default()` and `app.subagent_usage = TokenUsage::default()` itself, then asserts they are zero. This is circular — it tests `TokenUsage::default()`, not the production reset logic. If a future change drops the `subagent_usage` reset from `run_app`, this test will still pass. Either factor the reset into a method on `App` (e.g. `app.reset_for_session_switch()`) and test that, or add an integration-level test that drives the actual `run_app` switch flow.

**Finding 2 — Duplicated zero-token guard between `render_app` and `build_status_line`**
Files: `src/frontend/tui/tui_app.rs` (`render_app`, lines ~333–337) and `src/frontend/tui/status_line.rs` (`build_status_line`, lines ~106–115)
Two independent implementations of the same "non-zero token" rule. `render_app` filters before passing the reference; `build_status_line` filters again inside `and_then`. If the threshold ever changes, both must be updated. Single source of truth: drop the outer guard in `render_app` and pass `Some(&app.subagent_usage)` unconditionally — possibly even change the field to `&'a TokenUsage` since `Option` no longer carries information.

**Finding 3 — No integration test for `render_app` → `build_status_line` with non-zero `subagent_usage`**
`src/frontend/tui/tui_app.rs` (`render_app_includes_status_line`)
The existing `render_app` integration test never sets `app.subagent_usage`, so the new conditional in `render_app` (the `subagent_ref` branch where `Some` is chosen) is untested through the production rendering path. The new snapshot test in `status_line.rs` calls `render_status_line` directly, bypassing `render_app`. A test that does `app.subagent_usage.add(500, 100)` then `render_app(&mut app, ...)` and asserts a green span with the arrow glyph appears would close the gap.

### Suggestions

**Finding 4 — `subagent_usage` is never reset within a session (semantic ambiguity)**
`src/frontend/tui/tui_app.rs` (`submit_message` and friends)
The PR commits to "session-lifetime accumulator" semantics: `subagent_usage` only resets on session switch, never per submission. Neither the issue nor the PR documents this choice. If the intent is "tokens used by subagents in the current turn", the counter never clears within a session, which would be a silent display bug. If session-lifetime is intentional (matches `app.usage`), it is fine — just worth a sentence in the PR description and a test asserting non-reset on submit.

**Finding 5 — `and_then` + `if/else` can collapse to `filter` + `map`**
`src/frontend/tui/status_line.rs` (lines 106–119)
```rust
info.subagent_usage
    .filter(|su| su.input_tokens + su.output_tokens > 0)
    .map(|su| Span::styled(
        format!(" ↗↑{} ↓{} ", format_token_count(su.input_tokens), format_token_count(su.output_tokens)),
        Style::default().fg(Color::Green).bg(STATUS_BG),
    ))
```
Less nesting, no explicit `Some`/`None` arms.

**Finding 6 — `if let Some(ref ...) { ... } else { false }` → `is_some_and`**
`src/frontend/tui/status_line.rs` (lines 140–146)
```rust
let include_subagent = subagent_span.as_ref().is_some_and(|sub_span| {
    parent_span.width() + sub_span.width() + branch_span.width() + right_width <= total_width
});
```
Drops the intermediate `left_width_with` binding and the `if let`/`else` branching.

**Finding 7 — Mutable `left_spans` push-dance → iterator chain**
`src/frontend/tui/status_line.rs` (lines 148–153)
```rust
let left_spans: Vec<Span<'static>> = std::iter::once(parent_span)
    .chain(subagent_span.filter(|_| include_subagent))
    .chain(std::iter::once(branch_span))
    .collect();
```
Eliminates the `mut`, the `expect("checked above")`, and the cross-statement invariant. `Option<T>: IntoIterator` does the gating naturally.

**Finding 8 — Narrow-terminal test uses arbitrary `width=30`, not the boundary**
`src/frontend/tui/status_line.rs` (`build_status_line_subagent_span_dropped_on_narrow_terminal`)
Width is far below any plausible threshold; the test verifies "very narrow drops the span", not "the drop decision happens at the correct boundary". An off-by-one between `<=` and `<` in the fit check would not be caught. Compute the exact crossover width, assert the arrow glyph is present at +1, absent at boundary.

**Finding 9 — `None` vs `Some(&zero)` distinction is implicitly tested only**
`src/frontend/tui/status_line.rs`
Tests cover `Some(&zero)` (omits span) and `Some(&nonzero)` (renders span). `None` is only covered indirectly via the `make_info` default. No test explicitly asserts `subagent_usage: None` and `Some(&TokenUsage::default())` produce identical output. Low priority; would also become moot if Finding 2 is taken (drop the `Option`).

**Finding 10 — Unnecessary code comment violates project policy**
`src/frontend/tui/status_line.rs` (line 145)
```rust
// Determine whether the subagent span fits.
```
CLAUDE.md: comments only when something cannot be gleaned from the code. The variable name and the comparison are self-documenting. Becomes moot if Finding 6 is taken.

### Strengths

- All ten acceptance criteria from issue #150 are met.
- `cargo fmt --check`, `cargo clippy`, and `cargo test` all pass cleanly.
- Existing `render_status_line` and `render_status_line_narrow` snapshots are unchanged — confirming the byte-identical-when-no-subagent claim from the plan.
- Two unit tests verify the routing invariant in both directions: `subagent_usage_event_feeds_subagent_usage_only` (positive) and `parent_usage_event_does_not_touch_subagent_usage` (inverse).
- The `.expect("checked above")` is logically safe — `include_subagent` is `true` only inside the `Some` arm — and the message is informative. Policy compliant.
- New snapshot is one-line and reviewable.
- No unwrap, no unwrap-by-stealth.
- Token sums and width arithmetic carry no realistic overflow risk (`u64` totals, `usize` widths, `saturating_sub` on the gap).

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

Review comment: https://github.com/Fizzizist/illustrious-manager/pull/162#issuecomment-4328000372

### Classifications

|---|---|---|
| Finding 1 — `session_switch_resets_both_usage_counters` circular test | genuine | Confirmed. The test manually executes `app.usage = TokenUsage::default(); app.subagent_usage = TokenUsage::default()` then asserts they are zero — it mirrors the production block verbatim instead of calling through it. If the `subagent_usage` reset were removed from `run_app`, this test still passes. |
| Finding 2 — Duplicated zero-token guard | genuine | Confirmed. `render_app` (line 333) checks `input_tokens + output_tokens > 0` before passing `Some(&app.subagent_usage)`, and `build_status_line` (line 107–114) performs the identical check inside `and_then`. The outer guard in `render_app` is fully redundant — the test `build_status_line_with_subagent_usage_zero_omits_span` already passes `Some(&zero)` directly and confirms the inner guard is operative. |
| Finding 3 — No `render_app` integration test with non-zero `subagent_usage` | genuine | Confirmed. `render_app_includes_status_line` never sets `app.subagent_usage` to a non-zero value, so the `Some(&app.subagent_usage)` branch at line 333–334 is never exercised through the production rendering path. The snapshot tests call `render_status_line` / `build_status_line` directly, bypassing `render_app`. |
| Finding 4 — `subagent_usage` semantic ambiguity / missing test | nitpick | The accumulation-per-session semantics are the only sensible design and the existing tests already demonstrate this behavior (`subagent_usage_event_feeds_subagent_usage_only`, `session_switch_resets_both_usage_counters`). A "non-reset on submit" assertion would add marginal value. |
| Finding 5 — `and_then` + `if/else` → `filter` + `map` | nitpick | Stylistic. Both forms are idiomatic Rust; `and_then` with an `if` is perfectly readable. |
| Finding 6 — `if let Some(ref ...) { ... } else { false }` → `is_some_and` | nitpick | The `if let` body computes `left_width_with` before returning the boolean — it's not a pure predicate, so `is_some_and` would require folding that computation into the closure. No clarity gain. |
| Finding 7 — Mutable `left_spans` push-dance → iterator chain | nitpick | Stylistic. The current form is clear; an iterator chain would require `chain` + `flatten` + `once` and be harder to follow. The `.expect("checked above")` is safe and explained. |
| Finding 8 — Narrow-terminal test uses `width=30`, not boundary | nitpick | The test verifies the drop-on-narrow behavior at an extreme value. Computing the exact crossover width would make the test fragile to font/span changes. |
| Finding 9 — `None` vs `Some(&zero)` distinction only implicitly tested | false-positive | Moot given finding 2 is genuine. If the outer guard in `render_app` is dropped, `build_status_line` always receives `Some` and the existing test `build_status_line_with_subagent_usage_zero_omits_span` covers the zero case directly. |
| Finding 10 — Unnecessary comment at `status_line.rs` line 145 | genuine | Confirmed. Line 145 is `// Determine whether the subagent span fits.` — a `//` comment immediately before `let include_subagent = ...`. The variable name makes the intent self-evident. Violates the "minimize comments" policy in CLAUDE.md. (Note: the `.expect("checked above")` string is an `expect` message, not a comment — that part is policy-compliant.) |

### Action Plan

1. **Fix finding 3** (test gap in production path) — Add a `render_app` test that sets `app.subagent_usage` to a non-zero value and asserts the `↗` glyph appears in the rendered output. This exercises the `Some` branch at line 333.

2. **Fix finding 1** (circular test) — Either extract the reset into an `App::reset_for_session_switch()` method and call it from both `run_app` and the test, or replace the body with a call to `handle_agent_event` / a helper that exercises the real code path.

3. **Fix finding 2** (duplicated guard) — Remove the `input_tokens + output_tokens > 0` guard from `render_app` and pass `Some(&app.subagent_usage)` unconditionally. The inner guard in `build_status_line` is sufficient and already tested.

4. **Fix finding 10** (redundant comment) — Delete `// Determine whether the subagent span fits.` from `status_line.rs` line 145.

---
*Assessment by mach6*

Fizzizist:
## Review Assessment

Review: https://github.com/Fizzizist/illustrious-manager/pull/162#issuecomment-4328000372

### Classifications

|---|---|---|
| Finding 1 — Circular session-switch test | **genuine** | Confirmed. The test manually executes `app.usage = TokenUsage::default(); app.subagent_usage = TokenUsage::default()` then asserts zero — it mirrors the production block in `run_app` verbatim instead of exercising it. Drop `subagent_usage` from `run_app` and the test still passes. |
| Finding 2 — Duplicated zero-token guard | **genuine** | Confirmed. `render_app` (line ~333) and `build_status_line` (lines ~107–114) both check `input_tokens + output_tokens > 0`. The inner guard is operative — proven by `build_status_line_with_subagent_usage_zero_omits_span` passing `Some(&zero)` directly and the span being correctly suppressed. The outer guard is redundant. |
| Finding 3 — No `render_app` integration test with non-zero `subagent_usage` | **genuine** | Confirmed. `render_app_includes_status_line` never sets `app.subagent_usage` non-zero, so the `Some(&app.subagent_usage)` branch is untested through the production rendering path. Snapshot tests bypass `render_app` and call `render_status_line`/`build_status_line` directly. |
| Finding 4 — Semantic ambiguity / missing non-reset test | **nitpick** | Existing tests already demonstrate session-lifetime accumulation behavior. A "doesn't reset on submit" assertion adds marginal value. |
| Finding 5 — `and_then`/`if-else` → `filter`/`map` | **nitpick** | Both forms are idiomatic. No correctness impact. |
| Finding 6 — `if let … else { false }` → `is_some_and` | **nitpick** | The body also binds `left_width_with` before returning — not a pure predicate. `is_some_and` would require folding the computation into the closure, with no clarity gain. |
| Finding 7 — Mutable push-dance → iterator chain | **nitpick** | Stylistic. Current form is straightforward; a `once`/`chain` chain would be denser and arguably less readable. |
| Finding 8 — Narrow test uses `width=30` not exact boundary | **nitpick** | An exact-boundary test would be fragile to future span/content changes. The intent is to verify "subagent dropped, parent retained" at narrow widths — current test does that. |
| Finding 9 — `None` vs `Some(&zero)` distinction only implicitly tested | **false-positive** | Becomes moot once Finding 2 is fixed (outer guard drops, inner guard carries all weight, already directly tested). |
| Finding 10 — Redundant comment at `status_line.rs:145` | **genuine** | Confirmed. `// Determine whether the subagent span fits.` precedes `let include_subagent = ...` — the variable name plus the comparison expression are self-documenting. Violates the "minimize comments" policy. |

### Action Plan

1. **Finding 3** — Add a `render_app` test that sets `app.subagent_usage` non-zero and asserts the green arrow span appears in the rendered output. Closes the gap in production-path coverage of the new conditional.
2. **Finding 1** — Extract an `App::reset_for_session_switch()` method, call it from `run_app`'s session-switch branch, and rewrite the test to invoke that method. Removes the circularity.
3. **Finding 2** — Drop the zero-token guard from `render_app`; pass `Some(&app.subagent_usage)` unconditionally. Single source of truth for the suppression rule.
4. **Finding 10** — Delete the redundant comment at `status_line.rs:145`.

**Counts:** 4 genuine · 5 nitpicks · 1 false-positive · 0 deferred

---
*Assessment by mach6*

Fizzizist:
## Progress Update

Addressed all 4 genuine review findings.

**Finding 1** — Extracted `App::reset_for_session_switch()` method that zeroes `usage`, `subagent_usage`, and `last_input_total`. The session-switch block in `run_app` now calls this method. The test `session_switch_resets_both_usage_counters` was rewritten to call `reset_for_session_switch()` directly, exercising production code instead of mimicking it.

**Finding 2** — Removed the redundant `input_tokens + output_tokens > 0` guard from `render_app`. `subagent_usage` is now passed as `Some(&app.subagent_usage)` unconditionally; the identical inner guard in `build_status_line` is the single source of truth.

**Finding 3** — Added `render_app_with_subagent_usage_shows_arrow_glyph` test: constructs an `App` with non-zero `subagent_usage`, renders via `TestBackend`, and asserts the `↗` glyph and `↑300` token count appear in the output. Closes the gap in production-path coverage.

**Finding 10** — Deleted the redundant `// Determine whether the subagent span fits.` comment from `status_line.rs`.

**564 tests pass.**

**Commit:** `b889402`

---
*Progress tracked by mach6*

PR #163, opened 2026-04-27
Order sessions by last activity (issue 159)

### LINKED ISSUE #159: sessions ordered by updated at
Right now, the `/sessions` command lists sessions by when they were created. We should switch this to when they were last updated.

Fizzizist:
Closes #159

Order `/sessions` by most recent activity (`MAX(conversation.created_at)`) instead of the main `.db` file mtime. Fixes the WAL-lag bug where active sessions failed to bubble to the top until a checkpoint.


Fizzizist:
## Implementation Plan

### Problem
`/sessions` orders by main `.db` file mtime, which lags behind real activity because turso/libSQL writes go to the WAL until checkpoint. Active sessions don't bubble to the top of the picker.

### Approach (chosen by user: B)
Per-message `created_at`; derive session activity from `MAX(created_at)`. Truth-aligned: the timestamp lives inside the data, independent of filesystem semantics.

### Deliverables
1. **Schema:** `conversation` table gains `created_at INTEGER NOT NULL` (Unix epoch **seconds**, matching the `task` table convention).
2. **Idempotent migration:** `Session::new` checks `PRAGMA table_info(conversation)`; if `created_at` is missing, runs `ALTER TABLE conversation ADD COLUMN created_at INTEGER NOT NULL DEFAULT 0`. Done once per DB open; cheap.
3. **Insert path:** `ConversationRepo::insert_message` populates `created_at` with `now_epoch_secs()`.
4. **Listing:** `list_sessions` queries `SELECT MAX(created_at) FROM conversation` per session DB. Falls back to filesystem mtime when `MAX` is `0`/`NULL` (legacy DBs with no new messages since migration). Sorts descending.
5. **`SessionSummary.modified` semantics:** field becomes "last activity time" derived from `MAX(created_at)` (with mtime fallback). Type stays `SystemTime` so picker code is untouched.
6. **Tests:** new behavior tests; existing fragile sleep-based mtime test rewritten.

### Non-goals (deliberate)
- Picker UI does not display the timestamp (silent sort only).
- No new "sessions index" DB.
- No `session.updated_at` stored column. `MAX()` aggregate is the source of truth.
- `task` table is unchanged.

### Files to modify
- `src/session/mod.rs`
  - Extend `SCHEMA` to include `created_at INTEGER NOT NULL DEFAULT 0` on `conversation`.
  - Add a `migrate_conversation_created_at(&Connection)` helper: `PRAGMA table_info(conversation)` -> if `created_at` is absent, run `ALTER TABLE conversation ADD COLUMN created_at INTEGER NOT NULL DEFAULT 0`. Call from `Session::new` after `execute_batch(SCHEMA)`.
  - Replace the existing `read_first_user_message_from_path` call site with a combined helper that reads both first user message **and** `MAX(created_at)` in one DB open (avoid doubling per-session DB opens).
  - Compute `modified` as `SystemTime::UNIX_EPOCH + Duration::from_secs(max_created_at)` when `max_created_at > 0`, else fall back to file mtime.
- `src/session/conversation.rs`
  - Reuse `now_epoch_secs()` (already in `task.rs` -- promote to a shared helper or duplicate; lean toward duplicate to keep modules self-contained, decide during implementation).
  - `insert_message` SQL becomes `INSERT INTO conversation (role, content, created_at) VALUES (?1, ?2, ?3)`.
  - New method `max_created_at(&self) -> Result<Option<i64>>` returning the aggregate (None if empty/all-zero).

### Migration semantics for existing session DBs
- Pre-existing rows get `created_at = 0`.
- For a session DB whose rows are all pre-migration, `MAX(created_at) = 0` -> fall back to file mtime so old sessions still sort somewhere reasonable instead of all collapsing to the epoch.
- As soon as one new message is inserted, the session sorts by real activity.

### Testing approach
All in `src/session/mod.rs` and `src/session/conversation.rs` `#[cfg(test)]` blocks (existing infra: `tempfile::TempDir` + `tokio::test`).

**New tests:**
1. `conversation_insert_populates_created_at` -- insert a message, query raw `SELECT created_at FROM conversation`, assert `> 0` and within ~5s of `SystemTime::now()`.
2. `conversation_max_created_at_returns_none_for_empty_table`.
3. `conversation_max_created_at_returns_latest_insert_time`.
4. `list_sessions_orders_by_max_created_at_not_file_mtime` -- create session A, insert into A, then create session B and insert into B later; assert B sorts first. If feasible, also bump A's `.db` mtime to be newer than B's via `filetime` (or `std::fs::File::set_modified` if available on the toolchain) to prove sort is mtime-independent. **This is the regression test for the WAL-mtime bug.**
5. `list_sessions_falls_back_to_mtime_for_legacy_session_with_no_created_at` -- manually create a `<uuid>.db` with the old schema (no `created_at` column), insert a row with the old SQL, then call `list_sessions` and assert the session appears (migration ran on open, `MAX = 0`, mtime fallback engaged).
6. `migration_is_idempotent` -- open the same DB twice via `Session::new`, assert no error and column still present.
7. `migration_adds_column_to_existing_db_without_it` -- create DB with old schema directly, reopen via `Session::new`, assert `INSERT` with new SQL succeeds.

**Existing test to rewrite:**
- `list_sessions_orders_by_most_recently_modified_first` -- drop the `tokio::time::sleep(50ms)` and rely on monotonic ordering of `created_at` between two sequential inserts. Rename to `list_sessions_orders_by_most_recent_activity`.

### Risks and open questions
1. **Tied timestamps within the same second.** Multi-tool-call turns insert several rows back-to-back. Acceptable: `list_sessions` only sorts at the session level (one `MAX()` per DB), so per-message ties are irrelevant.
2. **Migration on legacy DBs.** turso/libSQL `ALTER TABLE ADD COLUMN ... NOT NULL DEFAULT 0` should be supported; verified by test item 5 before merge.
3. **Concurrent `Session::new` on same DB.** Not a current pattern; check-then-`ALTER` is not atomic but sessions are single-process. Acceptable.
4. **Listing perf.** Adds one `SELECT MAX(created_at)` per session DB on top of the existing `SELECT content ... LIMIT 1`. Combined into a single connection open per session (no regression vs. status quo).

### Acceptance criteria
- After change: create session A, then create session B and insert a message into B -- `list_sessions` returns `[B, A]` even when A's `.db` mtime is artificially newer than B's.
- Pre-existing session DBs (no `created_at` column) load, list, and accept new messages without error.
- All new tests pass; rewritten ordering test passes without `sleep`.
- `cargo fmt` clean, `cargo clippy` clean, no `.unwrap()` introduced.

---
*Plan created by mach6*

PR #164, opened 2026-04-27
Render markdown tables in TUI conversation

### LINKED ISSUE #133: markdown table display
`tui-markdown` doesn't support tables, but AI likes to display a lot of tables in it's output. We should either fork tui-markdown, move our stuff into the fork (including the diff logic (to unify everything)). Or find a way to tack an table rendering to work with the tui-markdown problem

Fizzizist:
Closes #133

Render GFM markdown tables in the TUI conversation by pre-processing them into monospaced fenced code blocks before handing the string to `tui_markdown::from_str`. No fork, no diff-logic refactor.


Fizzizist:
## Implementation Plan

### Problem
`tui-markdown 0.3.7` silently drops GFM tables. Assistant responses containing tables render as empty space or stray pipe/dash characters. The single render path is `render_role_lines` in `src/frontend/tui/conversation_area.rs:148`, which calls `tui_markdown::from_str` directly.

### Approach
**Pre-process tables into monospaced fenced code blocks before handing the string to `tui_markdown::from_str`.** No fork, no upstream patch, no diff-logic refactor. Reversible by deleting one module and one call site when upstream PR 125 (or a successor) ships table support.

The diff-logic unification suggested in the issue is explicitly rejected — both paths already converge on `Vec<Line<'static>>` and are not coupled. Re-architecting them solves nothing.

### Deliverables
1. New module `src/frontend/tui/markdown_tables.rs` exposing one function:
   ```rust
   pub fn preprocess_tables(input: &str) -> String
   ```
   - Parses input with `pulldown-cmark` using `Options::ENABLE_TABLES`.
   - For each `Tag::Table`, collects header + body cells, computes per-column display width via `unicode_width::UnicodeWidthStr`, formats a monospaced grid with Unicode box-drawing characters (`─ │ ┼` etc.), wraps the result in a triple-backtick fence so `tui-markdown` renders it as a verbatim code block, and substitutes it for the original byte range.
   - Honors source alignment (`Alignment::Left/Center/Right/None`).
   - Non-table input returns unchanged (cheap path: if no `|` in input, skip parsing).
2. Wire `preprocess_tables` into `render_role_lines` (one-line change at `conversation_area.rs:157`).
3. Add `pulldown-cmark` as a direct dependency via `cargo add pulldown-cmark` (already in the tree transitively at 0.13.3 via tui-markdown — no new download).
4. Insta snapshot tests for visual regression.
5. Unit tests for `preprocess_tables` covering parsing/formatting behavior.

### Files to Create or Modify
| File | Change |
|---|---|
| `Cargo.toml` | `cargo add pulldown-cmark` (no manual edit) |
| `src/frontend/tui/markdown_tables.rs` | **new** — table extraction + monospace grid formatter |
| `src/frontend/tui/mod.rs` | declare `mod markdown_tables;` |
| `src/frontend/tui/conversation_area.rs` | call `preprocess_tables(&display_content)` before `markdown_to_text` |
| `src/frontend/tui/snapshots/*table*.snap` | new insta snapshots (generated) |

### Acceptance Criteria
- A simple GFM table in an assistant message renders as a visible monospaced grid with aligned columns and visible borders (snapshot test passes).
- Table cells containing CJK / emoji / combining characters align correctly (column widths use `unicode-width`, not `.len()`).
- Column alignment markers (`:---`, `---:`, `:---:`) influence rendered cell padding.
- Input without tables is byte-identical after preprocessing for the no-pipe fast path; otherwise structurally equivalent (existing snapshots — bold, code, heading, list, italic — remain unchanged).
- A table wider than the conversation viewport is allowed to overflow into the code block's wrap behavior (no special truncation logic — out of scope, documented as known limitation).
- `cargo fmt`, `cargo clippy`, `cargo test` all green. No `.unwrap()`.

### Testing Approach
**Unit tests in `markdown_tables.rs`:**
- `preprocess_passes_through_non_table_input` — no `|`, returns same string
- `preprocess_passes_through_pipes_outside_tables` — `|` inside inline code or prose doesn't trigger table formatting
- `preprocess_formats_basic_table` — 2-col, 2-row table → fenced grid, exact string match
- `preprocess_respects_column_alignment` — `:---:` produces centered cell content
- `preprocess_handles_unicode_widths` — emoji + CJK column widths align (verify column boundary positions)
- `preprocess_handles_empty_cells` — empty cell renders as padded blank
- `preprocess_handles_multiple_tables` — two tables in one input both convert
- `preprocess_preserves_surrounding_markdown` — text/headings before and after the table are untouched

**Insta snapshot tests in `conversation_area.rs` (mirrors `render_markdown_list`):**
- `render_markdown_table` — basic 2-col table at width 60
- `render_markdown_table_with_alignment` — mixed left/center/right
- `render_markdown_table_in_assistant_with_surrounding_text` — heading + table + paragraph

### Risks / Open Questions
- **Box-drawing vs ASCII grid:** Plan uses Unicode box-drawing for readability. Trivially swappable to ASCII via character constants if terminal compatibility becomes a problem.
- **Wide-table overflow:** When a table is wider than the viewport, the fenced code block will wrap mid-row, producing ugly output. Acceptable for v1 — same behavior as long code blocks today. Out of scope to add horizontal scroll.
- **Streaming partial tables:** During streaming, `current_response` may contain a half-written table (no closing row). `pulldown-cmark` will simply not emit a `Tag::Table` until the structure is complete, so partial tables render as raw markdown until complete. Acceptable.
- **Upstream PR 125** (joshka/tui-markdown) is open and mergeable but the maintainer requested a split into smaller PRs that hasn't happened in ~2 months. This workaround is the right call; revisit when a tables-only release ships.
- **`pulldown-cmark` version drift:** Pinning to `0.13` matches what `tui-markdown 0.3.7` already pulls in, avoiding two parser versions in the binary. `cargo add pulldown-cmark` should resolve to `0.13.x`; verify post-add.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implemented the full plan from the plan comment. Changes in commit `91f38ee`:

- **`src/frontend/tui/markdown_tables.rs`** (new) — `preprocess_tables(input: &str) -> String`. Fast-path exit when no `|` present. Uses `pulldown-cmark` with `ENABLE_TABLES` to parse GFM tables, formats each into a Unicode box-drawing monospaced grid (box chars: `┌ ─ ┬ ┐ │ ╞ ═ ╪ ╡ ├ ┼ ┤ └ ┴ ┘`), wraps in triple-backtick fence, and substitutes in-place. Column widths via `unicode-width` (CJK/emoji safe). Left/center/right alignment honored. Multi-table inputs handled. 8 unit tests covering all plan cases.
- **`src/frontend/tui/mod.rs`** — `pub mod markdown_tables;` declared.
- **`src/frontend/tui/conversation_area.rs`** — `preprocess_tables` called before `markdown_to_text` in `render_role_lines`. 3 insta snapshot tests added: basic table, mixed alignment table, table with surrounding heading and paragraph.
- **`Cargo.toml` / `Cargo.lock`** — `pulldown-cmark` added as direct dependency via `cargo add` (resolves to `0.13.3`, matching existing transitive version).

All 568 tests pass. `cargo fmt` and `cargo clippy` clean.

**Commit:** `91f38ee`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

### Critical

*None.*

### Important

**Finding 1 — Alignment unit test makes no meaningful assertions** *(test-reviewer, high)*
`src/frontend/tui/markdown_tables.rs` — `preprocess_respects_column_alignment`

The test asserts `result.contains("Right")` and `result.contains("b")` — both would pass even if all padding logic were deleted. A test titled "respects column alignment" must assert the actual padded output, e.g.:
```rust
assert!(result.contains("│     c │"), "right-aligned cell should have leading spaces");
assert!(result.contains("│   b    │"), "center-aligned cell should have symmetric padding");
```
The insta snapshot `render_markdown_table_with_alignment` does lock in correct alignment at the integration level, but this unit test is effectively a no-op for the logic it claims to verify.

**Finding 2 — Mismatched column counts between header and body rows is untested** *(test-reviewer, high)*
`src/frontend/tui/markdown_tables.rs` — `render_table` / `format_row`

`render_table` derives `num_cols` from `header.len()`. A body row with **more** columns than the header silently truncates the extras. A body row with **fewer** columns silently pads. Neither behavior is tested. The truncation case is silent data loss that users would notice when the model produces an irregular table.

**Finding 3 — Rich inline markup (bold/italic/links) silently stripped — no test documents the behavior** *(test-reviewer + code-reviewer, high)*
`src/frontend/tui/markdown_tables.rs` — event handler `_ => {}` arm

`**bold**`, `*italic*`, `[text](url)` in cells lose their markup; only `Event::Text` is captured. This is a defensible design choice for a monospaced grid, but it is completely undocumented by tests. A test pinning the input `| **bold** | [text](url) |` → output containing `bold` and `text` (no asterisks, no brackets) would lock the intended behavior in.

**Finding 4 — Partial/streaming table input has no test coverage** *(test-reviewer, high)*
`src/frontend/tui/markdown_tables.rs` — `preprocess_tables` is called from `render_role_lines` for the streaming `current_response` too.

A header+separator with no body row yet (`| Name | Age |\n|------|-----|`) is parsed by pulldown-cmark and emits a header-only table. A truly partial input like `| Name | Age |\n|---` (incomplete separator) is **not** parsed as a table, so bare pipes flow through to `tui_markdown::from_str`. No test documents what either case actually renders. Without coverage, any attempt to "fix" streaming flicker may break silently.

### Suggestions

**Finding 5 — Redundant variable-declaration comments violate the commenting policy** *(code-reviewer, medium)*
`src/frontend/tui/markdown_tables.rs` — lines 17, 22, 28, 30, 32, 34

```rust
// State machine
let mut in_table = false;
...
// header row: one String per column
let mut header: Vec<String> = Vec::new();
// body rows
let mut body: Vec<Vec<String>> = Vec::new();
// current row being accumulated
let mut current_row: Vec<String> = Vec::new();
// current cell text
let mut current_cell = String::new();
```
The comments restate what the names and types already say. CLAUDE.md is explicit: *"Only comment in cases where something is unable to be gleaned from the code itself."* `// State machine` over a `for event in parser { match event { ... } }` is also pure noise — that pattern *is* the universally recognized state-machine form.

**Finding 6 — No test for inline code in table cells** *(test-reviewer, medium)*
`src/frontend/tui/markdown_tables.rs` — `Event::Code` branch

The handler for inline code backtick-wraps the extracted text back into the cell. This is the only `Event` other than `Text`/`SoftBreak`/`HardBreak` that is explicitly handled, and it has no test. A cell like `` | `ls -la` | ok | `` should round-trip with backticks preserved and column width sized to the full backtick-wrapped string.

**Finding 7 — No test for a header-only table (zero body rows)** *(test-reviewer + code-reviewer, medium)*
`src/frontend/tui/markdown_tables.rs` — `render_table`

`| A | B |\n|---|---|` is valid GFM with no body. The function should produce a top-border → header → double-separator → bottom-border grid (no mid-divider, no panic). Zero coverage for this shape today.

**Finding 8 — No test for single-column tables** *(test-reviewer, medium)*
`render_table` emits joiners (`┬`, `╪`, `┼`, `┴`) only `if i + 1 < num_cols`. For a single-column table no joiners are emitted and the borders become `┌──┐` / `└──┘`. This is a structurally distinct path from the multi-column case but it is entirely untested.

**Finding 9 — No test for tables in `User` (or other non-`Assistant`) role messages** *(test-reviewer, medium)*
`src/frontend/tui/conversation_area.rs` — `render_role_lines` is invoked for **all** roles. All three table snapshots use `Assistant`. A `User`-role table going through the same code path with a different label color and the `"  "` indent has no snapshot; an interaction between the indent and a wide table could push it off-screen unnoticed.

**Finding 10 — No snapshot test for tables in the streaming `current_response` path** *(test-reviewer, low)*
`src/frontend/tui/conversation_area.rs` — `render_current_response_lines`. If `preprocess_tables` were accidentally removed from that path (or the path were refactored), no test would catch the regression. A snapshot mirroring the basic-table test but invoking the streaming render would close that gap.

**Finding 11 — CJK/emoji alignment not covered by a snapshot test** *(completeness-checker, medium)*
Acceptance criterion 2 ("CJK / emoji / combining characters align correctly") is verified only by a weak unit test (`preprocess_handles_unicode_widths` only counts `│` characters per line, not actual padding). No insta snapshot pins the exact terminal render of a table with wide Unicode characters. A regression in padding arithmetic for those characters would not be caught.

**Finding 12 — ANSI escape sequences in table cells produce wrong column widths** *(error-auditor, low)*
`UnicodeWidthStr::width()` counts each ANSI escape byte as width 1. A cell containing `\x1b[1mbold\x1b[0m` (4 visible characters) returns `.width()` = 12, causing 8 excess spaces of padding per cell in that column. Silent display corruption, no panic. Unlikely (LLMs don't emit ANSI in table cells) but real.

**Finding 13 — Wide-table overflow limitation not documented in source** *(completeness-checker, low)*
The plan explicitly marks wide-table horizontal scroll as out of scope. That note lives only in the PR comment, not in the source. A sentence in the `preprocess_tables` doc comment would make it read as intentional rather than an oversight.

**Finding 14 — Four nearly-identical border-drawing loops** *(simplifier, medium)*
`src/frontend/tui/markdown_tables.rs` — `render_table`

The top border, `╞═╪═╡` separator, mid-row divider, and bottom border are four copies of the same `push(left)` / `for … push(fill)` / `push(mid|right)` loop, differing only in box-drawing characters. Extract a helper:
```rust
fn border_line(left: char, fill: char, mid: char, right: char, col_widths: &[usize]) -> String {
    let mut s = String::from(left);
    let last = col_widths.len().saturating_sub(1);
    for (i, &w) in col_widths.iter().enumerate() {
        s.push_str(&fill.to_string().repeat(w + 2));
        if i < last { s.push(mid); }
    }
    s.push(right);
    s.push('\n');
    s
}
```
Each call site collapses from 5 lines to 1.

**Finding 15 — Padding loops should use `" ".repeat(n)`** *(simplifier, low)*
`format_row` — `for _ in 0..padding { line.push(' '); }` is verbose. `line.push_str(&" ".repeat(padding))` is the idiomatic form; for `padding == 0` it's a no-op `push_str("")` so behavior is identical.

**Finding 16 — `Options::empty()` + `insert` is a verbose one-liner** *(simplifier, low)*
```rust
let mut options = Options::empty();
options.insert(Options::ENABLE_TABLES);
```
Should be `let options = Options::ENABLE_TABLES;`. A single bitflag *is* a valid `Options` value.

### Strengths

- **Byte-range substitution is correct.** Replacements are applied in reverse order, pulldown-cmark offsets are always valid char boundaries, and overlapping tables cannot occur in GFM. No panic paths from `replace_range`.
- **Column mismatch is handled safely.** `format_row` uses `.get(col)` throughout; body rows with fewer cells than the header gracefully render as padded blanks (truncation of *extra* cells is silent — see Finding 2).
- **`saturating_sub` on padding** correctly absorbs the case where a cell's display width exceeds the computed column width.
- **Partial/streaming tables don't corrupt surrounding text.** An incomplete table mid-stream is left as-is by pulldown-cmark (no `End(Table)` emitted), so no garbage substitution happens.
- **No `.unwrap()` anywhere.** Policy compliance confirmed.
- **Fast paths are correct.** No-pipe early return and empty-replacements fallback both work correctly.
- **`num_cols.max(1)` guard** handles a theoretically empty header gracefully.
- **All 568 tests pass.** `cargo fmt` and `cargo clippy` are clean.
- **All 6 acceptance criteria and all 5 plan deliverables are met.** Completeness check passes.

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

Review comment: https://github.com/Fizzizist/illustrious-manager/pull/164#issuecomment-4328317389

### Classifications

|---|---|---|
| Finding 1 — Alignment unit test makes no meaningful assertions | **Genuine issue** | `.contains("Right")` and `.contains("b")` pass even if all padding logic is deleted. The integration snapshot pins alignment, but the unit test verifies nothing about it. |
| Finding 2 — Mismatched column counts untested | **Genuine issue** | Extra body cells silently truncated; missing body cells silently padded. Both behaviors are real and untested. |
| Finding 3 — Bold/italic/link silently stripped, no test | **Genuine issue** | Defensible behavior (monospaced grid), but new code with zero coverage. A regression that drops the inner `Text` would be invisible. |
| Finding 4 — Partial/streaming table input untested | **Genuine issue** | `preprocess_tables` runs against every keystroke of `current_response`. Header+separator-only and mid-separator-truncated inputs are realistic streaming states with zero coverage. |
| Finding 5 — Redundant variable-declaration comments | **Genuine issue** | Direct CLAUDE.md policy violation. Easy mechanical fix. |
| Finding 6 — No test for inline code in cells | **Genuine issue** | `Event::Code` branch is distinct, hand-written, untested. |
| Finding 7 — No test for header-only table | **Genuine issue** | Valid GFM, structurally distinct path through `render_table`, untested. |
| Finding 8 — No test for single-column table | **Genuine issue** | The `if i + 1 < num_cols` joiner guard is structurally distinct; an off-by-one would go undetected. |
| Finding 9 — No User-role table snapshot | **Nitpick** | Roles only differ in label color and the `"  "` indent. The shared `render_role_lines` path is already covered by Assistant snapshots; adding User adds negligible confidence. |
| Finding 10 — No snapshot for streaming `current_response` table path | **Genuine issue** | If `preprocess_tables` were removed from `render_current_response_lines`'s call chain, no test would catch it. Distinct from Finding 4 (which is about parser-level behavior on partial input). |
| Finding 11 — CJK/emoji alignment not in a snapshot | **Genuine issue** | AC2 is currently verified only by a `│`-counting unit test that says nothing about actual padding. A snapshot is the only reliable check. |
| Finding 12 — ANSI escapes produce wrong widths | **Deferred** | Real but pre-existing limitation of `unicode-width` — not introduced by this PR. LLMs do not emit ANSI in table cells in practice. Tracking issue should be created. |
| Finding 13 — Wide-table overflow not documented in source | **Nitpick** | Doc-comment preference. Plan-comment record is sufficient. |
| Finding 14 — Four identical border-drawing loops | **Nitpick** | ~20 lines of duplication, but each loop reads clearly on its own. A helper would compress, not clarify. |
| Finding 15 — Padding loops should use `" ".repeat(n)` | **Nitpick** | Pure style. `push` in a loop is zero-allocation; `repeat` allocates. Neither is wrong. |
| Finding 16 — `Options::empty()` + `insert` is verbose | **Genuine issue** | Trivial mechanical simplification with zero downside. Worth fixing alongside Finding 5. |

### Counts
- **Genuine issues:** 11 (Findings 1, 2, 3, 4, 5, 6, 7, 8, 10, 11, 16)
- **Nitpicks:** 4 (Findings 9, 13, 14, 15)
- **False positives:** 0
- **Deferred:** 1 (Finding 12)

### Action Plan

Ordered by priority:

1. **Finding 1** — Strengthen `preprocess_respects_column_alignment` with assertions on the actual padded cell strings (e.g. `│     c │` for right-aligned `c` in a 5-char column).
2. **Finding 4 + Finding 10** — Add a snapshot test for a streaming/partial table (`current_response` path); add unit tests for header-only and mid-separator-truncated inputs.
3. **Finding 2** — Add tests for ragged rows: more columns than header (truncation) and fewer (padding).
4. **Finding 3** — Add a test pinning bold/italic/link stripping behavior (inner text preserved, markup discarded).
5. **Finding 6** — Add a test for `` `inline code` `` in a cell, asserting backticks are preserved in the rendered grid.
6. **Finding 7** — Add a test for a header-only table.
7. **Finding 8** — Add a test for a single-column table.
8. **Finding 11** — Add an insta snapshot for a table with CJK/emoji content, pinning exact column widths.
9. **Finding 5** — Delete the four redundant variable-declaration comments and the `// State machine` line.
10. **Finding 16** — Replace `Options::empty()` + `insert` with `let options = Options::ENABLE_TABLES;`.

Deferred (needs tracking issue): **Finding 12** — ANSI escape sequences in table cells. Pre-existing `unicode-width` limitation, not introduced by this PR. Tracking issue recommended.

---
*Assessment by mach6*

Fizzizist:
## Progress Update

Addressed all 11 genuine findings from the review assessment. Commit `f3489f4`:

**Tests added to `markdown_tables.rs`:**
- Finding 1 — `preprocess_respects_column_alignment` now asserts exact padded cell strings (`│     c │` for right-aligned, `│   b    │` for center-aligned)
- Finding 2 — Two new tests: body row with fewer columns pads with blanks; body row with more columns truncates extras
- Finding 3 — `preprocess_strips_inline_markup_preserving_text` pins bold/italic/link stripping (inner text kept, markup discarded)
- Finding 4 — `preprocess_header_only_partial_input_does_not_render_as_table` covers mid-separator truncation (incomplete GFM passes through unchanged)
- Finding 6 — `preprocess_inline_code_in_cell_preserves_backticks` covers the `Event::Code` branch
- Finding 7 — `preprocess_header_only_table_renders_without_body` covers valid GFM with no body rows
- Finding 8 — `preprocess_single_column_table` covers the no-joiner structural path

**Snapshots added to `conversation_area.rs`:**
- Finding 10 — `render_markdown_table_in_streaming_response` pins table rendering through the `current_response` path
- Finding 11 — `render_markdown_table_with_cjk_and_emoji` pins exact column widths for CJK/emoji content

**Code cleanup in `markdown_tables.rs`:**
- Finding 5 — Redundant variable-declaration comments and `// State machine` removed
- Finding 16 — `Options::empty()` + `insert` replaced with `let options = Options::ENABLE_TABLES`
- Finding 14 (nitpick, addressed anyway) — Extracted `border_line` helper eliminating four near-identical border-drawing loops

577 tests, 0 failures. `cargo fmt` and `cargo clippy` clean.

**Commit:** `f3489f4`

---
*Progress tracked by mach6*

PR #170, opened 2026-04-28
Wrap markdown table cells to viewport width (issue 167)

### LINKED ISSUE #167: markdown table looks bad at long string lengths
When the markdown table reaches a certain width, it starts to wrap and looks terrible. Can we have the inner text wrap instead of wrapping the table render?

Fizzizist:
Closes #167

Add width-aware GFM table preprocessing so wide tables wrap inside bounded columns instead of overflowing the conversation viewport. Fixes the box-drawing border corruption caused by `Paragraph::wrap` shredding the table mid-character.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1899,6 +1899,7 @@ dependencies = [
  "toml",
  "tui-markdown",
  "turso",
+ "unicode-segmentation",
  "unicode-width 0.2.2",
  "uuid",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ tokio-util = "0.7.18"
 toml = "1.1.0"
 tui-markdown = "0.3.7"
 turso = "0.5.3"
+unicode-segmentation = "1.13.2"
 unicode-width = "0.2.2"
 uuid = { version = "1.23.0", features = ["v7"] }
 
--- a/src/frontend/tui/conversation_area.rs
+++ b/src/frontend/tui/conversation_area.rs
@@ -1253,17 +1253,49 @@ mod tests {
 
     #[test]
     fn render_markdown_table_reflows_on_resize() {
+        use unicode_width::UnicodeWidthStr;
         let table = "| Column One | Column Two | Column Three | Column Four |\n\
                      |------------|------------|--------------|-------------|\n\
                      | Long value here | Another long value | Yet another long value | Final long value |";
         let mut entry = ConversationEntry::new(ConversationRole::Assistant, table.to_string());
         let wide_count = entry.wrapped_line_count(80);
+        let wide_render: String = entry
+            .lines()
+            .iter()
+            .map(|l| {
+                l.spans
+                    .iter()
+                    .map(|s| s.content.as_ref())
+                    .collect::<String>()
+            })
+            .collect::<Vec<_>>()
+            .join("\n");
         let narrow_count = entry.wrapped_line_count(40);
+        let narrow_render: String = entry
+            .lines()
+            .iter()
+            .map(|l| {
+                l.spans
+                    .iter()
+                    .map(|s| s.content.as_ref())
+                    .collect::<String>()
+            })
+            .collect::<Vec<_>>()
+            .join("\n");
         assert!(
             narrow_count >= wide_count,
             "narrow count {narrow_count} should be >= wide count {wide_count}"
         );
-        assert!(!entry.lines().is_empty(), "should have rendered lines");
+        assert_ne!(
+            wide_render, narrow_render,
+            "rendered content must differ between widths 80 and 40"
+        );
+        let wide_max = wide_render.lines().map(|l| l.width()).max().unwrap_or(0);
+        let narrow_max = narrow_render.lines().map(|l| l.width()).max().unwrap_or(0);
+        assert!(
+            narrow_max < wide_max,
+            "narrow max line width {narrow_max} should be < wide max {wide_max}"
+        );
     }
 
     #[test]
@@ -1274,8 +1306,32 @@ mod tests {
         let mut entry = ConversationEntry::new(ConversationRole::Assistant, table.to_string());
         let wide_count = entry.wrapped_line_count(80);
         let lines_at_80 = entry.lines().len();
+        let render_at_80: String = entry
+            .lines()
+            .iter()
+            .map(|l| {
+                l.spans
+                    .iter()
+                    .map(|s| s.content.as_ref())
+                    .collect::<String>()
+            })
+            .collect::<Vec<_>>()
+            .join("\n");
+
         let narrow_count = entry.wrapped_line_count(40);
         let lines_at_40 = entry.lines().len();
+        let render_at_40: String = entry
+            .lines()
+            .iter()
+            .map(|l| {
+                l.spans
+                    .iter()
+                    .map(|s| s.content.as_ref())
+                    .collect::<String>()
+            })
+            .collect::<Vec<_>>()
+            .join("\n");
+
         assert!(
             narrow_count >= wide_count,
             "narrow wrapped count {narrow_count} should be >= wide {wide_count}"
@@ -1283,6 +1339,30 @@ mod tests {
         assert!(
             lines_at_40 >= lines_at_80,
             "lines_at_40={lines_at_40} should be >= lines_at_80={lines_at_80}"
+        );
+        // Content must actually differ — a bug producing more lines at the same
+        // width would not change the rendered text.
+        assert_ne!(
+            render_at_80, render_at_40,
+            "cached_lines content must change after width-driven rebuild"
+        );
+        // Re-rendering at width 80 again should match the original wide render
+        // (cache is rebuilt cleanly, not corrupted).
+        let _ = entry.wrapped_line_count(80);
+        let render_at_80_again: String = entry
+            .lines()
+            .iter()
+            .map(|l| {
+                l.spans
+                    .iter()
+                    .map(|s| s.content.as_ref())
+                    .collect::<String>()
+            })
+            .collect::<Vec<_>>()
+            .join("\n");
+        assert_eq!(
+            render_at_80, render_at_80_again,
+            "rebuilding back to width 80 should produce identical content"
         );
     }
 }
--- a/src/frontend/tui/conversation_area.rs
+++ b/src/frontend/tui/conversation_area.rs
@@ -51,30 +51,33 @@ pub struct ConversationEntry {
     /// Per-turn 1-based index for tool use / result entries. `None` for other roles.
     pub tool_index: Option<usize>,
     cached_lines: Vec<Line<'static>>,
+    cached_lines_width: u16,
     cached_wrapped_count: u16,
     cached_width: u16,
 }
 
 impl ConversationEntry {
     pub fn new(role: ConversationRole, content: String) -> Self {
-        let cached_lines = render_entry_lines(&role, None, &content);
+        let cached_lines = render_entry_lines(&role, None, &content, 0);
         Self {
             role,
             content,
             tool_index: None,
             cached_lines,
+            cached_lines_width: 0,
             cached_wrapped_count: 0,
             cached_width: 0,
         }
     }
 
     pub fn new_indexed(role: ConversationRole, content: String, index: usize) -> Self {
-        let cached_lines = render_entry_lines(&role, Some(index), &content);
+        let cached_lines = render_entry_lines(&role, Some(index), &content, 0);
         Self {
             role,
             content,
             tool_index: Some(index),
             cached_lines,
+            cached_lines_width: 0,
             cached_wrapped_count: 0,
             cached_width: 0,
         }
@@ -114,6 +117,7 @@ impl ConversationEntry {
             content,
             tool_index: index,
             cached_lines: all_lines,
+            cached_lines_width: u16::MAX,
             cached_wrapped_count: 0,
             cached_width: 0,
         }
@@ -124,6 +128,12 @@ impl ConversationEntry {
     }
 
     pub fn wrapped_line_count(&mut self, text_width: u16) -> u16 {
+        if self.cached_lines_width != u16::MAX && self.cached_lines_width != text_width {
+            self.cached_lines =
+                render_entry_lines(&self.role, self.tool_index, &self.content, text_width);
+            self.cached_lines_width = text_width;
+            self.cached_width = 0;
+        }
         if text_width == self.cached_width && self.cached_width > 0 {
             return self.cached_wrapped_count;
         }
@@ -148,6 +158,7 @@ fn render_role_lines(
     index: Option<usize>,
     content: &str,
     trailing_blank: bool,
+    text_width: u16,
 ) -> Vec<Line<'static>> {
     let label = role_label(role, index);
     let mut lines = Vec::new();
@@ -156,7 +167,8 @@ fn render_role_lines(
         Style::default().fg(role.color()),
     )));
     let display_content = maybe_truncate(content, role);
-    let preprocessed = preprocess_tables(&display_content);
+    let budget = usize::from(text_width).saturating_sub(2);
+    let preprocessed = preprocess_tables(&display_content, budget);
     let rendered = markdown_to_text(&preprocessed);
     for line in rendered.lines {
         let mut prefixed = Line::from(Span::raw("  "));
@@ -177,12 +189,19 @@ fn render_entry_lines(
     role: &ConversationRole,
     index: Option<usize>,
     content: &str,
+    text_width: u16,
 ) -> Vec<Line<'static>> {
-    render_role_lines(role, index, content, true)
+    render_role_lines(role, index, content, true, text_width)
 }
 
-fn render_current_response_lines(current_response: &str) -> Vec<Line<'static>> {
-    render_role_lines(&ConversationRole::Assistant, None, current_response, false)
+fn render_current_response_lines(current_response: &str, text_width: u16) -> Vec<Line<'static>> {
+    render_role_lines(
+        &ConversationRole::Assistant,
+        None,
+        current_response,
+        false,
+        text_width,
+    )
 }
 
 fn estimate_wrapped_count(lines: &[Line<'_>], text_width: u16) -> u16 {
@@ -234,7 +253,7 @@ impl<'a> ConversationArea<'a> {
         let response_count = if self.current_response.is_empty() {
             0u16
         } else {
-            let response_lines = render_current_response_lines(self.current_response);
+            let response_lines = render_current_response_lines(self.current_response, text_width);
             estimate_wrapped_count(&response_lines, text_width)
         };
 
@@ -251,7 +270,7 @@ impl<'a> ConversationArea<'a> {
         let response_lines = if self.current_response.is_empty() {
             Vec::new()
         } else {
-            render_current_response_lines(self.current_response)
+            render_current_response_lines(self.current_response, text_width)
         };
         let response_count = if response_lines.is_empty() {
             0u16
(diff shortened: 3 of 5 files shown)


Fizzizist:
## Implementation Plan

### Problem

`markdown_tables::render_table` sizes columns to the longest cell unbounded. When total width exceeds the conversation viewport, `Paragraph::wrap { trim: false }` hard-wraps mid-character, shredding box-drawing borders. Fix: wrap **cell text** inside bounded columns so the rendered table always fits the available width.

### Approach

1. Make `preprocess_tables` width-aware: `preprocess_tables(input: &str, width: usize) -> String`.
2. In `render_table`, compute natural column widths, then if total exceeds budget, allocate via **proportional shrink** with a minimum-width floor.
3. Wrap each cell into multiple visual rows using a unicode-aware hard-char wrapper (grapheme-safe via `unicode-segmentation` + `unicode-width`).
4. `format_row` becomes `format_logical_row` and emits N visual lines per logical row, padding shorter wrapped cells with blanks.
5. Plumb `text_width` from the render path through to `preprocess_tables` for both persisted-entries and streaming paths.
6. Invalidate `ConversationEntry::cached_lines` when terminal width changes (today only `cached_wrapped_count` is width-keyed).

### Decisions (resolving assessment ambiguities)

1. **Allocation**: proportional shrink, ratio = `budget / sum(natural)`, floor each column at `MIN_COL_WIDTH`.
2. **Min column width**: `3` chars.
3. **Pathological narrow case**: if even mins × N + borders exceed budget, fall back to current unbounded behavior (overflow). Documented as known limitation.
4. **Wrap mode**: hard char-wrap on grapheme boundaries. No new dep.
5. **Indent accounting**: render path already passes `text_width` (= area.width − borders); subtract the 2-char role indent before passing to `preprocess_tables`. Helper: `table_budget = text_width.saturating_sub(2)`.

### Cache invalidation

`ConversationEntry::cached_lines` is built once at construction without width. Tables baked in will not reflow on resize. Fix:

- Track `cached_lines_width: u16` alongside `cached_lines`.
- In `wrapped_line_count(text_width)`, when `text_width != cached_lines_width`, **rebuild `cached_lines`** by re-running `render_entry_lines(role, index, content, text_width)`, then recompute the wrapped count.
- For `new_with_lines_indexed` (diff path — does not call `preprocess_tables`), set sentinel so it never rebuilds (diffs are pre-rendered Lines, not markdown).

### Deliverables

1. `preprocess_tables(input, width)` — new signature, width threaded into `render_table`.
2. New private helpers in `markdown_tables.rs`:
   - `allocate_widths(natural: &[usize], budget: usize) -> Vec<usize>` — proportional shrink with min floor; returns natural widths if budget already fits or if even mins do not fit.
   - `wrap_cell(text: &str, width: usize) -> Vec<String>` — grapheme-aware greedy hard-wrap respecting `UnicodeWidthStr`.
   - `format_logical_row(cells, col_widths, alignments) -> String` — emits N visual lines, padding short wrapped cells.
3. `ConversationEntry`: add width-keyed `cached_lines` rebuild path.
4. Update call sites in `render_role_lines` and `render_current_response_lines` to take and pass width.
5. Update `compute_entry_counts` / `render` / `max_scroll` to thread `text_width` (already available locally).

### Files to modify

- `src/frontend/tui/markdown_tables.rs` — width plumbing, new helpers, mechanical update of 17 existing tests, ~5 new tests (overflow proportional shrink, min-floor fallback, unicode wrap, single-column wrap, body-row missing-cell with wrap).
- `src/frontend/tui/conversation_area.rs` — thread `text_width` to `render_role_lines` / `render_current_response_lines`; invalidate `cached_lines` on width change in `wrapped_line_count`; add 2 new snapshot tests (narrow viewport with wide table, very narrow CJK).

### Acceptance criteria

1. A table whose natural width exceeds viewport renders within the viewport with wrapped cell text and intact box-drawing borders. Verifiable via new insta snapshot at width 40 with a wide table.
2. Existing 5 narrow-content snapshots remain byte-identical when re-run.
3. `cargo test`, `cargo clippy`, `cargo fmt --check` pass.
4. No `.unwrap()` introduced. All width arithmetic uses saturating ops.
5. Resizing the terminal causes tables to reflow (cache invalidation works) — verified by a unit test calling `wrapped_line_count` twice with different widths and inspecting `entry.lines()`.
6. Pathological narrow case (e.g. 8-column table at width 30) does not panic — falls back to overflow.

### Testing approach

**Unit tests (`markdown_tables.rs`)** — mechanical update of existing 17 tests to pass a generous width (e.g., `200`) so behavior is unchanged. New tests:

- `preprocess_wraps_wide_table_to_budget` — table natural width 80, budget 40, asserts no rendered line exceeds 40 visual cols.
- `allocate_widths_proportional_shrink` — direct test of allocator with known inputs.
- `allocate_widths_falls_back_when_minimums_exceed_budget` — returns natural widths so caller produces overflow rather than corruption.
- `wrap_cell_grapheme_safe` — emoji + CJK do not split mid-grapheme; widths correct.
- `preprocess_wide_table_keeps_borders_aligned` — every visual row has identical `│` count.

**Snapshot tests (`conversation_area.rs`)** — two new:

- `render_markdown_table_overflow_wraps_cells` — viewport 40 wide, table with natural width ~80; expect wrapped cells, intact borders.
- `render_markdown_table_reflows_on_resize` — render at width 80, then 40, asserting wrapped-line count differs (could also live as a unit test on `ConversationEntry`).

**Cache invalidation unit test** — `entry.wrapped_line_count(80)` then `entry.wrapped_line_count(40)` for an entry containing a wide table should produce a larger count at width 40, and `entry.lines()` should reflect the rewrapped layout.

### Risks / open questions

- **Snapshot churn**: confidence that narrow snapshots remain identical depends on a short-circuit `if total_natural <= budget { return natural }` *before* any allocator branch. Will add an explicit test for this.
- **`tui_markdown::from_str` re-parsing the fenced block**: emitting the same fenced-code-block structure with multi-line cells inside is treated as verbatim. Verified against current behavior.
- **Performance**: rebuilding `cached_lines` on every width change could be costly for long histories. Mitigated by single-shot rebuild + width-keyed cache. Resize is rare; full transcript re-render on resize is acceptable.
- **Per-grapheme width cost**: O(N) per cell per render. N is small. Acceptable.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implemented all plan deliverables for markdown table cell wrapping (issue 167).

**Changes:**

- `markdown_tables.rs` — `preprocess_tables` now takes a `width: usize` budget; `render_table` calls `allocate_widths` (proportional shrink, `MIN_COL_WIDTH=3`, overflow fallback) and `wrap_cell` (grapheme-cluster-aware hard-wrap via `unicode-segmentation`); `format_logical_row` emits N visual lines per logical row with correct alignment padding. Added `unicode-segmentation` dependency.
- `conversation_area.rs` — `ConversationEntry` gains `cached_lines_width: u16`; `wrapped_line_count` rebuilds `cached_lines` when `text_width` changes (cache invalidation on resize). `render_role_lines` threads `text_width.saturating_sub(2)` as table budget to `preprocess_tables`. `new_with_lines_indexed` (diff path) uses sentinel `u16::MAX` to skip rebuild.
- New snapshot: narrow (40-wide) viewport with wide 4-column table — cells wrap, box-drawing borders intact.
- Tests: 627 passing (24 in `markdown_tables`, including 7 new; 3 new in `conversation_area`). `cargo clippy` clean.

**Commit:** `7225cf8`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

### Critical

_None._

### Important

**Finding 1 — `allocate_widths` overshoots budget when natural widths are skewed** (code-reviewer + error-auditor, confidence 92–95)
**File:** `src/frontend/tui/markdown_tables.rs` (~lines 156–186)

The proportional shrink applies `.max(MIN_COL_WIDTH)` per column without redistributing the surplus that clamping creates. When one or more columns have small natural widths, they each get bumped up to 3, and the sum of the returned widths exceeds `budget_for_content`. The min-total guard (`num_cols × MIN_COL_WIDTH + overhead ≤ budget`) is necessary but not sufficient.

Repro: 5 columns, `natural = [1, 1, 100, 1, 1]`, `budget = 40`.
- `border_overhead = 16`, `budget_for_content = 24`, `min_total = 31 ≤ 40` → guard passes
- Proportional shares `[0, 0, 23, 0, 0]` → after clamp `[3, 3, 23, 3, 3]`, sum = 35
- Total rendered = 35 + 16 = **51**, exceeding budget of 40

Result: the table silently overflows the viewport — exactly what this PR is supposed to prevent. The existing `preprocess_wraps_wide_table_to_budget` test uses uniform column widths so it never hits this case. Fix: a second pass that redistributes clamped columns' surplus by trimming columns above `MIN_COL_WIDTH` until the sum fits.

**Finding 2 — Pathological narrow case lacks an end-to-end panic-safety test** (test-reviewer, confidence 95)
**File:** `src/frontend/tui/markdown_tables.rs`

Acceptance criterion 6 says "Pathological narrow case (e.g. 8-column table at width 30) does not panic — falls back to overflow." The only related test (`allocate_widths_falls_back_when_minimums_exceed_budget`) calls the allocator in isolation. There is no end-to-end test calling `preprocess_tables` with an 8-column table at width 30 to verify (a) no panic, (b) output is still a valid fenced code block. Add such a test.

**Finding 3 — Resize cache-invalidation tests inspect counts only, not `entry.lines()` content** (test-reviewer + completeness-checker, confidence 85–88)
**File:** `src/frontend/tui/conversation_area.rs`

Acceptance criterion 5 specifies: "verified by a unit test calling `wrapped_line_count` twice with different widths and **inspecting `entry.lines()`**." Neither `render_markdown_table_reflows_on_resize` nor `cached_lines_rebuild_on_width_change` examines the rendered string content between widths — they only assert line counts grow. A bug in the rebuild path that produces more lines but at the wrong width would still pass. Add an assertion that at least one `Line` at width 40 contains a shorter cell fragment than the corresponding line at width 80, or that the rendered content actually differs.

**Finding 4 — PR body says `Closes #131`; should be `Closes #167`** (completeness-checker, confidence 99)

Issue 131 ("interrupt keybinding") is already closed and unrelated. The actual subject of this PR is issue 167. Merging as-is leaves issue 167 open. Update the PR body to `Closes #167`.

### Suggestions

**Finding 5 — `wrap_cell` does not document the `width >= max_grapheme_width` invariant** (code-reviewer + error-auditor, confidence 82–85)
**File:** `src/frontend/tui/markdown_tables.rs` (~lines 192–219)

When `current` is empty and the next grapheme's display width exceeds `width`, the `!current.is_empty()` guard prevents flushing, so the grapheme is appended unconditionally — producing a line wider than `width`. Under normal allocation (`MIN_COL_WIDTH = 3`, max grapheme width = 2) this can't happen, but the overflow fallback in `allocate_widths` returns natural widths unchanged, which can be 1. A column with natural width 1 containing CJK text will silently overflow per cell. Add a `// invariant: width >= 2` comment or `debug_assert!`, or accept the overflow explicitly as documented best-effort fallback.

**Finding 6 — `cached_lines_width` sentinel `u16::MAX` collides with a legitimate `text_width = 65535`** (code-reviewer, confidence 80)
**File:** `src/frontend/tui/conversation_area.rs` (~lines 128–139)

`u16::MAX` is the sentinel meaning "diff entry, never rebuild." If `wrapped_line_count` is ever called with `text_width = 65535` on a normal entry, the rebuild correctly fires once, then sets `cached_lines_width = 65535`, after which the entry permanently looks like a diff entry — no future rebuild at any width. No real terminal hits 65535 columns, so this is theoretical, but the invariant is invisible. Use `Option<u16>` (None = externally managed) or a separate `bool is_externally_managed` field.

**Finding 7 — `allocate_widths` uses f64 where integer arithmetic suffices** (simplifier + code-reviewer, confidence 90)
**File:** `src/frontend/tui/markdown_tables.rs` (~lines 180–184)

```rust
let shrunk = (w as f64 * budget_for_content as f64 / natural_sum as f64) as usize;
```
Integer arithmetic gives identical floor rounding without two implicit conversions:
```rust
let shrunk = w * budget_for_content / natural_sum;
```
`natural_sum > 0` is guaranteed at that point (each natural width is `.max(1)`). For any realistic terminal width the multiplication cannot overflow `usize` on 64-bit targets.

**Finding 8 — `preprocess_wraps_wide_table_to_budget` does not assert content survived** (test-reviewer, confidence 85)
**File:** `src/frontend/tui/markdown_tables.rs`

Asserts `line_width <= 40 || line == "```"` but never asserts that the wrapped table still contains all original cell content fragments. A regression that silently drops cells or collapses to an empty box would still pass.

**Finding 9 — No test for alignment correctness after wrapping** (test-reviewer, confidence 90)
**File:** `src/frontend/tui/markdown_tables.rs`

`format_logical_row` applies alignment per sub-line. `preprocess_right_alignment` and `preprocess_center_alignment` only verify cell text presence in wide-budget output; neither exercises wrapping. Add a test for a right- or center-aligned column with a cell long enough to wrap, verifying each sub-line is correctly padded.

**Finding 10 — `wrap_cell_grapheme_safe` does not test wide-grapheme behavior at the wrap boundary** (test-reviewer, confidence 85)
**File:** `src/frontend/tui/markdown_tables.rs`

Existing test uses ASCII `"hello world test"`. Add tests where a 2-wide grapheme (CJK or emoji) sits exactly at the boundary — e.g. `wrap_cell("ab🦀cd", 4)` should produce `["ab🦀", "cd"]`, and `wrap_cell("ab🦀", 3)` should produce `["ab", "🦀"]`.

**Finding 11 — `allocate_widths_proportional_shrink` does not check per-column values** (test-reviewer, confidence 80)
**File:** `src/frontend/tui/markdown_tables.rs`

Asserts only sum and per-column min. A regression that returned `[MIN_COL_WIDTH, MIN_COL_WIDTH, MIN_COL_WIDTH]` for any input would pass. Assert proportional ratios for unequal-width inputs.

**Finding 12 — `render_entry_lines` is a one-line pass-through wrapper** (simplifier, confidence 88)
**File:** `src/frontend/tui/conversation_area.rs` (~lines 188–195)

The function exists only to hard-code `trailing_blank = true`. Inline at the three call sites; behavior is identical and the `trailing_blank` flag becomes locally visible.

**Finding 13 — `cached_lines_width` and `cached_width` track largely the same thing** (simplifier, confidence 85)
**File:** `src/frontend/tui/conversation_area.rs` (~lines 54–56, 131–143)

Two fields plus a `cached_width = 0` reset dance after rebuild. Unify into a single `Option<u16>` (or keep one `u16` field) — `None`/`u16::MAX` means "lines provided externally", `Some(w)` means "rendered + count cached at w". Removes one field and the reset.

**Finding 14 — `wrap_cell` manual mutable loop could be a `fold`** (simplifier, confidence 80)
**File:** `src/frontend/tui/markdown_tables.rs` (~lines 189–215)

The manual `current` / `current_width` accumulator pair can be expressed as `graphemes(true).fold(...)`, dropping one mutable variable. Stylistic only.

### Strengths

- Plan-driven implementation with clear deliverables and acceptance criteria, all of which were attempted (most were met).
- Saturating arithmetic used consistently — no risk of `usize` underflow panics in the new code.
- No `.unwrap()` introduced; complies with project policy.
- `unicode-segmentation` + `unicode-width` is the correct toolset for the wrapping problem.
- Cache invalidation on resize is a real correctness improvement over the prior code (which never rebuilt `cached_lines`).
- New snapshot test demonstrates the headline behavior end-to-end.
- 627 tests passing, clippy clean.

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

Review: https://github.com/Fizzizist/illustrious-manager/pull/170#issuecomment-4338510127

### Classifications

|---|---|---|
| 1 — `allocate_widths` overshoots budget on skewed naturals | genuine (important) | Math confirmed: `natural=[1,1,100,1,1], budget=40` → clamped `[3,3,23,3,3]`, sum=35, total=51 vs budget=40. The `min_total` guard checks pre-clamp feasibility, not post-clamp sum. Directly defeats the PR's core guarantee. |
| 2 — Pathological narrow case has no end-to-end panic test | genuine (important) | Existing test calls allocator in isolation. No `preprocess_tables(...)` call with 8-col + width-30 input. PR adds the fallback path; test must ship with it. |
| 3 — Resize cache tests inspect counts, not `entry.lines()` content | genuine (important) | Both tests assert only `narrow >= wide` line counts. Acceptance criterion 5 explicitly required content inspection. A rebuild bug producing more identical lines would pass. |
| 4 — PR body `Closes #131` should be `Closes #167` | genuine (important) | Confirmed in PR JSON. #131 is unrelated and already closed. Issue 167 stays open on merge unless body is fixed. |
| 5 — `wrap_cell` invariant undocumented | nitpick | Project policy minimizes comments. `MIN_COL_WIDTH=3` prevents the case in normal path; only the documented overflow fallback can hit it. Low value addition. |
| 6 — `u16::MAX` sentinel collides with width 65535 | nitpick | Confirmed real but no real terminal is 65535 cols. Theoretical only. `Option<u16>` is cleaner but not blocking. |
| 7 — f64 arithmetic where integer suffices | genuine (suggestion) | `natural_sum > 0` is guaranteed; `w × budget` cannot overflow `usize` on 64-bit. Integer version has identical floor rounding, no implicit casts, matches codebase style. |
| 8 — `preprocess_wraps_wide_table_to_budget` doesn't assert content survived | genuine (suggestion) | Width assertion alone passes if cells are silently dropped. One-line fix; new code in this PR. |
| 9 — No alignment-under-wrap test | genuine (suggestion) | `format_logical_row` applies alignment per sub-line; no test exercises this in the wrapped case. Bug-prone path, untested. |
| 10 — `wrap_cell_grapheme_safe` is ASCII-only | genuine (suggestion) | The grapheme-safe wrapper exists precisely for CJK/emoji; the test doesn't exercise the boundary case it was written to handle. |
| 11 — `allocate_widths_proportional_shrink` doesn't check per-column values | genuine (suggestion) | A degenerate impl returning `[3,3,3]` for any input passes. Sum + min checks are insufficient. |
| 12 — `render_entry_lines` is a one-line wrapper | nitpick | Inlining at 3 sites is feasible; the named wrapper also makes calls readable. Stylistic. |
| 13 — Two cache fields track similar things | nitpick | Fields serve subtly distinct roles: `cached_lines_width` guards expensive markdown re-render; `cached_width` guards cheap line-count recompute. Resetting `cached_width=0` after rebuild forces a count-only refresh. Unification possible but conflates two concerns. |
| 14 — `wrap_cell` could use `fold` | nitpick | `fold` would need a 3-tuple state, arguably less readable than named mutable variables. |

**Counts:** 7 genuine, 7 nitpicks, 0 false positives, 0 deferred.

### Action Plan

1. **(critical correctness)** Fix `allocate_widths` overshoot — add a redistribution second pass; trim columns above `MIN_COL_WIDTH` until `sum + overhead ≤ budget`. Add regression test with skewed naturals.
2. **(merge hygiene)** Edit PR body: `Closes #131` → `Closes #167`.
3. **(coverage)** Add end-to-end panic-safety test: `preprocess_tables(8-col table, width=30)` — must not panic; output must be valid.
4. **(coverage)** Strengthen cache-invalidation tests to inspect actual `entry.lines()` content at different widths, not just counts.
5. **(coverage)** Add wide-grapheme boundary tests: `wrap_cell("ab🦀cd", 4)` and `wrap_cell("ab🦀", 3)`.
6. **(coverage)** Add alignment-under-wrap test for right- and/or center-aligned columns.
7. **(coverage)** Strengthen `preprocess_wraps_wide_table_to_budget` with an assertion that cell content fragments survived.
8. **(coverage)** Strengthen `allocate_widths_proportional_shrink` to assert exact per-column values or use unequal inputs.
9. **(simplification)** Replace f64 cast with integer arithmetic in `allocate_widths`.
10. **(opportunistic)** Address nitpicks 5, 6, 12, 13, 14 if time permits — none are blocking.

---
*Assessment by mach6*

Fizzizist:
## Progress Update

Addressed all genuine findings from the review (1, 2, 3, 4, 7, 8, 9, 10, 11). Nitpicks (5, 6, 12, 13, 14) deferred.

**Critical correctness:**
- `allocate_widths` now redistributes surplus after the `MIN_COL_WIDTH` clamp so skewed natural widths (e.g. `[1,1,100,1,1]` @ budget 40) cannot overshoot the budget. Trims columns above the floor (largest first) until `sum + overhead ≤ budget`. Also switched from `f64` to integer arithmetic.
- New regression test: `allocate_widths_skewed_naturals_does_not_overshoot_budget`.

**Test coverage:**
- `preprocess_pathological_narrow_does_not_panic` — 8-col table at width 30 stays a valid fenced block.
- `preprocess_right_alignment_with_wrapping`, `preprocess_center_alignment_with_wrapping` — verify per-sub-line alignment in wrapped cells.
- `wrap_cell_wide_grapheme_at_boundary_fits`, `wrap_cell_wide_grapheme_exceeds_remaining_wraps`, `wrap_cell_cjk_at_boundary` — boundary cases for wide graphemes.
- `allocate_widths_proportional_shrink_unequal_naturals` — asserts proportional ordering, not just sum/floor.
- `preprocess_wraps_wide_table_to_budget` — now asserts every cell value survives wrapping.
- `cached_lines_rebuild_on_width_change` and `render_markdown_table_reflows_on_resize` now inspect actual `entry.lines()` content (not just counts); also verify re-rendering at the original width restores identical content.

**Merge hygiene:**
- PR body updated: `Closes #131` → `Closes #167`.

**Tests:** 635 passing. `cargo clippy` clean. `cargo fmt` clean.

**Commit:** `45ab11b`

---
*Progress tracked by mach6*

PR #185, opened 2026-05-03
Add active column for soft-delete readiness

### LINKED ISSUE #181: add soft-delete for session entries
Add an `active` column to session entries and have the conversation display only show `active` entries.
This issue will not add any functionality for soft-deleting entries. This issue just sets the stage for soft-deletes.

# Acceptance Criteria
- Add `active` boolean column to `conversation` table
- by default, everything is active
- Displayed conversation, is filtered by `active = 1`
- conversation sent to Agent is also filtered by `active = 1`

# User story
- User has a conversation with agent
- User manually runs `UPDATE conversation SET active = 0 where id = 2` from a `turnsodb` cli
- User re-opens the session and the entry with ID of 2 is no longer in the conversation display. The agent is also unaware of it.

Fizzizist:
Closes #181

Add an `active` boolean column to the `conversation` table, defaulting to `1`. Filter `load_history()` to only return active entries. No soft-delete UI — this is strictly a schema + filtering preparatory change.

### DIFF
--- a/src/session/conversation.rs
+++ b/src/session/conversation.rs
@@ -23,7 +23,7 @@ impl<'a> ConversationRepo<'a> {
         self.session
             .conn
             .execute(
-                "INSERT INTO conversation (role, content) VALUES (?1, ?2)",
+                "INSERT INTO conversation (role, content, active) VALUES (?1, ?2, 1)",
                 [
                     turso::Value::Text(role_str.to_string()),
                     turso::Value::Text(content_json),
@@ -57,7 +57,10 @@ impl<'a> ConversationRepo<'a> {
         let mut rows = self
             .session
             .conn
-            .query("SELECT role, content FROM conversation ORDER BY id ASC", ())
+            .query(
+                "SELECT role, content FROM conversation WHERE active = 1 ORDER BY id ASC",
+                (),
+            )
             .await
             .context("Failed to query conversation history")?;
 
@@ -113,5 +116,165 @@ impl<'a> ConversationRepo<'a> {
         }
 
         Ok(String::new())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::types::{Message, Role};
+    use tempfile::TempDir;
+
+    async fn create_test_session() -> (TempDir, super::super::Session) {
+        let dir = TempDir::new().expect("temp dir");
+        let session = super::super::Session::new(None, dir.path().to_path_buf())
+            .await
+            .expect("create session");
+        (dir, session)
+    }
+
+    #[tokio::test]
+    async fn load_history_filters_out_inactive_entries() {
+        let (_dir, session) = create_test_session().await;
+
+        session
+            .conversation()
+            .insert_message(&Message::text(Role::User, "active one".to_string()))
+            .await
+            .expect("insert 1");
+        session
+            .conversation()
+            .insert_message(&Message::text(Role::Assistant, "inactive".to_string()))
+            .await
+            .expect("insert 2");
+        session
+            .conversation()
+            .insert_message(&Message::text(Role::User, "active two".to_string()))
+            .await
+            .expect("insert 3");
+
+        session
+            .conn
+            .execute("UPDATE conversation SET active = 0 WHERE id = 2", ())
+            .await
+            .expect("deactivate row 2");
+
+        let history = session
+            .conversation()
+            .load_history()
+            .await
+            .expect("load history");
+        assert_eq!(history.len(), 2, "should only return active entries");
+    }
+
+    #[tokio::test]
+    async fn load_history_returns_all_when_all_active() {
+        let (_dir, session) = create_test_session().await;
+
+        session
+            .conversation()
+            .insert_message(&Message::text(Role::User, "a".to_string()))
+            .await
+            .expect("insert 1");
+        session
+            .conversation()
+            .insert_message(&Message::text(Role::Assistant, "b".to_string()))
+            .await
+            .expect("insert 2");
+        session
+            .conversation()
+            .insert_message(&Message::text(Role::User, "c".to_string()))
+            .await
+            .expect("insert 3");
+
+        let history = session
+            .conversation()
+            .load_history()
+            .await
+            .expect("load history");
+        assert_eq!(history.len(), 3);
+    }
+
+    #[tokio::test]
+    async fn insert_message_sets_active_to_1() {
+        let (_dir, session) = create_test_session().await;
+
+        session
+            .conversation()
+            .insert_message(&Message::text(Role::User, "check active".to_string()))
+            .await
+            .expect("insert");
+
+        let mut rows = session
+            .conn
+            .query(
+                "SELECT active FROM conversation ORDER BY id DESC LIMIT 1",
+                (),
+            )
+            .await
+            .expect("query active column");
+        let row = rows
+            .next()
+            .await
+            .expect("get row")
+            .expect("should have a row");
+        match row.get_value(0).expect("get active value") {
+            turso::Value::Integer(n) => assert_eq!(n, 1, "active column should be 1"),
+            other => panic!("expected integer, got {:?}", other),
+        }
+    }
+
+    #[tokio::test]
+    async fn is_empty_counts_inactive_entries() {
+        let (_dir, session) = create_test_session().await;
+
+        session
+            .conversation()
+            .insert_message(&Message::text(Role::User, "will deactivate".to_string()))
+            .await
+            .expect("insert");
+
+        session
+            .conn
+            .execute("UPDATE conversation SET active = 0 WHERE id = 1", ())
+            .await
+            .expect("deactivate");
+
+        let empty = session.conversation().is_empty().await.expect("is_empty");
+        assert!(
+            !empty,
+            "is_empty should return false because the row still exists regardless of active"
+        );
+    }
+
+    #[tokio::test]
+    async fn read_first_user_message_ignores_active_filter() {
+        let (_dir, session) = create_test_session().await;
+
+        session
+            .conversation()
+            .insert_message(&Message::text(Role::User, "first".to_string()))
+            .await
+            .expect("insert 1");
+        session
+            .conversation()
+            .insert_message(&Message::text(Role::User, "second".to_string()))
+            .await
+            .expect("insert 2");
+
+        session
+            .conn
+            .execute("UPDATE conversation SET active = 0 WHERE id = 1", ())
+            .await
+            .expect("deactivate first");
+
+        let msg = session
+            .conversation()
+            .read_first_user_message()
+            .await
+            .expect("read first user message");
+        assert_eq!(
+            msg, "first",
+            "should return the absolute first user message, not filtered by active"
+        );
     }
 }
--- a/src/session/mod.rs
+++ b/src/session/mod.rs
@@ -128,28 +128,7 @@ impl Session {
             .await
             .context("Failed to run schema DDL")?;
 
-        {
-            let mut rows = conn
-                .query("PRAGMA table_info(conversation)", ())
-                .await
-                .context("Failed to query conversation table info")?;
-            let mut has_active = false;
-            while let Some(row) = rows.next().await? {
-                if let turso::Value::Text(name) = row.get_value(1)?
-                    && name == "active"
-                {
-                    has_active = true;
-                    break;
-                }
-            }
-            if !has_active {
-                conn.execute_batch(
-                    "ALTER TABLE conversation ADD COLUMN active INTEGER NOT NULL DEFAULT 1",
-                )
-                .await
-                .context("Failed to add active column to conversation table")?;
-            }
-        }
+        migrate_active_column(&conn).await?;
 
         Ok(Self {
             id: sess_id,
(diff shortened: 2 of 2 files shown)


Fizzizist:
## Implementation Plan — Add `active` column for soft-delete readiness

### Analysis

Issue 181 requests adding an `active` boolean column to the `conversation` table, defaulting to `1`, and filtering all displayed/agent-bound conversation history to only `active = 1` rows. No soft-delete UI or mechanism is in scope — this is purely a schema + filtering preparatory change.

The current `conversation` table has 3 columns: `id`, `role`, `content`. There is no migration system — schema is applied via `CREATE TABLE IF NOT EXISTS` on every `Session::new()`. Since `CREATE TABLE IF NOT EXISTS` won't add columns to an existing table, we need an idempotent `ALTER TABLE` migration.

### Deliverables

1. **Schema update**: Add `active INTEGER NOT NULL DEFAULT 1` to the `conversation` table DDL in `SCHEMA`
2. **Migration**: Add an idempotent migration that adds the `active` column to existing DBs (guarded by `PRAGMA table_info` check)
3. **`insert_message()`**: Explicitly include `active` column with value `1` in the INSERT statement
4. **`load_history()`**: Add `WHERE active = 1` filter
5. **`session_history()`** (in agent): Already delegates to `load_history()`, so no change needed — filtering is automatic
6. **`is_empty()`**: No filter (counts all entries, governs deletion semantics)
7. **`read_first_user_message()`**: No filter (reads the absolute first user message regardless of status)
8. **Tests**: Add new tests for the `active` column behavior in `src/session/mod.rs` and create a new test module in `src/session/conversation.rs`

### Files to Modify

| File | Changes |
|---|---|
| `src/session/mod.rs` | Update `SCHEMA` constant; add idempotent `ALTER TABLE` migration in `Session::new()` |
| `src/session/conversation.rs` | Update `insert_message()` to include `active = 1`; update `load_history()` to filter `WHERE active = 1`; add `#[cfg(test]` module with tests |

### Acceptance Criteria

- New sessions get a `conversation` table with `active INTEGER NOT NULL DEFAULT 1` column
- Existing sessions (created before this change) automatically get the `active` column added on next open, with all existing rows defaulting to `1`
- `load_history()` only returns rows where `active = 1`
- `insert_message()` explicitly sets `active = 1` on every insert
- `is_empty()` still counts all rows regardless of `active` status
- `read_first_user_message()` still reads the absolute first user message regardless of `active` status
- Agent's `session_history()` returns only active entries (since it delegates to `load_history()`)
- Agent's `Agent::new()` and `load_session()` both use `load_history()`, so both get filtered results automatically
- All existing tests continue to pass
- New tests verify the filtering behavior

### Testing Approach

**New tests in `src/session/conversation.rs`** (creating a `#[cfg(test)] mod tests` — currently absent):

1. **`load_history_filters_out_inactive_entries`** — Insert 3 messages, manually set one to `active = 0` via raw SQL, assert `load_history()` returns only 2
2. **`load_history_returns_all_when_all_active`** — Insert 3 messages, assert all 3 returned (regression check)
3. **`insert_message_sets_active_to_1`** — Insert a message, query the `active` column directly, assert it's `1`
4. **`is_empty_counts_inactive_entries`** — Insert a message, set it to `active = 0`, assert `is_empty()` returns `false` (inactive entries still count)
5. **`read_first_user_message_ignores_active_filter`** — Insert two user messages, deactivate the first, assert `read_first_user_message()` still returns the original first message

**New tests in `src/session/mod.rs`:**

6. **`active_column_migration_adds_column_to_existing_db`** — Create a DB with the old schema (no `active` column), open it with `Session::new()`, verify the `active` column now exists and defaults to `1`

**Test infrastructure:** All tests use the existing `TempDir` + `Session::new()` pattern. No new test helpers needed. For migration testing, we'll create a DB with the old schema (without `active`), then open it with `Session::new()` to trigger the migration.

### Risks and Open Questions

1. **Migration idempotency** — SQLite doesn't support `IF NOT EXISTS` on `ALTER TABLE ADD COLUMN`. We must check `PRAGMA table_info(conversation)` for the `active` column before running `ALTER TABLE`. This is the same pattern used for the `task` table addition in the existing `task_table_is_added_on_reopen_of_existing_db` test.
2. **ToolUse/ToolResult pairing** — If a future issue adds soft-delete UI, deactivating a ToolUse message while leaving the corresponding ToolResult active (or vice versa) would violate the Anthropic API contract. This is out of scope for this issue but the design should be aware of it.
3. **`is_empty()` semantics preserved** — Sessions with all entries manually deactivated still appear as "non-empty" in `is_empty()`, which is correct for deletion semantics.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implemented all 8 deliverables from the plan for adding the `active` column to the `conversation` table:

- **Schema**: Added `active INTEGER NOT NULL DEFAULT 1` to the `conversation` table DDL
- **Migration**: Added idempotent `ALTER TABLE` migration guarded by `PRAGMA table_info` check for existing DBs
- **`insert_message()`**: Explicitly sets `active = 1` on every insert
- **`load_history()`**: Filters with `WHERE active = 1`
- **`is_empty()`**: Unchanged — counts all rows regardless of `active` status
- **`read_first_user_message()`**: Unchanged — reads absolute first user message regardless of `active`
- **`session_history()`**: Unchanged — delegates to `load_history()`, filtering is automatic

Added 6 new tests:
- `load_history_filters_out_inactive_entries`
- `load_history_returns_all_when_all_active`
- `insert_message_sets_active_to_1`
- `is_empty_counts_inactive_entries`
- `read_first_user_message_ignores_active_filter`
- `active_column_is_added_on_reopen_of_existing_db`

All 732 tests pass. `cargo fmt` and `cargo clippy` clean.

**Commit:** `4dce763`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

### Critical
None.

### Important

**1. `load_history_filters_out_inactive_entries` only checks count, never validates content**
The test inserts three messages, deactivates the middle one, and asserts `history.len() == 2`. It never verifies *which* two entries were returned. If the `WHERE active = 1` clause were somehow buggy and filtered the wrong rows, the count assertion would still pass coincidentally. The test should assert the actual text content of the returned messages.

**2. `is_empty()` and `load_history()` have divergent semantics around `active` — zombie sessions**
`load_history()` filters `WHERE active = 1`, but `is_empty()` counts all rows regardless of `active`. The sole production caller is `cleanup_empty_session()` in `agent.rs:191`. If all messages in a session are soft-deleted (all `active = 0`), `load_history()` returns empty but `is_empty()` returns `false`, so the DB file is preserved forever with no visible content and no cleanup path. Either `is_empty` should filter by `WHERE active = 1`, or it should be renamed to make its different semantics explicit (e.g., `has_any_rows`).

**3. `read_first_user_message()` ignores `active` filter — stale session picker previews**
`read_first_user_message()` queries without `WHERE active = 1`, so the session picker displays the original first message even after it's been soft-deleted. The user opens a session expecting content that `load_history()` doesn't return. This is tested and intentional, but creates a UX inconsistency once soft-delete is wired up.

**4. No test for `load_history` when all entries are inactive**
There is no test covering the edge case where every row has `active = 0`. This is a realistic scenario once soft-delete is wired up. Should verify `load_history()` returns an empty `Vec<Message>` (not an error).

### Suggestions

**5. `INSERT` redundantly specifies `active` column that has `DEFAULT 1`**
The INSERT `INSERT INTO conversation (role, content, active) VALUES (?1, ?2, 1)` includes `active` explicitly, but the schema already declares `active INTEGER NOT NULL DEFAULT 1`. Since new inserts should always be active, the `DEFAULT 1` already encodes that contract. Dropping `active` from the INSERT would reduce noise — the current form misleadingly suggests the value might sometimes differ.

**6. Migration logic inlined in `Session::new` obscures the function's purpose**
~20 lines of `PRAGMA table_info` + conditional `ALTER TABLE` are inlined in `Session::new`. Extracting into `fn migrate_schema(conn: &Connection) -> Result<()>` would keep `Session::new` focused, give the migration a self-documenting name, and make future migrations easier to add.

**7. `active_column_is_added_on_reopen_of_existing_db` doesn't directly verify the `active` column value**
The migration test asserts `history.len() == 1` but never directly checks that `active = 1` on the migrated row. While the assertion would catch `DEFAULT 0` (since `load_history` would return 0 rows), the test should explicitly verify the schema change by querying the `active` column value.

**8. `load_history_returns_all_when_all_active` duplicates pre-existing test with weaker assertions**
The existing `insert_and_load_multiple_messages` test already inserts 3 messages and asserts length, roles, and order. The new test does the same but *omits* role assertions, adding no meaningful new coverage.

### Strengths
- Clean, focused, well-scoped implementation that matches the plan exactly
- Idempotent migration pattern using `PRAGMA table_info` is correct and follows existing patterns
- Explicit test for `is_empty` ignoring `active` and `read_first_user_message` ignoring `active` — these behavioral contracts are well-documented in tests
- Good test coverage for the core filtering behavior
- No `.unwrap()` calls — all errors properly propagated with `.context()`

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

[Review comment](https://github.com/Fizzizist/illustrious-manager/pull/185#issuecomment-4366445955)

### Classifications

|---|---|---|
| load_history test only checks count | Nitpick | With `WHERE active = 1 ORDER BY id ASC`, the wrong rows can't be returned — row 2 is excluded by the filter. Content assertions would marginally strengthen it but aren't needed for correctness. |
| is_empty / load_history divergent semantics → zombie sessions | **Genuine issue** | `is_empty()` counts all rows; `load_history()` filters `WHERE active = 1`. `cleanup_empty_session()` uses `is_empty()`, so fully soft-deleted sessions are never garbage-collected. Real semantic inconsistency with a concrete impact path. |
| read_first_user_message ignores active filter → stale picker | **Genuine issue** | Session picker shows a message that's been soft-deleted, but opening the session shows empty history. UX inconsistency. The existing test documents the intent, but the intent conflicts with `load_history()` semantics. |
| No test for all-inactive load_history | Nitpick | `WHERE active = 1` returning empty on all-inactive is straightforward SQL. Low regression risk, but the edge case is untested. |
| INSERT redundantly specifies active | False positive | Explicit `active = 1` in INSERT is defensive — it documents intent and won't silently change if someone alters the default. |
| Migration inlined in Session::new | Nitpick | 17-line block scoped in braces is readable enough. Single-use extraction adds indirection for no clear benefit. |
| Migration test doesn't directly verify active value | False positive | `load_history()` uses `WHERE active = 1`, so `history.len() == 1` indirectly verifies the correct default. Dedicated `insert_message_sets_active_to_1` test already provides direct verification. |
| load_history_returns_all_when_all_active duplicates test | Nitpick | Tests different code paths (ConversationRepo vs Session), but the weaker assertions mean it adds marginal value. |

### Action Plan

1. **Fix `is_empty()` to filter by `WHERE active = 1`** — Change the SQL in `is_empty()` to `SELECT COUNT(*) FROM conversation WHERE active = 1`. Update `is_empty_counts_inactive_entries` test to expect `true` when only inactive entries exist. This fixes the zombie-session problem and aligns semantics with `load_history()`.

2. **Fix `read_first_user_message()` to filter by `WHERE active = 1`** — Add `AND active = 1` to the query. Update `read_first_user_message_ignores_active_filter` test to verify the new behavior (returning the first *active* user message, not the absolute first). This prevents stale session picker previews.

3. **(Optional) Add content assertions to `load_history_filters_out_inactive_entries`** — Verify the returned messages are "active one" and "active two" for a more robust test.

4. **(Optional) Add test for all-inactive history returning empty** — Edge case coverage for `load_history()` when every row has `active = 0`.

---
*Assessment by mach6*

Fizzizist:
## Review Assessment (Revised)

[Original review comment](https://github.com/Fizzizist/illustrious-manager/pull/185#issuecomment-4366445955)

### Classifications

|---|---|---|
| load_history test only checks count | Nitpick | Count assertion verifies filtering; `WHERE active = 1` makes wrong-row return logically impossible. Content assertions would marginally strengthen but aren't needed. |
| is_empty / load_history divergent semantics → zombie sessions | **False positive** | Explicitly intentional per PR plan. `is_empty()` governs deletion semantics — a session with only soft-deleted rows still occupies disk and must not be auto-deleted. `load_history()` filters for display. These are two different operations with two different correct behaviors. No divergence — deliberate, justified asymmetry. |
| read_first_user_message ignores active → stale picker | **False positive** | Explicitly intentional per PR plan. The session picker must identify sessions by their first message regardless of deletion status. Hiding this would *prevent* users from finding old runs — the opposite of the desired UX. |
| No test for all-inactive load_history | Nitpick | Missing edge-case test, not a bug. `WHERE active = 1` returning empty on all-inactive is deterministic SQL. Low regression risk. |
| INSERT redundantly specifies active | False positive | Explicit `active = 1` documents intent and won't silently change if the default is altered. Defensive practice. |
| Migration inlined in Session::new | Nitpick | Scoped block is readable; single-use extraction adds indirection. Defer if migrations grow. |
| Migration test doesn't directly verify active value | False positive | `load_history()` with `WHERE active = 1` indirectly verifies the default; `insert_message_sets_active_to_1` provides direct verification. |
| load_history_returns_all_when_all_active duplicates test | Nitpick | Tests different code paths (ConversationRepo vs Session). Weaker assertions add marginal value. |

### Action Plan

**Zero genuine issues.** The two findings originally classified as genuine were false positives — they misread deliberate design decisions documented in the PR plan.

Optional improvements (all nitpicks, none required for merge):

1. Add `load_history_returns_empty_when_all_inactive` test — fills a genuine edge-case gap with low cost.
2. (Optional) Add content assertions to `load_history_filters_out_inactive_entries`.

---
*Assessment by mach6 (revised)*

Fizzizist:
## Progress Update

Extracted the inline migration block from `Session::new()` into a standalone `migrate_active_column()` function. The brace-scoped block was doing a distinct job and obscured the flow of the constructor — now it reads as a single clear call.

**Commit:** `11ea930`

---
*Progress tracked by mach6*

PR #190, opened 2026-05-05
feat: auto-compaction when context exceeds threshold

### LINKED ISSUE #183: auto-compaction
Auto-compact session context when context reaches a certain threshold defined by the user. 

# Acceptance Criteria
- new config option: `max_context_window_len` for determining when to auto-compact the conversation
- use existing compaction infrastructure used with `/compact` command
- auto-compact when a usage event comes through that pushes the context over the threshold
- Clear info message to the user about context being auto-compacted including the current usage, and max context window length

Fizzizist:
Closes #183

Add automatic context compaction when conversation token usage exceeds a user-configurable threshold (`max_context_window_len`). Reuses existing compaction infrastructure (`/compact`), adding only a trigger mechanism, storm guard, and user-facing info message.


Fizzizist:
## Implementation Plan

### Problem

Context grows unbounded during long sessions. Manual `/compact` exists, but users must remember to invoke it. When the API context window is exceeded, requests fail.

### Deliverables

1. **`max_context_window_len` config option** — `u32` in `ToolsConfig`, default `0` (disabled). Non-zero value = token threshold for auto-compaction.
2. **Auto-compact trigger** — After `ResponseComplete`, if `last_input_total > max_context_window_len` and not already auto-compacted last turn, trigger compaction automatically.
3. **Info message** — Push a `ConversationRole::Info` entry to the conversation showing current usage vs. limit before auto-compact begins.
4. **Storm guard** — `auto_compacted_last_turn: bool` field on `App`, set `true` after auto-compact, reset `false` on the next `ResponseComplete` that doesn't trigger auto-compact. Prevents infinite compaction loops.
5. **Tests** — Config parsing, trigger logic, storm guard, info message.

### Acceptance Criteria

- `max_context_window_len = 0` → auto-compact never triggers (default).
- `max_context_window_len > 0` → after each complete assistant turn, if `last_input_total > max_context_window_len`, auto-compact triggers.
- Info message shown: `"Auto-compacting context (current: {last_input_total} tokens, limit: {max_context_window_len} tokens)"`
- Compaction storm guard prevents consecutive auto-compactions.
- `/compact` remains manual and unaffected.
- Single-shot mode unaffected.

### Files to Modify

| File | Change |
|------|--------|
| `src/config.rs` | Add `max_context_window_len: u32` to `ToolsConfig` (default `0`), update template, add tests |
| `src/frontend/tui/tui_app.rs` | Add `auto_compacted_last_turn: bool` to `App`. After `ResponseComplete`, check threshold and spawn auto-compact. After `CompactionComplete`, reset usage and reload. Update `reset_for_session_switch` to reset the flag. Add tests. |

### Design Decisions

- **Trigger point**: After `ResponseComplete` in `handle_agent_event`, set a flag on `App` indicating auto-compact should trigger. The `run_app` loop reads this flag and initiates compaction (needs access to `agent` and `event_tx`).
- **Reuse existing compaction flow**: The auto-compact path calls the exact same `agent.compact()` + `CompactionComplete` event handling as `/compact`. No new event type needed.
- **Info message before compacting**: Push `ConversationRole::Info` entry, then transition to `AppState::Compacting` and spawn the task.
- **Storm guard**: After auto-compact, `auto_compacted_last_turn = true`. On the next `ResponseComplete`, if the flag is true and threshold is still exceeded, skip auto-compact. If the flag is true and threshold is NOT exceeded, reset to `false`.

### Testing Approach

- **Config**: Parse `max_context_window_len = 100000`, verify default is `0`, verify absent key defaults to `0`.
- **Auto-compact trigger**: Test `should_auto_compact()` logic: threshold > 0, input exceeds threshold, not already auto-compacted → true.
- **Storm guard**: Test that `auto_compacted_last_turn = true` prevents auto-compact even if threshold is still exceeded.
- **Info message**: Verify the message format with current usage and limit.
- **Reset on session switch**: Verify `auto_compacted_last_turn` is reset.

### Risks

1. **Compaction during tool-use loop** — Mitigated by only triggering after `ResponseComplete` (agent is idle).
2. **Post-compaction context still exceeds limit** — Mitigated by storm guard; user gets one auto-compact, then must act manually.
3. **Config validation** — `max_context_window_len` should be a non-negative integer. `0` = disabled.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implemented the auto-compaction feature per the plan in this PR. All deliverables are complete:

**Deliverable 1 — Config option:** Added `max_context_window_len: u32` to `ToolsConfig` (default `0` = disabled). Updated config template. Added 3 tests for config parsing.

**Deliverable 2 — Auto-compact trigger + storm guard:** Added `auto_compacted_last_turn: bool` to `App` with `should_auto_compact()` method. Wired trigger after `ResponseComplete` in `run_app` — pushes Info message, transitions to `Compacting`, spawns `agent.compact()`. Storm guard preserved across `CompactionComplete` by saving/restoring `auto_compacted_last_turn`. If threshold not exceeded after a guarded turn, flag resets. Added 6 tests for trigger logic, storm guard, info message, and session reset.

**Verification:** 786 unit tests + all integration tests pass. `cargo clippy` and `cargo fmt` clean.

**Commit:** `1f05f8c`

---
*Progress tracked by mach6*

Fizzizist:
## Progress Update

Merged trunk and resolved conflict in `src/config.rs`. The merge brought in the openai_compat backend feature (#191) which added new types (`OpenAiCompatConfigToml`, `ReasoningStyleConfig`) and config tests. My `max_context_window_len` changes were cleanly integrated alongside the new openai_compat tests.

All 813 tests pass. `cargo clippy` and `cargo fmt` clean.

**Commit:** `9832640`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

### Critical

**1. Compaction may not reduce context below threshold → infinite auto-compact loop** (confidence: 95)
If auto-compaction succeeds but the compacted context (summary + context prefix) still exceeds `max_context_window_len`, the storm guard clears after one turn, and the system auto-compacts again on the next `ResponseComplete`. This repeats indefinitely: compact → guard → clear → compact → guard → clear. Each cycle burns API tokens on the compaction LLM call. If the compaction summary itself is large or the context prefix is huge, the user gets stuck in this loop with no way out other than manually raising the limit or disabling auto-compact.

**2. Permanent compaction failure → infinite retry loop** (confidence: 85)
If compaction fails (e.g., misconfigured compaction role, persistent network error), the `saved_auto_compacted` restore keeps the storm guard at `true`. On the next `ResponseComplete`, `should_auto_compact` returns `false` (storm-guarded), the `else if` branch clears the guard, and the turn after that triggers auto-compact again. This creates an infinite retry loop with no backoff or max-retry limit, making the TUI unusable.

### Important

**3. `last_input_total` is estimated (not real) after compaction** (confidence: 82)
After `CompactionComplete`, `reset_for_session_switch()` zeroes `last_input_total`, and `load_history()` re-estimates it from character-count heuristics (±30-50% off from real token counts). Between compaction completion and the next real `Usage` event, `should_auto_compact` operates on inaccurate data. This can cause the storm guard to clear prematurely (estimated value below threshold, real value above) or delay compaction (estimated value above, real value below). The counter self-corrects once the API returns a `Usage` event, but the decision at the critical `ResponseComplete` moment may be wrong.

**4. Auto-compact is TUI-only — stdout frontend silently lacks the feature** (confidence: 92)
The entire auto-compaction feature lives in `src/frontend/tui/tui_app.rs`. `src/frontend/stdout.rs` has no `max_context_window_len` handling, no auto-compact trigger, and no `CompactionComplete` handling. If users configure `max_context_window_len` expecting it to work globally, single-shot/multi-turn stdout usage will silently exceed the threshold without auto-compacting. This may be acceptable for single-shot, but should be documented.

**5. No validation floor on `max_context_window_len`** (confidence: 85)
A user can set `max_context_window_len = 1`, causing auto-compact to trigger after literally every API turn. There's no minimum-value check in `validate()`. The config template comment says "0 = disabled" but doesn't advise a reasonable range.

**6. Auto-compact spawn has no cancellation token** (confidence: 88)
The `tokio::spawn` block for auto-compact creates a fire-and-forget task. On TUI exit, `compaction_task.abort()` cancels the `JoinHandle`, but the underlying HTTP connection in `agent.compact()` may linger. This is the same pattern as manual `/compact` (not a regression), but auto-compact makes it more likely to be hit since it fires automatically.

### Suggestions

**7. `saved_auto_compacted` save/restore pattern is misleading** (confidence: 90)
The `CompactionComplete` handler saves `auto_compacted_last_turn` before calling `reset_for_session_switch()` (which sets it to `false`), then restores it. The save/restore exists solely because `reset_for_session_switch` is the wrong operation for compaction completion — it's a session-switch reset being reused where a lighter "reload history" operation is needed. A dedicated `reload_history` method that doesn't touch `auto_compacted_last_turn` would eliminate the save/restore entirely and make the intent clear.

**8. Storm guard `else if` branch is a consequence of inverted control flow** (confidence: 85)
The `else if app.auto_compacted_last_turn { app.auto_compacted_last_turn = false; }` branch clears the guard when `should_auto_compact` returns false. Simpler: unconditionally clear `auto_compacted_last_turn` at the *start* of each `ResponseComplete`, before the `should_auto_compact` check. If compaction triggers, set it back to `true`. If not, it stays `false`. This inverts the flow to a clearer "reset-then-possibly-set" pattern and eliminates the `else if`.

**9. `should_auto_compact(max_context_window_len)` takes an external parameter for a single call site** (confidence: 80)
The method requires the caller to pass `config.tools.max_context_window_len`, which `App` doesn't own. Either inline the three-condition check (it's only called from one place) or move `max_context_window_len` onto `App` as a field, making `should_auto_compact()` parameterless and a true property of `App`'s state.

**10. Core trigger flow in `run_app` is untested** (confidence: 95)
All six TUI app tests exercise `should_auto_compact()` or `reset_for_session_switch()` in isolation. None test the actual behavioral sequence inside the `tokio::select!` loop: pushing the Info entry, setting `auto_compacted_last_turn`, transitioning to `Compacting`, spawning the task, and assigning `compaction_task`. The integration path — where the feature actually works — has zero test coverage.

**11. Storm guard reset branch (`else if`) is untested** (confidence: 90)
The `else if app.auto_compacted_last_turn` branch that clears the storm guard when the threshold is no longer exceeded has zero test coverage. A regression that breaks the reset would go undetected.

**12. `CompactionComplete` recovery path is untested for auto-compaction** (confidence: 90)
The `saved_auto_compacted` preservation mechanism — which exists specifically to keep the storm guard across the compaction boundary — is untested. A regression that drops the restore would cause immediate double-compaction.

**13. `compaction_task` lifecycle is untested** (confidence: 85)
No test verifies `compaction_task` is `Some` after auto-compact trigger, `None` after `CompactionComplete`, or correctly aborted on exit.

### Strengths

- Clean, minimal addition that reuses existing compaction infrastructure rather than inventing a new path
- Storm guard concept is sound and prevents the most obvious compaction loop
- `should_auto_compact()` is a clean, testable predicate
- Config option defaults to `0` (disabled), making the feature opt-in with zero risk to existing users
- Good unit test coverage for the predicate and info message format
- Info message includes both current usage and limit, giving the user actionable context
- `reset_for_session_switch` correctly resets the storm guard on session switch

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

https://github.com/Fizzizist/illustrious-manager/pull/190#issuecomment-4406970867

### Classifications

|---|---|---|
| 1. Compaction may not reduce context below threshold → infinite auto-compact loop | **Genuine** | After auto-compact, if context still exceeds threshold, the storm guard clears after one turn and compaction re-triggers every other ResponseComplete indefinitely, burning API tokens with no circuit breaker. |
| 2. Permanent compaction failure → infinite retry loop | **Genuine** | When compaction always fails, the same cycle occurs but with zero user feedback that compaction is retrying. No backoff, no max-retry, no session disable. |
| 3. `last_input_total` is estimated (not real) after compaction | **Nitpick** | After compaction, the storm guard is set, so `should_auto_compact` returns false regardless of the estimated value. By the time the guard clears on the next `ResponseComplete`, a real `Usage` event has already corrected `last_input_total`. Estimation never drives a decision. |
| 4. Auto-compact is TUI-only — stdout frontend silently lacks the feature | **Deferred** | Stdout is single-shot mode — no multi-turn context accumulation, so auto-compact is inapplicable. Config template should clarify it's TUI-only, but the feature gap is by-design. |
| 5. No validation floor on `max_context_window_len` | **Genuine** | Setting `max_context_window_len = 1` causes auto-compact every other turn — a token-burning DoS. No minimum check in `validate()`, no guidance in config template. |
| 6. Auto-compact spawn has no cancellation token | **Genuine** | Fire-and-forget spawn with no `CancellationToken`. HTTP connection from `agent.compact()` can linger after `JoinHandle::abort()`. Auto-compact makes this more likely since it fires automatically (e.g., right before Ctrl-C). |
| 7. `saved_auto_compacted` save/restore pattern is misleading | **Nitpick** | Works correctly but exists because `reset_for_session_switch` is the wrong abstraction for compaction completion. A `reload_history` method would be cleaner, but this is a smell, not a bug. |
| 8. Storm guard `else if` branch is consequence of inverted control flow | **Nitpick** | Correctly works. Suggestion to "reset-first, then possibly set" would simplify reasoning but is stylistic. |
| 9. `should_auto_compact` takes external parameter for single call site | **Nitpick** | Design preference — inline the check or move the config field onto `App`. Not a correctness issue. |
| 10. Core trigger flow in `run_app` is untested | **Genuine** | The main behavioral sequence (ResponseComplete → check → spawn → CompactionComplete → reload) has zero test coverage. Unit tests only cover `should_auto_compact` in isolation. Per project TDD policy, testable code should ship with behavioral tests. |
| 11. Storm guard reset branch (`else if`) is untested | **Genuine** | The `else if app.auto_compacted_last_turn` branch governs a critical invariant but has zero test coverage. A regression here would silently break the storm guard. |
| 12. `CompactionComplete` recovery path is untested for auto-compaction | **Genuine** | `saved_auto_compacted` preservation is untested. Dropping the restore line would cause double-compaction, and no test would catch it. |
| 13. `compaction_task` lifecycle is untested | **Nitpick** | Simple set/clear/abort lifecycle that's trivially correct by inspection. Testing would require mocking `tokio::spawn` or simulating the full loop. Low risk. |

### Action Plan

**Priority 1 — Must fix before merge:**

1. **Add compaction circuit breaker** (Findings 1 & 2): After N consecutive auto-compact attempts (2-3) that fail to bring context below threshold, or after any compaction failure, disable auto-compaction for the session and surface a warning. This prevents infinite compaction/retry loops.

2. **Add validation floor for `max_context_window_len`** (Finding 5): In `validate()`, reject values > 0 and < a sensible minimum (e.g., 1000). Update config template comment with a recommended range.

3. **Add cancellation token to auto-compact spawn** (Finding 6): Create a `CancellationToken` for the compaction task. Pass it into the spawned closure and check before calling `agent.compact()`. Cancel on TUI exit.

**Priority 2 — Should fix before merge (TDD policy):**

4. **Test the storm guard reset branch** (Finding 11): Add test for the `else if` path — when `auto_compacted_last_turn` is true and threshold not exceeded, verify guard is cleared.

5. **Test the `CompactionComplete` save/restore mechanism** (Finding 12): Add test that verifies `auto_compacted_last_turn` is preserved across `reset_for_session_switch` via the save/restore pattern.

6. **Extract and test the auto-compact trigger sequence** (Finding 10): Extract the ResponseComplete → threshold-check → spawn logic into a testable method (e.g., `handle_response_complete_for_auto_compact`) to enable behavioral testing without the full TUI event loop.

**Priority 3 — Can defer to follow-up:**

7. Simplify storm guard control flow (Finding 8): Unconditionally clear `auto_compacted_last_turn` at start of each `ResponseComplete`, set back to `true` if compaction triggers.
8. Replace `saved_auto_compacted` save/restore with dedicated `reload_history` method (Finding 7).
9. Move `max_context_window_len` onto `App` or inline the check (Finding 9).
10. Clarify stdout mode limitation in config template (Finding 4).

---
*Assessment by mach6*

Fizzizist:
## Progress Update

Addressed all priority 1 and priority 2 review findings.

**Priority 1 — Circuit breaker (findings 1 & 2):** Added `consecutive_auto_compact_count: u32` and `auto_compact_disabled: bool` to `App`. After 3 consecutive auto-compact attempts (counter >= 3), auto-compaction is disabled for the session and a `Warn` entry is pushed. On compaction failure (`CompactionComplete` with `is_error`), auto-compaction is immediately disabled with a `Warn` entry. Added `ConversationRole::Warn` variant to `conversation_area.rs`. Both fields reset on session switch.

**Priority 1 — Validation floor (finding 5):** Added check in `validate()`: if `max_context_window_len > 0 && max_context_window_len < 1000`, returns an error. Updated config template comment to show minimum and recommended range.

**Priority 1 — Cancellation token (finding 6):** Added `auto_compact_cancel_token: Option<CancellationToken>` to `App`. Token is created before spawn, stored on `app`, cloned into the spawn closure, and checked before calling `agent.compact()`. Token is cancelled on TUI exit. Cleared on `CompactionComplete`.

**Priority 2 — Tests (findings 10, 11, 12):**
- Finding 11: `storm_guard_clears_when_threshold_no_longer_exceeded` — tests the else-if branch that resets `auto_compacted_last_turn` and `consecutive_auto_compact_count`
- Finding 12: `compaction_complete_preserves_auto_compacted_last_turn_across_reset` — tests the save/restore pattern across `reset_for_session_switch`
- Finding 10: `auto_compact_trigger_sequence_pushes_info_message_and_sets_guard` — tests trigger sequence: predicate → info message → storm guard set → re-trigger blocked
- Finding 10: `auto_compact_circuit_breaker_disables_after_three_attempts` — tests circuit breaker at count >= 3

**Verification:** 823 tests pass. `cargo clippy` and `cargo fmt` clean.

**Commit:** `5a8a2b4`

---
*Progress tracked by mach6*

Fizzizist:
## Progress Update

Reduced the auto-compact circuit breaker from 3 to 2 consecutive attempts. If the first auto-compact doesn't bring context below the threshold, a second consecutive attempt that also fails is sufficient evidence to disable auto-compaction for the session. Three was unnecessarily lenient.

Also updated the test name and logic from `auto_compact_circuit_breaker_disables_after_three_attempts` → `auto_compact_circuit_breaker_disables_after_two_attempts`.

**Commit:** `625d769`

---
*Progress tracked by mach6*

PR #192, opened 2026-05-06
Generalize zai backend into a reusable openai_compat backend

### LINKED ISSUE #191: Generalize zai backend into a reusable openai_compat backend (enables internal vLLM endpoint)
### Summary
The existing `zai` backend (`src/backend/zai.rs`) is, in practice, a complete OpenAI Chat Completions client — it speaks the same wire format that any vLLM / OpenAI-compatible server expects. However, it is hardcoded to z.ai's endpoint and assumes an API key is always present, which prevents reuse for other OpenAI-compatible deployments.

This ticket proposes refactoring `zai` into a generic `openai_compat` backend so we can point it at arbitrary OpenAI-compatible endpoints (z.ai, internal vLLM, OpenRouter, vanilla OpenAI, etc.) via configuration alone, instead of duplicating ~400 lines per provider.

### Motivation
Concrete near-term driver: an internal company vLLM deployment is now available with the following details, and we want to use it without forking the codebase:

- **Endpoint**: `https://vllm.k8s.dc.rxrx.io/v1`
- **API**: OpenAI Chat Completions
- **Model**: `Qwen/Qwen3.6-27B-FP8`
- **Reasoning**: enabled via the qwen-chat-template format
- **Context window**: 256k tokens
- **Max output**: 16,384 tokens
- **Input types**: text + image

`zai.rs` already implements everything needed for the text + tool-use + reasoning portion of this:

- SSE transport with `data: ... [DONE]` framing (`sse.rs`).
- Reads `choices[0].delta.content`, `choices[0].delta.reasoning_content`, `choices[0].delta.tool_calls[].function.{name,arguments}`, `choices[0].finish_reason`, `usage.{prompt_tokens,completion_tokens}`.
- Sends `messages[]` with `role: user|assistant|tool`, `tool_call_id`, and `tools[].function.{name,description,parameters}` plus `parallel_tool_calls`.
- `POST /chat/completions` with `Bearer` auth, `max_tokens`, and `stream: true`.

### What blocks direct reuse today

1. **Endpoint is a `const`** in `zai.rs`:
   ```rust
   const ENDPOINT: &str = "https://api.z.ai/api/coding/paas/v4/chat/completions";
   ```
   `ZaiConfig` has no `base_url` field.
2. **Auth is mandatory**: `ZaiBackend::new` rejects empty API keys. Internal deployments behind network ACLs may not require one.
3. **Reasoning toggle is provider-specific**: zai uses a top-level `enable_thinking: true` body field. vLLM + Qwen typically expects either `chat_template_kwargs: {"enable_thinking": true}` or has thinking on by default. Needs to be configurable.
4. **No image input support** anywhere in the codebase: `ContentBlock` has no `Image` variant; `build_request_body` only emits `{"type":"text", ...}` parts. OpenAI image inputs use `{"type":"image_url","image_url":{"url":"..."}}`. This is cross-cutting (types + every backend + frontend rendering), so it should be tracked as a follow-up, not bundled here.

#### 1. New backend module `src/backend/openai_compat.rs`

Move (not copy) the bulk of `zai.rs` here. Keep the SSE parser (rename to `OpenAiSseParser`) and `build_request_body` essentially intact. Make endpoint/auth/reasoning configurable instead of hardcoded.

```rust
pub struct OpenAiCompatConfig {
    pub base_url: String,            // e.g. "https://vllm.k8s.dc.rxrx.io/v1"
    pub api_key: Option<String>,     // None => no Authorization header
    pub model: String,
    pub max_tokens: u32,
    pub reasoning: ReasoningStyle,
}

pub enum ReasoningStyle {
    None,
    /// z.ai: top-level `enable_thinking: true`
    ZaiEnableThinking,
    /// vLLM + Qwen: `chat_template_kwargs.enable_thinking = true`
    QwenChatTemplate,
    /// Pass-through: no reasoning toggle (server-side default)
    Default,
}
```

Backend constructor:
- Appends `/chat/completions` to `base_url` (handle trailing slash).
- Only sets `Authorization: Bearer <key>` when `api_key` is `Some(non_empty)`.
- Applies the reasoning toggle in `build_request_body` based on `ReasoningStyle`.

#### 2. Config layout

In `config.toml`, deprecate the dedicated `[zai]` section in favor of named OpenAI-compatible providers. Migration: keep `[zai]` working as a thin shim that constructs an `OpenAiCompatConfig` with `ReasoningStyle::ZaiEnableThinking` and the z.ai base URL, so existing user configs do not break.

Example new layout:

```toml
backend = "rxrx-vllm"   # name of the provider entry below

[providers.zai]
kind = "openai_compat"
base_url = "https://api.z.ai/api/coding/paas/v4"
api_key = "..."
model = "glm-5.1"
reasoning = "zai_enable_thinking"

[providers.rxrx-vllm]
kind = "openai_compat"
base_url = "https://vllm.k8s.dc.rxrx.io/v1"
api_key = ""             # optional
model = "Qwen/Qwen3.6-27B-FP8"
max_tokens = 16384
reasoning = "qwen_chat_template"
```

Exact schema is up for discussion — the simpler v1 alternative is to add a second top-level section like `[openai_compat]` with the same fields, and pick it via `backend = "openai_compat"`. Either way the existing `[zai]` keys must continue to function.

#### 3. Delete `src/backend/zai.rs` once the shim is in place

Or shrink it to a thin wrapper around `OpenAiCompatBackend`. The SSE parser tests in `zai.rs` move to `openai_compat.rs` unchanged.

### Out of scope (follow-up tickets)

- **Image input support.** Requires adding `ContentBlock::Image { ... }`, threading it through `vertex`, `ollama`, and `openai_compat` request builders, and deciding how the TUI/stdout frontends accept and display image inputs. File a separate ticket once the backend refactor lands.
- **Auto-detection of `reasoning_style`** from the model string. Explicit config is fine for v1.

### Acceptance Criteria

- [ ] New backend `openai_compat` exists and is selectable via config.
- [ ] Pointing `openai_compat` at the internal vLLM endpoint produces working text streaming, reasoning streaming (`reasoning_content` deltas), and tool use against `Qwen/Qwen3.6-27B-FP8`.
- [ ] Existing user configs with `backend = "zai"` and a `[zai]` section continue to work without edits.
- [ ] No new `.unwrap()` calls; `cargo fmt` and `cargo clippy` clean.
- [ ] SSE parser tests from the old `zai.rs` are preserved (renamed where appropriate) and pass.
- [ ] At least one integration-style test that builds a request body for the vLLM/Qwen profile and asserts the reasoning toggle goes into `chat_template_kwargs`, not the top-level body.
- [ ] `CLAUDE.md` "Architecture" section updated to describe `openai_compat` and how `[zai]` maps onto it.

### Risks / Open Questions

- **Reasoning flag for vLLM + Qwen**: needs confirmation from the platform team whether `chat_template_kwargs.enable_thinking` is the right knob, or whether reasoning is on by default and there is no toggle. If the latter, `ReasoningStyle::Default` covers it and `QwenChatTemplate` becomes optional.
- **`max_tokens` semantics**: vLLM enforces a hard cap (16,384 here). Make sure config values above the server cap surface a clear error rather than a silent truncation, similar to the existing `"length"` finish_reason handling in `ZaiSseParser`.
- **Endpoint normalization**: decide whether `base_url` is the full URL up to but not including `/chat/completions`, or whether we accept either form. Document and validate.
  - **Resolved (rxrx-vllm)**: the platform team confirmed the completion URL is `https://vllm.k8s.dc.rxrx.io/v1` (i.e. the OpenAI-style root, _not_ `/v1/chat/completions`). The backend should append `/chat/completions` to whatever `base_url` is configured, after stripping a trailing slash. Validation should reject a `base_url` that already ends in `/chat/completions` with a clear error pointing at the correct form.

Fizzizist:
Closes #191

Generalize the `zai` backend into a reusable `openai_compat` backend so any OpenAI Chat Completions endpoint (z.ai, internal vLLM, OpenRouter, vanilla OpenAI) can be used via configuration alone. Existing `[zai]` configs continue to work via a thin shim.


Fizzizist:
## Implementation Plan

### Problem

`src/backend/zai.rs` is in practice a complete OpenAI-compatible Chat Completions client (~1000 LOC) but is hardcoded to z.ai's endpoint (`const ENDPOINT`), requires an API key, and uses z.ai's specific top-level `enable_thinking` flag. We can't point it at the internal vLLM endpoint (`https://vllm.k8s.dc.rxrx.io/v1`, `Qwen/Qwen3.6-27B-FP8`) without forking it. This plan refactors it into a generic `openai_compat` backend that subsumes z.ai while keeping existing `[zai]` configs working.

### Design Choice — config schema

V1 uses a top-level `[openai_compat]` section selected via `backend = "openai_compat"`, mirroring the existing `[vertex]` / `[zai]` / `[ollama]` pattern. Smallest diff to `AppConfig`, `validate`, `apply_overrides`, `normalize_back_compat`, `BackendFactory::for_role`. The `[providers.<name>]` map proposed in the issue is a forward-compatible follow-up.

### Deliverables

**1. `src/backend/openai_compat.rs`** — new module containing:
- `OpenAiCompatBackend` struct.
- `OpenAiCompatConfig { base_url, api_key: Option<String>, model, max_tokens: Option<u32>, reasoning: ReasoningStyle }`.
- `ReasoningStyle` enum: `None`, `ZaiEnableThinking`, `QwenChatTemplate`, `Default`.
- `OpenAiSseParser` — moved verbatim from `ZaiSseParser`.
- `build_request_body` — moved verbatim from `zai.rs`, with the reasoning toggle dispatched on `ReasoningStyle`:
  - `ZaiEnableThinking` → top-level `body["enable_thinking"] = true`
  - `QwenChatTemplate` → `body["chat_template_kwargs"]["enable_thinking"] = true`
  - `Default` / `None` → no toggle written
- `max_tokens` precedence: when `OpenAiCompatConfig.max_tokens` is `Some(n)`, the body uses `n`; otherwise falls back to `RequestConfig.max_tokens`. This gives the user control via config while preserving current behavior when omitted.
- Constructor:
  - Strips trailing `/` from `base_url`.
  - Returns `Err` if `base_url` already ends in `/chat/completions` (per issue's resolved open question — point user at the correct form).
  - Appends `/chat/completions` once.
  - Skips `Authorization` header entirely when `api_key` is `None` or empty.

**2. `src/backend/zai.rs`** — collapsed to a ~30-line shim:
- `ZaiBackend::new(api_key)` constructs an `OpenAiCompatBackend` with `base_url = "https://api.z.ai/api/coding/paas/v4"`, `reasoning = ReasoningStyle::ZaiEnableThinking`, `max_tokens = None` (so existing `RequestConfig.max_tokens` path is preserved exactly).
- Existing SSE parser / body builder tests **move** to `openai_compat.rs` (renamed where needed). The `zai.rs` shim keeps a small smoke test ensuring defaults are wired correctly.

**3. `src/config.rs`**:
- New `OpenAiCompatConfigToml { base_url, api_key, model, max_tokens, reasoning }` deserialized from `[openai_compat]`. `reasoning` is a string enum: `"none"`, `"zai_enable_thinking"`, `"qwen_chat_template"`, `"default"`.
- Add `pub openai_compat: Option<OpenAiCompatConfigToml>` field on `AppConfig`.
- `validate` arm for `"openai_compat"`: requires `base_url` non-empty; reasoning string must parse if present.
- `normalize_back_compat`: when `backend = "openai_compat"`, synthesize `default` role from `[openai_compat].model`.
- `apply_overrides`: `--model` flag overrides `[openai_compat].model` and `models["default"]`.
- Update `CONFIG_TEMPLATE` with a commented `[openai_compat]` example pointed at the rxrx-vllm endpoint.
- `[zai]` continues to deserialize and work unchanged.

**4. `src/backend/mod.rs`**:
- `BackendFactory::for_role` arm for `"openai_compat"` constructs `OpenAiCompatBackend` from the `[openai_compat]` table.
- The `"zai"` arm continues to dispatch to `zai::ZaiBackend::new(...)` (which is now the shim).

**5. `CLAUDE.md`** — Architecture section: list `openai_compat` as the fourth backend, describe `ReasoningStyle`, document that `[zai]` is a back-compat shim over it, mention `base_url` normalization rule.

### Acceptance Criteria

- [ ] `backend = "openai_compat"` with the rxrx-vllm `[openai_compat]` block produces working text streaming, reasoning streaming, and tool use end-to-end against `Qwen/Qwen3.6-27B-FP8` (manual verification noted in PR description; no test can hit the live endpoint).
- [ ] Existing user configs with `backend = "zai"` and a `[zai]` section continue to work without edits — verified by an integration test that loads such a config and asserts the resulting backend type and request body shape.
- [ ] All existing SSE parser + body builder tests from `zai.rs` are preserved in `openai_compat.rs` (renamed where appropriate) and pass.
- [ ] `build_request_body` with `ReasoningStyle::QwenChatTemplate` writes `chat_template_kwargs.enable_thinking = true` and does not write top-level `enable_thinking`.
- [ ] `build_request_body` with `ReasoningStyle::ZaiEnableThinking` writes top-level `enable_thinking = true` and does not write `chat_template_kwargs`.
- [ ] When `OpenAiCompatConfig.max_tokens = Some(n)`, the request body's `max_tokens` reflects `n` (not the `RequestConfig.max_tokens` value). When `None`, falls back to `RequestConfig.max_tokens`.
- [ ] Constructor strips trailing `/` on `base_url`.
- [ ] Constructor returns `Err` when `base_url` ends in `/chat/completions`, with a message pointing at the correct form.
- [ ] When `api_key` is `None` or empty, no `Authorization` header is included in outbound requests (verified via a unit-testable `build_headers` helper).
- [ ] `validate` rejects `[openai_compat]` with empty `base_url`.
- [ ] `cargo fmt` + `cargo clippy` clean. No new `.unwrap()` in production paths.
- [ ] CLAUDE.md updated.

### Files touched

| File | Change |
|---|---|
| `src/backend/openai_compat.rs` | new — backend, parser, body builder, tests |
| `src/backend/zai.rs` | shrink to ~30-line shim + 1 smoke test |
| `src/backend/mod.rs` | add `pub mod openai_compat;` + factory arm |
| `src/config.rs` | new struct, validation, overrides, normalize, template |
| `CLAUDE.md` | Architecture section update |

### Testing approach

- Move existing `zai.rs` tests verbatim to `openai_compat.rs`, renaming `ZaiSseParser` → `OpenAiSseParser` and `ZaiBackend::new("k")` → `OpenAiCompatBackend::new(test_config())`.
- Add the new tests listed in Acceptance Criteria, including an explicit `max_tokens` precedence test (config `Some(16384)` overrides `RequestConfig.max_tokens = 1024` in the body).
- Add a config-level test in `src/config.rs` that loads a TOML string with `backend = "openai_compat"` + `[openai_compat]` and asserts `normalize_back_compat` synthesizes the `default` role from `[openai_compat].model`.
- Add a config-level test that loads a legacy `backend = "zai"` config and confirms `validate` + `normalize_back_compat` still pass.
- Extract `build_headers(&self) -> reqwest::header::HeaderMap` so the auth-header-presence test is a pure unit test (no `wiremock` dep).

### Risks / open questions

- **`base_url` validation strictness**: rejecting `/chat/completions` suffix is per the issue's resolved open question. We will not auto-strip — explicit error pointing at correct form.
- **Image input** is explicitly out of scope per the issue (separate ticket).
- **`max_tokens` semantics for vLLM**: vLLM enforces a hard server cap (16,384 for the rxrx deployment). We rely on the existing `"length"` finish_reason handling in the SSE parser to surface a clear error if the user configures a value above the cap. No additional client-side validation in this PR.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implementation complete. All plan deliverables implemented in a single commit.

**Changes:**

- **`src/backend/openai_compat.rs`** (new) — `OpenAiCompatBackend`, `OpenAiCompatSseParser` (moved from `zai.rs`), `ReasoningStyle` enum (`None | ZaiEnableThinking | QwenChatTemplate | Default`), `build_headers` (auth header omitted when `api_key` is `None`/empty), `build_request_body` with per-style reasoning dispatch and `max_tokens` config override. Constructor strips trailing `/`, rejects `/chat/completions` suffix. Full test suite covering all acceptance criteria.

- **`src/backend/zai.rs`** — collapsed to ~35-line shim over `OpenAiCompatBackend` (`base_url = "https://api.z.ai/api/coding/paas/v4"`, `ReasoningStyle::ZaiEnableThinking`). Re-exports `OpenAiCompatSseParser as ZaiSseParser` for back-compat with `tests/zai_test.rs`.

- **`src/backend/mod.rs`** — `pub mod openai_compat` registered; `BackendFactory::for_role` arm for `"openai_compat"` maps `ReasoningStyleConfig` → `ReasoningStyle` and constructs the backend.

- **`src/config.rs`** — `ReasoningStyleConfig` enum, `OpenAiCompatConfigToml` struct, `openai_compat: Option<OpenAiCompatConfigToml>` field on `AppConfig`, `validate`/`normalize_back_compat`/`apply_overrides` arms, commented `[openai_compat]` block in `CONFIG_TEMPLATE` pointed at rxrx-vllm. `openai_compat: None` added to all `AppConfig` struct literals across the codebase.

- **`CLAUDE.md`** — architecture section updated: 4th backend listed, `ReasoningStyle` described, zai-as-shim noted, URL normalization and auth-header rules documented.

**Test results:** 793 passed, 0 failed across all test suites. `cargo clippy` clean. `cargo fmt` clean.

**Commit:** `ced66bd`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

---

### Critical

*(none)*

---

### Important

**Finding 1: Parallel tool calls silently discard all but the last tool**
**File:** `src/backend/openai_compat.rs` — `fill_buffer` / `send_message`
**Severity:** high | **Confidence:** 95

The backend sets `"parallel_tool_calls": true`, which tells OpenAI-compatible APIs to stream multiple tool calls in a single response. But `fill_buffer` emits exactly **one** `ToolUseDone` event — on `finish_reason: "tool_calls"` — for the entire response, regardless of how many tools were called.

The agent state machine (`src/agent/mod.rs`) tracks a single `current_tool: Option<PendingToolCall>`. Each `ToolUseStart` overwrites the previous value. When two parallel tools arrive — tool 0 (name + deltas), tool 1 (name + deltas), then the single `ToolUseDone` — only tool 1 is committed. Tool 0's accumulated `input_json` is silently discarded.

Fix: emit a synthetic `ToolUseDone` before each new `ToolUseStart` when a tool is already in progress, mirroring Vertex's per-block semantics.

---

**Finding 2: `Thinking` and `RedactedThinking` blocks serialise as `null` in assistant content arrays**
**File:** `src/backend/openai_compat.rs` — `build_request_body`, lines ~307-323
**Severity:** high | **Confidence:** 95

In the assistant non-tool-use serialisation path:
```rust
ContentBlock::Thinking { .. }
| ContentBlock::RedactedThinking { .. } => serde_json::Value::Null,
```
These `null` values are collected and sent verbatim. A resumed session with prior thinking blocks sends `[null, {"type":"text","text":"..."}]` as assistant content, which most APIs will reject or silently misinterpret. Fix: filter out `Null` values before pushing the message, or skip these blocks without mapping to `Null`.

---

**Finding 3: `is_error` on `ToolResult` is silently dropped — tool errors invisible to the model**
**File:** `src/backend/openai_compat.rs` — `build_request_body`, lines ~233-241
**Severity:** medium | **Confidence:** 98

`ToolResult` is destructured with `..` and `is_error` is never included in the serialised `role: "tool"` message. The model receives a successful-looking response regardless of whether the tool errored. The model may proceed confidently on error output, producing incorrect follow-up actions. OpenAI's spec does not have a standard `is_error` field for `role: "tool"`, but the error content should at minimum be prefixed or formatted to signal failure to the model.

---

**Finding 4: `ToolUseDelta` emitted without guaranteed preceding `ToolUseStart`**
**File:** `src/backend/openai_compat.rs` — `fill_buffer`, lines ~133-155
**Severity:** medium | **Confidence:** 92

`tool_calls_by_index` is inserted into when a `name` is found, but is never queried before emitting `ToolUseDelta`. If a provider sends an arguments-only chunk before the name chunk (out-of-order or batched), the agent receives `ToolUseDelta` with no preceding `ToolUseStart`. The agent's `current_tool.push_str` call on `None` silently discards the delta. Fix: guard `ToolUseDelta` emission with `self.tool_calls_by_index.contains_key(&index)`.

---

**Finding 5: `reasoning_content` early-return silently drops co-located `content`**
**File:** `src/backend/openai_compat.rs` — `fill_buffer`, lines ~117-131
**Severity:** medium | **Confidence:** 85

```rust
if let Some(reasoning) = json["choices"][0]["delta"]["reasoning_content"].as_str()
    && !reasoning.is_empty()
{
    self.event_buffer.push(StreamEvent::ThinkingDelta(...));
    return Ok(());  // ← exits before checking `content`
}
```
If a provider sends both `reasoning_content` and `content` in the same delta chunk, the text delta is silently dropped. Fix: remove the early return; push thinking delta and fall through to the content check.

---

### Suggestions

**Finding 6: Usage silently lost when provider sends a standalone usage-only chunk**
**File:** `src/backend/openai_compat.rs` — `fill_buffer`, lines ~79-115
**Severity:** low | **Confidence:** 88

Usage is only read when `json["choices"][0]["finish_reason"]` is present. vLLM with `stream_options: {include_usage: true}` sends a final SSE chunk with an empty `choices` array and a top-level `usage` object — `json["choices"][0]` evaluates to `Null`, so the usage block is bypassed and tokens are silently zeroed.

---

**Finding 7: `finish_reason="stop"` with co-located usage emits `Usage` before the final `TextDelta`**
**File:** `src/backend/openai_compat.rs` — `fill_buffer`
**Severity:** low | **Confidence:** 82

When a provider sends `finish_reason` and the last content token in the same chunk, `Usage` is pushed before the `TextDelta` check (no early return in the `_` arm). The agent emits usage statistics before the last character appears in the TUI.

---

**Finding 8: `tool_calls_by_index` HashMap written but never read**
**File:** `src/backend/openai_compat.rs` — `OpenAiCompatSseParser`
**Severity:** low | **Confidence:** 98

`tool_calls_by_index: HashMap<u64, String>` is inserted into on `ToolUseStart` and cleared on done, but is never queried. The ID is constructed inline (`format!("tool_{}", index)`) and deltas don't route by index. Dead state — should be removed.

---

**Finding 9: `OpenAiCompatConfig::model` field is silently discarded**
**File:** `src/backend/openai_compat.rs` — `OpenAiCompatConfig`, `OpenAiCompatBackend::new`
**Severity:** low | **Confidence:** 96

`OpenAiCompatConfig` has a `model` field that the constructor reads and drops. The actual model comes from `RequestConfig::model`. Either remove the field from `OpenAiCompatConfig` (it's already threaded through `RequestConfig`) or update the doc comment to be unambiguous.

---

**Finding 10: Missing tests for SSE `finish_reason="stop"` usage emission and `tool_calls` + usage co-emission**
**File:** `src/backend/openai_compat.rs` — test module
**Severity:** medium | **Confidence:** 95

Two uncovered parser paths:
- `finish_reason="stop"` with `usage` object: the `_` arm's `Usage` event is never exercised.
- `finish_reason="tool_calls"` with `usage`: both `ToolUseDone` and `Usage` are emitted, but only `ToolUseDone` is verified in the existing test.

---

**Finding 11: `create_sse_event_stream` has zero async integration tests**
**File:** `src/backend/sse.rs`
**Severity:** high | **Confidence:** 98

The core streaming function (used by both `openai_compat` and `vertex`) has no tests for: SSE frames split across byte chunks, multiple frames per chunk, trailing data without `\n\n`, or byte-stream error propagation. Only the trivial `extract_sse_data` helper is tested.

---

**Finding 12: `QwenChatTemplate` thinking-disabled path is untested; mixed text+tool-use assistant message untested**
**File:** `src/backend/openai_compat.rs` — tests
**Severity:** low | **Confidence:** 90

- No test for `QwenChatTemplate` when `config.thinking` is `None` (symmetric to the existing `ZaiEnableThinking` absent test).
- No test for an assistant message with both `Text` and `ToolUse` blocks — the `text_parts` accumulation path is untested.

---

**Finding 13: Missing `apply_overrides` and role-validation tests for `openai_compat`**
**File:** `src/config.rs` — test module
**Severity:** low | **Confidence:** 88

`apply_overrides` and the per-role backend validation loop have tests for `vertex`, `zai`, and `ollama` but not for `openai_compat`. The `--model` CLI override and the role-misconfiguration error branches are uncovered.

---

**Finding 14: Two missing integration tests required by plan acceptance criteria**
**File:** `src/config.rs`, `tests/`
**Severity:** low | **Confidence:** 88

The plan required:
1. A `load_config_from_path`-based test for a legacy `backend = "zai"` config confirming `validate` + `normalize_back_compat` still pass and synthesize `models["default"]`. (Struct-level tests exist but the end-to-end load is not covered.)
2. An integration test constructing `ZaiBackend` from a TOML-loaded config via `BackendFactory` and asserting request body shape. (Only the endpoint URL is checked in the existing smoke test.)

---

**Finding 15: `default_openai_compat_model` is an unnecessary named default; redundant no-op match arms; `parse()` should be `#[cfg(test)]`**
**File:** `src/config.rs`, `src/backend/openai_compat.rs`
**Severity:** low | **Confidence:** 90+

- `default_openai_compat_model()` returns `String::new()` — identical to `#[serde(default)]`. The named function can be deleted.
- In the assistant tool-use branch, `Thinking | RedactedThinking => {}` is immediately followed by `_ => {}` — the first arm is dead.
- `parse()` is never called in production; it exists only for tests. Should be `#[cfg(test)]` to keep the public API clean.

---

**Finding 16: `ReasoningStyleConfig` → `ReasoningStyle` is a 4-arm identity mapping duplicated at each call site**
**File:** `src/backend/mod.rs`, lines 97-108
**Severity:** low | **Confidence:** 80

The two enums are structurally identical. A `From<ReasoningStyleConfig> for ReasoningStyle` impl in `openai_compat.rs` replaces the inline match and makes future call sites trivial.

---

### Strengths

- The overall refactor is architecturally sound: moving a 963-LOC provider-specific client into a generic backend with a clean config surface is the right call.
- `ReasoningStyle` enum cleanly captures provider-specific divergences without leaking them into shared code.
- Constructor validation (trailing slash, `/chat/completions` suffix rejection, empty-key header omission) is correct and well-tested.
- `build_headers` extracted as a public method enables pure unit tests without network stubs — good design.
- `BackendFactory::for_role` arm is clean and uses `ok_or_else` correctly.
- The `ZaiBackend` shim is minimal and correct; the `ZaiSseParser` re-export alias is a thoughtful back-compat touch.
- `cargo clippy` and `cargo fmt` are clean.

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

Review comment: https://github.com/Fizzizist/illustrious-manager/pull/192#issuecomment-4391032083

### Classifications

|---|---|---|
| finding 1: Parallel tool calls discard all but last tool | **genuine** | Confirmed. `fill_buffer` emits one `ToolUseDone` for the entire `finish_reason:"tool_calls"` chunk. Agent's `current_tool` is a single `Option<PendingToolCall>` — each `ToolUseStart` overwrites the previous. For N parallel tools, only tool N survives. Silent data loss in production when `parallel_tool_calls: true`. |
| finding 2: Thinking/RedactedThinking blocks serialise as `null` | **genuine** | Confirmed. Lines 314-315 map both block types to `serde_json::Value::Null`, which land in the sent content array. Resumed thinking sessions will send `[null, {"type":"text",...}]` — most APIs reject or mishandle null content items. |
| finding 3: `is_error` on ToolResult silently dropped | **false positive** | OpenAI Chat Completions `role: "tool"` messages have no `is_error` field. Dropping it is protocol-correct. Error information reaches the model through the content string. |
| finding 4: ToolUseDelta without guaranteed ToolUseStart | **false positive** | OpenAI streaming guarantees `name` in the first chunk per index and streams deltas in index order. Out-of-order delivery is a hypothetical that doesn't match the protocol. |
| finding 5: `reasoning_content` early-return drops co-located `content` | **genuine** | Confirmed. `return Ok(())` after pushing `ThinkingDelta` prevents the `content` field from being examined in the same delta chunk. Structurally wrong even if no current provider triggers it. |
| finding 6: Usage lost on standalone usage-only chunk | **deferred** | Real gap, but the backend never sends `stream_options: {include_usage: true}`, so no provider will respond with a standalone usage chunk. Only relevant when that request-side feature is added. |
| finding 7: `Usage` event emitted before final `TextDelta` | **genuine** | Confirmed. `finish_reason` block runs before content check; for co-located `finish_reason + delta.content`, `Usage` is buffered before `TextDelta`. Incorrect ordering — stats appear before last character in TUI. |
| finding 8: `tool_calls_by_index` HashMap never read | **genuine** | Confirmed. Inserted on `ToolUseStart`, cleared on done, never queried. Pure dead state. |
| finding 9: `OpenAiCompatConfig::model` field silently discarded | **nitpick** | Documented in the field's doc comment as intentional. Misleading API symmetry but not a bug. |
| finding 10: Missing parser tests for `stop` usage + `tool_calls` + usage | **genuine** | Confirmed. Both the `_` arm (finish_reason="stop" with usage) and `tool_calls` + co-located usage paths are untested. New code in this PR. |
| finding 11: `create_sse_event_stream` has zero async tests | **deferred** | Pre-existing gap — `sse.rs` predates this PR. Not introduced here. |
| finding 12: `QwenChatTemplate` disabled-thinking untested; mixed text+tool-use untested | **genuine** | Confirmed. No symmetric test to `zai_enable_thinking_absent_when_thinking_disabled` for `QwenChatTemplate`. No test for assistant `Text + ToolUse` mixed message. Both new code paths in this PR. |
| finding 13: `apply_overrides` + role-validation untested for `openai_compat` | **genuine** | Confirmed. Analogous tests exist for vertex, zai, ollama — not for the new `openai_compat` branch. New code in this PR. |
| finding 14: Two plan acceptance-criteria integration tests missing | **deferred** | Pre-existing zai path is already exercised via `tests/zai_test.rs` and smoke tests. Gaps overlap with findings 10, 12, 13 which are marked genuine. |
| finding 15: `default_openai_compat_model` unnecessary; redundant arms; `parse()` visibility | **nitpick** | `default_openai_compat_model` and redundant match arms are minor. `parse()` being public is **not** a false positive nor a problem — it is used by `tests/zai_test.rs` as an external crate (needs public symbol). |
| finding 16: `ReasoningStyleConfig → ReasoningStyle` 4-arm identity mapping | **nitpick** | Real duplication. A `From` impl would be cleaner. No correctness impact. |

### Action Plan

**Priority 1 — Fix before merge**

1. **Fix parallel tool call ToolUseDone semantics (finding 1).** Emit one `ToolUseDone` per tool, not one per `finish_reason` chunk. Emit a synthetic `ToolUseDone` before each new `ToolUseStart` when an index change is detected, finalising the previous tool. Add a parser test simulating a real two-tool parallel streaming sequence.

2. **Fix null thinking blocks in assistant content (finding 2).** Skip `ContentBlock::Thinking` and `ContentBlock::RedactedThinking` in the non-tool-use assistant branch instead of mapping to `Null`. Add a test asserting the content array contains no null values.

3. **Fix early-return dropping co-located `content` after `reasoning_content` (finding 5).** Remove the `return Ok(())` after pushing `ThinkingDelta` and fall through to the content check. Add a parser test for a dual-field delta chunk.

4. **Fix `Usage`/`TextDelta` ordering inversion (finding 7).** Move the `Usage` push to occur after the content delta check, or restructure `fill_buffer` so finish_reason emission happens last.

**Priority 2 — Dead code cleanup**

5. **Remove `tool_calls_by_index` HashMap (finding 8).** Delete the field, the `insert`, and the `clear`. The ID is generated directly at use; the map is never queried.

**Priority 3 — Test coverage for new code**

6. **Add missing parser tests (finding 10).** Cover `finish_reason:"stop"` with usage → `Usage` event; `finish_reason:"tool_calls"` with usage → both `ToolUseDone` and `Usage`.

7. **Add `QwenChatTemplate` disabled-thinking test and mixed text+tool-use test (finding 12).**

8. **Add `apply_overrides` and role-validation tests for `openai_compat` (finding 13).**

**Deferred (create tracking issues separately):**
- Finding 6: standalone usage chunk support (needs `stream_options` on request side)
- Finding 11: `sse.rs` async integration tests (pre-existing gap)

---
*Assessment by mach6*

Fizzizist:
## Code Review

### Critical

**1. Parallel tool calls silently discard all but the last tool** (code-reviewer, 90)
`src/backend/openai_compat.rs` — `OpenAiCompatSseParser::fill_buffer`
The parser emits a single `ToolUseDone` for the whole `finish_reason: "tool_calls"` chunk. Because the agent layer's `current_tool` is overwritten on each `ToolUseStart`, only the last tool in a parallel-tool-call response actually executes; preceding tools are silently dropped. This breaks parallel tool calls — a feature the request body explicitly enables (`parallel_tool_calls: true`). One `ToolUseDone` should be emitted per tool index, not one per terminal SSE chunk.

**2. `Thinking` / `RedactedThinking` blocks serialise as `null` in assistant content arrays** (code-reviewer 90, error-auditor 90, simplifier 82 — three independent flags)
`src/backend/openai_compat.rs` — non-tool-use assistant arm in `build_request_body`
```rust
ContentBlock::Thinking { .. } | ContentBlock::RedactedThinking { .. } => serde_json::Value::Null,
other => serde_json::to_value(other).unwrap_or(serde_json::Value::Null),
```
This sends literal `null` items inside the `content` array, which OpenAI-compatible servers reject or mishandle. Resumed conversations that contain prior thinking blocks will fail or get corrupted output. The fix is `filter_map` + drop the variants entirely (thinking is Anthropic-only; OpenAI-compat endpoints don't model it). The `unwrap_or(Null)` fallback for `other` is the same pattern in the user-message arm and should also become `.ok()` filter-map.

### Important

**3. `reasoning_content` early-return drops co-located `content` in the same SSE delta** (code-reviewer, 85)
`src/backend/openai_compat.rs` — `fill_buffer`
After emitting a `ThinkingDelta`, the function `return Ok(())`s, so a delta that carries both `reasoning_content` and `content` in the same chunk loses the text portion. Some providers do co-emit both fields; ordering should be: parse reasoning, parse content, parse tool_calls — no early return.

**4. `Usage` event emitted before final `TextDelta` for co-located stop+content chunks** (code-reviewer, 80)
`src/backend/openai_compat.rs` — `fill_buffer`
The `finish_reason` block runs before the content check, inverting the order downstream consumers expect (text first, then usage/done). Move `Usage` emission to after content/tool checks.

**5. SSE frames containing only `usage` (no `choices`) are silently dropped** (error-auditor, 88)
`src/backend/openai_compat.rs` — `fill_buffer`
vLLM with `stream_options: {include_usage: true}` emits a final frame that is just `{"usage": {...}}` with no `choices`. Every branch in `fill_buffer` is gated on `json["choices"][0]`, so such frames are consumed and produce no event. Token usage is silently lost. This PR doesn't currently set `stream_options`, so it's latent — but mitigates the value of the parser when later wired up.

**6. `validate()` for `[openai_compat]` doesn't reject `base_url` ending in `/chat/completions`** (error-auditor, 87)
`src/config.rs` — `validate`
`OpenAiCompatBackend::new` rejects this URL form, but `validate` only checks for emptiness. A misconfigured user gets a runtime error on first request rather than a startup validation failure. Mirror the constructor's check at validation time.

**7. `base_url` containing query string or fragment produces malformed endpoint** (error-auditor, 82)
`src/backend/openai_compat.rs` — `OpenAiCompatBackend::new`
A `base_url` like `https://example.com/v1?token=…` becomes `https://example.com/v1?token=…/chat/completions` — silently malformed. The suffix-only check is partial. Either parse with `Url::parse` or document/validate that `base_url` must not contain a query string.

**8. `tool_calls_by_index` HashMap is dead state** (code-reviewer 95, simplifier 97 — two independent flags)
`src/backend/openai_compat.rs` — `OpenAiCompatSseParser`
The map is inserted into and cleared but never read. The doc-comment claims it provides "stable IDs across multiple SSE chunks" but the lookup never happens — IDs are generated in-place from the index. Remove the field and the `HashMap` import (if otherwise unused).

### Suggestions

**9. Missing tests for new behaviors** (test-reviewer, 82–88; code-reviewer 80–85)
- `QwenChatTemplate` with `thinking: None` / `enabled: false` — no test asserts `chat_template_kwargs` is **absent**. Symmetric `ZaiEnableThinking` test exists.
- `finish_reason: "tool_calls"` + non-zero `usage` — both `ToolUseDone` and `Usage` are buffered, but the test that exercises the `tool_calls` path uses zero usage, so the buffered `Usage` drain path is never tested.
- Role-validation: `[models.foo]` with `backend = "openai_compat"` and an `[openai_compat]` section that has empty `base_url` — untested.
- `apply_overrides` for `backend = "openai_compat"` — vertex/zai/ollama all have an apply-overrides test, openai_compat has none.
- Mixed text+tool-use assistant serialization — untested.

**10. `parse()` is a test-only public method** (simplifier, 88)
`src/backend/openai_compat.rs`
`send_message` calls `fill_buffer` + `mem::take` directly. `parse` exists only as a test convenience but is `pub` on a `pub` struct, leaking into the API surface. Mark `#[cfg(test)]` or make private.

**11. `ReasoningStyleConfig` ↔ `ReasoningStyle` 4-arm identity map** (simplifier 85)
Files: `src/backend/mod.rs`, `src/config.rs`
Implement `From<ReasoningStyleConfig> for ReasoningStyle` and replace the manual match in the factory.

**12. `ZaiBackend` newtype is a one-line delegate** (simplifier, 80)
`src/backend/zai.rs`
The wrapper struct delegates `send_message` to `self.0.send_message`. The factory could construct `OpenAiCompatBackend` directly with the z.ai constants, eliminating the file. The `ZaiSseParser` re-export can stay as a type alias if external tests need it. (Counter-consideration: keeping the shim makes the back-compat contract visible at a glance — defer if the team values that.)

**13. `generate_intro_message` swallows config-serialisation failures via `unwrap_or`** (error-auditor, 80)
`src/config.rs`
`toml::Value::try_from(config).unwrap_or(toml::Value::String("(error serializing config)"))` silently substitutes a stub for a real error.

### Strengths

- Clean separation: `openai_compat` is the engine, `zai` is config. Future providers (OpenRouter, vLLM, vanilla OpenAI) drop in trivially.
- Constructor explicitly rejects `/chat/completions` suffix with a directive error message (per resolved issue question).
- `build_headers` extracted as a pure helper — auth-header presence/absence is a pure unit test, no `wiremock` dep.
- `max_tokens` precedence (config `Some(n)` overrides `RequestConfig.max_tokens`) is implemented and explicitly tested.
- `[zai]` back-compat preserved end-to-end: existing user configs and `tests/zai_test.rs` continue to pass against the shim.
- All new tests behavior-driven, no test-induced design pollution.
- CLAUDE.md updated with `ReasoningStyle`, normalization rules, and the zai-as-shim note.
- Per completeness check: every issue #191 acceptance criterion is met (with the live-vLLM check correctly noted as manual).

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

Review: https://github.com/Fizzizist/illustrious-manager/pull/192#issuecomment-4391103420

Each finding was independently re-verified against the code. Two findings the original review flagged as critical/important turned out to be false positives.

### Classifications

|---|---|---|
| 1 — Parallel tool calls silently discarded | **false positive** | The agent layer does NOT overwrite `current_tool` and lose prior tools. `current_tool.take()` on each `ToolUseDone` accumulates calls into `tool_calls: Vec<PendingToolCall>`. Verified by the existing test `multiple_tool_calls_in_single_response_all_executed`. The reviewer confused the SSE protocol (sequential per-tool stream) with a hypothetical parallel-in-one-chunk shape. |
| 2 — `Thinking`/`RedactedThinking` serialise as `null` | **genuine** | Confirmed in the non-tool-use `Role::Assistant` arm of `build_request_body`. Replaying a session that contains thinking blocks will send literal `[..., null, ...]` content arrays. Tool-use arm correctly skips them; non-tool-use arm should match. |
| 3 — `reasoning_content` early-return drops co-located `content` | **genuine** | Confirmed: `return Ok(())` after `ThinkingDelta` push prevents the `content` check from running on the same chunk. Some providers co-emit. |
| 4 — `Usage` emitted before final `TextDelta` | **genuine** | Confirmed: `finish_reason` block runs first, pushes `Usage`, then content delta is pushed after. Drains in wrong order. Latent for providers that send `usage` in a separate frame; real for those that co-locate. |
| 5 — `usage`-only SSE frames silently dropped | **genuine** | Confirmed: every branch in `fill_buffer` is gated on `choices[0]`. Latent today (PR doesn't set `stream_options`), but a real correctness gap once `include_usage` is wired up. |
| 6 — `validate()` doesn't reject `/chat/completions` suffix | **genuine** | Confirmed: validate only checks emptiness; constructor catches it later at runtime. Should fail at startup. |
| 7 — query-string in `base_url` produces malformed endpoint | **genuine** | Confirmed: naive `format!` concatenation. Low real-world likelihood (vLLM/Qwen don't use query auth) but trivial to guard. |
| 8 — `tool_calls_by_index` dead state | **genuine** | Confirmed by grep — written and cleared, never read. Pure dead code. |
| 9a — `QwenChatTemplate` thinking-disabled negative test | **false positive** | Marginal coverage gap, but `reasoning_none_never_adds_enable_thinking` and `zai_enable_thinking_absent_when_thinking_disabled` adequately cover the contract. |
| 9b — `tool_calls + non-zero usage` drain test | **genuine** | Untested; given finding 4 (ordering), this is exactly the test that would have caught the bug. Must ship with this PR. |
| 9c — `openai_compat` empty-`base_url` validation test | **false positive** | `validate_openai_compat_backend_with_empty_base_url_errors` already exists. |
| 9d — `apply_overrides` test for `openai_compat` | **genuine** | All other backends have one; this one doesn't. Must ship with this PR. |
| 9e — mixed text+tool-use assistant serialization | **false positive** | Marginal; the tool-use arm is independently tested and the mixed shape is rare in practice. |
| 10 — `parse()` is a test-only public method | **nitpick** | Style; correct and clear. |
| 11 — `ReasoningStyleConfig` ↔ `ReasoningStyle` 4-arm map | **nitpick** | A `From` impl would be marginally cleaner; no correctness impact. |
| 12 — `ZaiBackend` newtype is a one-line delegate | **nitpick** | The shim is justified — named type, validated constructor, back-compat alias export, and its own tests make the contract visible. Inlining would lose clarity. |
| 13 — `generate_intro_message` swallows config-serialisation failures | **false positive** | This is `unwrap_or` (graceful fallback), not `.unwrap()`. Compliant with policy. |

### Action Plan

Fix before merge — genuine issues, ordered by severity:

1. **Finding 2** — Skip `Thinking`/`RedactedThinking` in non-tool-use assistant content (mirror tool-use arm); do not emit `null`.
2. **Finding 3** — Remove the `return Ok(())` after `ThinkingDelta` push so `content` in the same delta is also processed.
3. **Finding 4** — Reorder `fill_buffer` so `Usage` is pushed after content/tool deltas in the same chunk.
4. **Finding 5** — Add a top-level branch that emits `Usage` for chunks containing `usage` but no `choices`.
5. **Finding 6** — In `validate()` for `[openai_compat]`, mirror the constructor's `/chat/completions` suffix rejection.
6. **Finding 7** — Reject `base_url` containing `?` in `validate()` and/or constructor with a clear error.
7. **Finding 8** — Delete the `tool_calls_by_index` field, its insert, and its clear.
8. **Finding 9b** — Add a parser test: `tool_calls` finish_reason + non-zero `usage` drains `ToolUseDone` then `Usage` (in that order).
9. **Finding 9d** — Add an `apply_overrides` test for `backend = "openai_compat"`, mirroring the vertex/zai/ollama versions.

Deferred (track or do in a follow-up cleanup PR — not blocking):
- Finding 10 (`parse()` visibility): mark `#[cfg(test)]`.
- Finding 11 (`From` impl): mechanical refactor.
- Findings 9a, 9e: marginal coverage gaps.

No tracking issues need to be created — the deferred items are all small style/coverage cleanups that can ride along the next backend touch.

---
*Assessment by mach6*

Fizzizist:
## Progress Update — Review findings addressed

All genuine findings from the assessment fixed in commit `36cd7db`.

**Finding 2** — `Thinking`/`RedactedThinking` blocks in the non-tool-use assistant content arm now use `filter_map` to skip them entirely; no more `null` items in content arrays.

**Finding 3** — Removed the `return Ok(())` after pushing `ThinkingDelta`; a co-located `content` field in the same SSE chunk is now also emitted.

**Finding 4** — Restructured `fill_buffer`: content/tool deltas are always buffered before `Usage`, so the TUI receives the last character before stats appear.

**Finding 5** — `fill_buffer` now emits `Usage` for standalone usage-only SSE frames (missing or empty `choices` array), covering `stream_options: {include_usage: true}`.

**Finding 6** — `validate()` for `[openai_compat]` now rejects `base_url` ending in `/chat/completions` and `base_url` containing a query string.

**Finding 7** — `OpenAiCompatBackend::new` constructor also rejects `base_url` containing `?`.

**Finding 8** — `tool_calls_by_index` `HashMap` field removed; dead state gone, `HashMap` import dropped.

**Finding 9b** — Added parser tests: `tool_calls` finish with non-zero usage drains `ToolUseDone` then `Usage`; `stop` finish co-located with content drains `TextDelta` then `Usage`.

**Finding 9d** — Added `apply_overrides` test for `openai_compat` backend, mirroring vertex/zai/ollama coverage.

**Additional** — Tests for co-located `reasoning_content`+`content`, standalone usage frame, empty-`choices` usage frame, query-string rejection in constructor, and `Thinking`-block null regression.

803 tests passing, 0 failed. `cargo clippy` clean. `cargo fmt` clean.

**Commit:** `36cd7db`

---
*Progress tracked by mach6*

PR #201, opened 2026-05-09
feat: auto-compact session context when usage exceeds threshold

### LINKED ISSUE #183: auto-compaction

Fizzizist:
Closes #183

Adds `max_context_window_len` config option to automatically compact conversation history when the API's reported `input_tokens` exceed the configured threshold.

## Summary

- New config option: `max_context_window_len` (default: `0` = disabled) under `[tools]`
- Auto-compact trigger lives in the agent core (`Agent::send()`), not the TUI — the agent decides *when* to compact, frontends decide *how* to display it
- After each complete assistant turn, if `input_tokens` from `Usage` events exceeds the threshold, `Agent::compact()` is called automatically
- Consecutive-compaction guard: if the previous turn already auto-compacted and the threshold is still exceeded, a warning is emitted instead of re-compacting
- TUI renders `AutoCompactTriggered` as an info conversation entry with usage stats
- Single-shot and sub-agents are single-turn contexts — auto-compaction is a multi-turn concern and won't trigger there


Fizzizist:
## Implementation Plan

### Problem Analysis

Issue 183 requests auto-compaction: when conversation context exceeds a configurable `max_context_window_len` threshold, the system should automatically trigger compaction using the existing `/compact` infrastructure. The trigger must happen after a complete assistant turn, emit a clear info message, and work in the REPL context. A consecutive-compaction guard must prevent infinite loops.

### Design Decisions

- **Agent-core trigger**: The threshold check and auto-compact trigger live in `Agent::send()`, not in any frontend. The agent decides *when* to compact; frontends decide *how* to display it.
- **Single-shot and sub-agents excluded**: These are single-turn contexts. Auto-compaction is inherently multi-turn, so it only matters for the REPL. No special code needed in `stdout.rs` or `run_headless` — they silently absorb `AutoCompactTriggered`.
- **Trigger timing**: After `ResponseComplete` (when `tool_calls.is_empty()`), never mid-tool-loop.
- **Token measurement**: Absolute `input_tokens` from the API's `Usage` event, accumulated within the spawned task in `Agent::send()`.
- **Consecutive compaction guard**: Track whether the previous turn auto-compacted. If so, skip compaction and emit a warning instead of looping.
- **Config default**: `0` means disabled (no auto-compaction).

### Deliverables

1. `max_context_window_len: u32` config field in `ToolsConfig` (default: `0`)
2. Config template update with documented `[tools]` entry
3. `AutoCompactTriggered { current_tokens: u32, threshold: u32 }` variant in `AgentEvent`
4. Auto-compact logic in `Agent::send()` — accumulate `input_tokens` from `StreamEvent::Usage`, check threshold after `ResponseComplete`, call `self.compact()` if exceeded, emit events
5. Consecutive-compaction guard using `Arc<AtomicBool>` on `Agent`
6. TUI handling of `AutoCompactTriggered` in `handle_agent_event()` — push info `ConversationEntry`
7. Stdout and `run_headless` silently absorb the event
8. Tests for config, threshold logic, consecutive guard, event emission, and TUI rendering

### Files to Modify

| File | Change |
|------|--------|
| `src/config.rs` | Add `max_context_window_len: u32` to `ToolsConfig`; add `default_max_context_window_len()` returning `0`; update `ToolsConfig::default()`; update `CONFIG_TEMPLATE`; update all test fixtures that construct `AppConfig` |
| `src/types.rs` | Add `AutoCompactTriggered { current_tokens: u32, threshold: u32 }` variant to `AgentEvent` |
| `src/agent/mod.rs` | Add `max_context_window_len: u32` and `last_auto_compacted: Arc<AtomicBool>` fields to `Agent`; set `max_context_window_len` in `with_tool_config()`; in `send()`'s spawned task: accumulate `input_tokens` from `StreamEvent::Usage`, after `ResponseComplete` check threshold and trigger auto-compact if needed; reset `last_auto_compacted` at start of `send()` |
| `src/agent/compact.rs` | No changes — existing `compact()` method is sufficient |
| `src/frontend/tui/tui_app.rs` | Handle `AgentEvent::AutoCompactTriggered` in `handle_agent_event()` — push `ConversationEntry::Info` with usage/threshold message |
| `src/frontend/stdout.rs` | Silently absorb `AutoCompactTriggered` (single-shot won't trigger it) |
| `src/agent/spawner.rs` | Silently absorb `AutoCompactTriggered` in `run_headless()` |

### Testing Approach

| Test file | What to test |
|-----------|-------------|
| `src/config.rs` | `max_context_window_len` defaults to 0; deserializes from TOML; round-trips through serialize/deserialize |
| `src/agent/mod.rs` | Auto-compact triggers when `input_tokens > threshold`; does not trigger when `threshold == 0`; does not trigger when `input_tokens <= threshold`; consecutive guard prevents re-compaction; `AutoCompactTriggered` and `CompactionComplete` events emitted in correct order |
| `src/frontend/tui/tui_app.rs` | `handle_agent_event` for `AutoCompactTriggered` pushes info conversation entry with usage message |
| `src/frontend/stdout.rs` | `run_text` and `collect_response` handle `AutoCompactTriggered` without error |

### Risks

1. **Consecutive compaction storm** — mitigated by `last_auto_compacted` flag that skips compaction and emits a warning on the second consecutive trigger.
2. **Compact failure mid-stream** — if `compact()` fails during auto-compact, the error is already handled by the existing `CompactionComplete { is_error: true }` path.
3. **Config validation** — `max_context_window_len` must be a non-negative integer; `0` disables the feature.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implemented the auto-compact feature per the plan in PR 201:

**Config (`src/config.rs`):**
- Added `max_context_window_len: u32` (default 0 = disabled) to `ToolsConfig`
- Updated config template with documented entry
- 3 new tests + updates to 2 existing tests

**Types (`src/types.rs`):**
- Added `AutoCompactTriggered { current_tokens, threshold }` variant to `AgentEvent`

**Agent core (`src/agent/mod.rs`, `src/agent/compact.rs`):**
- Added `max_context_window_len` and `last_auto_compacted: Arc<AtomicBool>` fields to `Agent`
- Accumulate `input_tokens` from `StreamEvent::Usage` in the spawned task
- After `ResponseComplete` with no tool calls: check threshold, emit `AutoCompactTriggered` or `Warn`
- Consecutive-compaction guard: if previous turn auto-compacted, emit `Warn` instead of re-triggering
- Guard resets when tokens drop below threshold
- Extracted `compact_with()` free function from `Agent::compact()` for reuse
- 6 new agent-level tests (threshold zero, below threshold, above threshold, consecutive guard, guard reset, `with_tool_config`)

**Frontends:**
- TUI (`tui_app.rs`): `AutoCompactTriggered` handled in `run_app` — pushes info entry, sets `Compacting` state, spawns compact task
- Stdout (`stdout.rs`): silently absorbs in both `run_text` and `collect_response`
- Spawner (`spawner.rs`): silently absorbs in `run_headless`
- 1 new TUI test, 2 new stdout tests

**All 816 tests pass.** Clippy clean. Insta snapshots updated.

**Commit:** `bc668b4`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

### Critical

1. **Tool-use turns never trigger auto-compact** — The auto-compact check (`accumulated_input_tokens > max_context_window_len`) is nested inside `if tool_calls.is_empty()` in `send()`. When the LLM emits tool calls, the branch is skipped entirely, tool calls execute, and the outer loop continues without ever evaluating the threshold. This means agentic workflows — the primary context-bloat scenario — can accumulate tokens indefinitely without triggering auto-compact. The 6 agent tests all use `text_with_usage_response` which produces zero tool calls. **Confidence: 95**

2. **TUI `run_app` auto-compact state machine has zero integration tests** — The only TUI test (`handle_agent_event_auto_compact_triggered_is_noop`) deliberately asserts the event is *not* handled in `handle_agent_event`, because the real handling lives in `run_app`. Yet `run_app`'s auto-compact flow — pushing info entry, setting `Compacting` state, spawning compact task, receiving `CompactionComplete`, reloading history on success, showing error on failure — has no test coverage at all. **Confidence: 90**

### Important

3. **`last_auto_compacted` not reset on `load_session()`** — `Agent::load_session()` replaces history and session but never resets `last_auto_compacted`. If session A triggers auto-compact (guard = true), switching to session B leaks the flag. Session B's first above-threshold turn incorrectly emits `Warn` instead of `AutoCompactTriggered`. **Confidence: 88**

4. **Failed compaction leaves `last_auto_compacted = true`, blocking retries** — When `compact()` fails, `CompactionComplete { is_error: true }` correctly transitions the TUI back to `Input`, but `last_auto_compacted` remains `true` on the Agent. Since compaction failed, context is still above threshold, so the flag will never reset via the `else` branch. Subsequent turns emit only `Warn` instead of retrying auto-compact. **Confidence: 90**

5. **Consecutive guard only tested for 2 turns** — The `auto_compact_consecutive_guard_emits_warning` test validates exactly one trigger + one warning. The 3+ turn pattern (trigger → warn → warn) is unverified, as is the realistic scenario where compaction succeeds but the next turn is still above threshold. **Confidence: 92**

6. **Multiple `Usage` events in a single turn not tested for accumulation** — `accumulated_input_tokens` sums all `Usage` events via `saturating_add`, but every test backend emits exactly one `Usage` event. A backend emitting `[Usage(30_000), TextDelta, Usage(30_000)]` should accumulate to 60_000, but this path is untested. **Confidence: 85**

7. **No way to cancel in-progress compaction from TUI except quitting** — While in `AppState::Compacting`, all keyboard input is silently discarded except Ctrl-C which exits the entire app. If the compaction sub-agent hangs, the user has no Esc/cancel option. **Confidence: 88**

8. **Compaction task not cancelled on session switch** — The `SessionPickerAction::Select` handler does not abort `app.compaction_task`. If the user switches sessions while compacting, the old task continues and its `CompactionComplete` event overwrites the TUI with stale (old session's compacted) history. **Confidence: 75**

9. **`compact_tx.send()` failure silently discards error** — If `event_rx` is closed when the compaction task tries to send `CompactionComplete`, `let _ = compact_tx.send(event).await` silently discards the error. The TUI would stay stuck in `Compacting` state forever. **Confidence: 80**

### Suggestions

10. **Config edge values untested** — No test for `max_context_window_len = 1`, `u32::MAX`, or the exact-boundary case where `accumulated_input_tokens == threshold` (the `>` vs `>=` boundary). **Confidence: 82**

11. **Event ordering (`ResponseComplete` before `AutoCompactTriggered`) not asserted** — Tests check presence of both events but not relative ordering. Frontend behavior could be fragile if ordering changes. **Confidence: 80**

12. **`compact_with()` free function extraction may be over-abstracted** — The free function takes 4 `Arc` parameters that mirror `Agent` fields. Currently only called from `Agent::compact()`. The extraction adds complexity without a second call site. If no future call site is planned, keeping the logic in `Agent::compact()` would be simpler. **Confidence: 75**

13. **`accumulated_input_tokens` saturation edge case** — `saturating_add` caps at `u32::MAX`. In the extreme case of summing multiple Usage events, `AutoCompactTriggered { current_tokens: u32::MAX }` would display a confusing message. Practically impossible given LLM context sizes, but worth noting. **Confidence: 70**

### Strengths

- Clean separation of concerns: agent decides *when* to compact, frontends decide *how* to display
- Consecutive-compaction guard is a thoughtful safety mechanism to prevent infinite loops
- Guard correctly resets when tokens drop below threshold
- Config default of `0` (disabled) means no behavior change for existing users
- `compact_with()` extraction enables future reuse without coupling to `&self`
- Comprehensive agent-level tests covering threshold zero, below/above threshold, and guard reset scenarios

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

[Review comment](https://github.com/Fizzizist/illustrious-manager/pull/201#issuecomment-4412885083)

### Classifications

|---|---|---|
| 1. Tool-use turns skip auto-compact | **Genuine issue** | Auto-compact check is inside `if tool_calls.is_empty()` — tool-use loops never check the threshold, causing unbounded context growth in the primary agentic scenario |
| 2. TUI run_app auto-compact flow untested | **Deferred** | Valid test gap but TUI event loops are inherently hard to unit-test; underlying logic is well-tested |
| 3. `last_auto_compacted` not reset on `load_session()` | **Genuine issue** | Confirmed: `load_session` replaces history/session but never resets `last_auto_compacted`, leaking the flag across sessions |
| 4. Failed compaction leaves `last_auto_compacted = true` | **Genuine issue** | If `compact()` fails, the guard stays true and all subsequent turns emit Warn instead of retrying |
| 5. Consecutive guard only tested for 2 turns | **Nitpick** | A simple boolean toggle; 3+ turns is logically identical (flag stays true). No behavioral difference |
| 6. Multiple Usage events per turn untested for accumulation | **Nitpick** | Trivial arithmetic with `saturating_add`; single-turn accumulation is tested. Multiple Usage events per turn is a streaming protocol detail |
| 7. No way to cancel in-progress compaction from TUI | **Genuine issue** | Compacting state only allows Ctrl-C (exits entire app). Esc is ignored. Users cannot cancel compaction without quitting |
| 8. Compaction task not cancelled on session switch | **Genuine issue** | Session switch doesn't abort `compaction_task`. A compaction completing after session switch would corrupt the new session's history |
| 9. `compact_tx.send()` failure silently discarded | **Nitpick** | The `let _ =` pattern is used throughout the codebase. If `event_rx` is closed, the TUI is already shutting down via Ctrl-C. Stuck-in-Compacting can't happen in practice |
| 10. Config edge values untested | **Nitpick** | `>` vs `>=` is a deliberate design choice. Values 1 and u32::MAX don't affect real usage |
| 11. Event ordering not asserted in tests | **False positive** | Ordering is guaranteed by sequential code execution, not a race condition |
| 12. `compact_with()` over-abstracted | **Nitpick** | Single call site today, but extraction enables testability and future callers. Negligible cost |
| 13. Saturation edge case produces confusing message | **False positive** | u32::MAX tokens (~4 billion) is unreachable in practice. `saturating_add` prevents overflow panics as intended |

### Action Plan

1. **Move auto-compact check outside `if tool_calls.is_empty()`** — The threshold check should run at the end of every loop iteration (after tool results are persisted), not only on turns with no tool calls. This is the most impactful fix since agentic workflows are the primary context-bloat scenario.

2. **Reset `last_auto_compacted` on `load_session()`** — Add `self.last_auto_compacted.store(false, Ordering::SeqCst);` to `Agent::load_session()`.

3. **Reset `last_auto_compacted` on failed compaction** — When `CompactionComplete { is_error: true }` is received, the TUI should reset the agent's guard flag. Add a method like `Agent::reset_auto_compacted_flag()` and call it from the error branch, or have the agent itself reset the flag when compaction fails.

4. **Add Esc handler for `AppState::Compacting`** — Allow users to cancel an in-progress compaction by aborting `compaction_task` and transitioning back to `AppState::Input`.

5. **Abort `compaction_task` on session switch** — Before `agent.load_session(session)`, abort any existing compaction task and set `app.compaction_task = None`.

6. *(Deferred)* Consider integration tests for the TUI auto-compact flow — architecturally hard to unit-test, but the orchestration logic in `run_app` could be extracted into a testable function.

---
*Assessment by mach6*

Fizzizist:
## Review Fix Push

Addressed all genuine findings from the review:

1. **Finding 1 (Critical)**: Moved auto-compact check outside `tool_calls.is_empty()` — now triggers after every completed iteration (both text-only and tool-use turns). Added regression test `auto_compact_triggers_on_tool_use_turn`.

2. **Finding 3**: Reset `last_auto_compacted` in `Agent::load_session()` — prevents flag leaking across sessions. Added test `load_session_resets_auto_compact_flag`.

3. **Finding 4**: Added `Agent::reset_auto_compact_flag()` — called by TUI when `CompactionComplete { is_error: true }`, allowing subsequent turns to retry auto-compact.

4. **Finding 7**: Added `Esc` handler for `AppState::Compacting` — aborts `compaction_task` and transitions back to `Input`.

5. **Finding 8**: Abort `compaction_task` on session switch — prevents stale `CompactionComplete` from corrupting new session's history.

Nitpick/deferred items (findings 5, 6, 9–13) acknowledged but not addressed — they represent edge cases or stylistic preferences with minimal real-world impact.

**Commit:** `3fb9f3d`

---
*Progress tracked by mach6*

Fizzizist:
## Refactor: CompactionConfig Section

Moved `max_context_window_len` out of `[tools]` and `compaction_role` off the top level into a new `[compaction]` config section:

```toml
[compaction]
role = "compaction"           # compaction sub-agent model role
max_context_window_len = 0    # auto-compact threshold (0 = disabled)
```

Changes:
- New `CompactionConfig` struct with `role` and `max_context_window_len`
- `Agent::with_compaction_config()` method replaces the `ToolsConfig` path
- All test fixtures updated
- Config template updated
- All 835 tests pass

**Commit:** `9753e75`

---
*Progress tracked by mach6*

Fizzizist:
## Bug Fix: Auto-compact not triggering at 12.8k/15k threshold

**Root cause**: Two bugs working together:

1. **Config not wired to Agent** (`f63966c`): `with_compaction_config()` was never called on the Agent in `main.rs` or `spawn_agent_with_selection()`. The `max_context_window_len` field stayed at 0 (disabled). This is why the 12.8k session never triggered — the agent didn't know about the 15k threshold.

2. **Accumulated vs peak tokens** (`6cc8c74`): The original implementation used `saturating_add` to sum all `Usage.input_tokens` values. The API reports the **total** context size per request, not incremental. In a tool-use loop: `12k + 15k = 27k` when the actual context is only 15k. Changed to track `peak_input_tokens` (max) instead. This is why the 190k session worked (single huge request) but typical sessions didn't (multiple iterations inflating the sum).

Both fixes are now in with regression tests.

---
*Progress tracked by mach6*

PR #215, opened 2026-05-21
fix: route diagnostic output through logging module (issue 207)

### LINKED ISSUE #207: BUG: warning message printing to stdout in TUI

Fizzizist:
Closes #207

Purge `eprintln!`/`println!` from `src/` and route all diagnostic output through the `logging` module. Add a clippy-enforced policy so this cannot recur. Two pre/post-TUI user-facing notices (Session ID at exit, "Created default config" on first run) and the single-shot interactive confirmation prompt are kept user-visible via `write!(io::stderr(), ...)`, which the new lint does not flag.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,6 +18,7 @@ These are _strict_ policies that must be followed by all engineers and developer
 - Always run `cargo fmt` before committing code.
 - Always run `cargo clippy` before committing code.
 - Keep code comments to a minimum. Only comment in cases where something is unable to be gleaned from the code itself.
+- The use of `eprintln!`, `println!`, `eprint!`, `print!`, and `dbg!` is forbidden in `src/`. All diagnostic output must go through the `logging` module. Direct stdout/stderr writes corrupt the TUI. Genuinely interactive prompts or end-of-process user-facing notices (e.g. single-shot TTY confirmation, session-id epilogue) may use `write!(io::stderr(), ...)` and must include a comment justifying why.
 
 ### Testing
 
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -144,28 +144,6 @@ impl BackendFactory {
 }
 
 #[cfg(test)]
-pub(crate) mod test_support {
-    use super::*;
-    use futures::stream;
-
-    /// A minimal `LlmBackend` implementation that returns an empty stream.
-    /// Shared across test modules to avoid duplication.
-    #[allow(dead_code)]
-    pub struct MockBackend;
-
-    #[async_trait]
-    impl LlmBackend for MockBackend {
-        async fn send_message(
-            &self,
-            _messages: &[Message],
-            _config: &RequestConfig,
-        ) -> Result<BoxStream<Result<StreamEvent>>> {
-            Ok(Box::pin(stream::empty()))
-        }
-    }
-}
-
-#[cfg(test)]
 mod tests {
     use crate::config::CompactionConfig;
     use anyhow::Result;
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -150,6 +150,7 @@ pub(crate) mod test_support {
 
     /// A minimal `LlmBackend` implementation that returns an empty stream.
     /// Shared across test modules to avoid duplication.
+    #[allow(dead_code)]
     pub struct MockBackend;
 
     #[async_trait]
--- a/src/backend/vertex.rs
+++ b/src/backend/vertex.rs
@@ -371,26 +371,37 @@ impl LlmBackend for VertexBackend {
     }
 }
 
+fn filter_content_blocks<'a>(
+    blocks: &'a [crate::types::ContentBlock],
+    warnings: &mut Vec<String>,
+) -> Vec<&'a crate::types::ContentBlock> {
+    blocks
+        .iter()
+        .filter(|block| {
+            if let crate::types::ContentBlock::Thinking { signature, .. } = block
+                && signature.is_empty()
+            {
+                warnings.push(
+                    "dropping thinking block with empty signature from history \
+                     (corrupted session data); it cannot be replayed."
+                        .to_string(),
+                );
+                return false;
+            }
+            true
+        })
+        .collect()
+}
+
 fn build_request_body(messages: &[Message], config: &RequestConfig) -> Result<serde_json::Value> {
     let messages_json: Vec<serde_json::Value> = messages
         .iter()
         .filter_map(|m| {
-            let content: Vec<&crate::types::ContentBlock> = m
-                .content
-                .iter()
-                .filter(|block| {
-                    if let crate::types::ContentBlock::Thinking { signature, .. } = block
-                        && signature.is_empty()
-                    {
-                        eprintln!(
-                            "WARNING: dropping thinking block with empty signature from history \
-                             (corrupted session data); it cannot be replayed."
-                        );
-                        return false;
-                    }
-                    true
-                })
-                .collect();
+            let mut warnings = Vec::new();
+            let content = filter_content_blocks(&m.content, &mut warnings);
+            for w in warnings {
+                crate::logging::log_warn(&w);
+            }
             if content.is_empty() {
                 return None;
             }
@@ -1223,5 +1234,29 @@ mod tests {
             "message with only empty-signature thinking blocks must be dropped"
         );
         assert_eq!(msgs[0]["role"], "user");
+    }
+
+    #[test]
+    fn filter_content_blocks_drops_empty_signature_thinking_and_records_warning() {
+        use crate::types::ContentBlock;
+        let blocks = vec![
+            ContentBlock::Thinking {
+                text: "reasoning".to_string(),
+                signature: String::new(), // empty → should be dropped
+            },
+            ContentBlock::Text("hello".to_string()),
+        ];
+        let mut warnings = Vec::new();
+        let filtered = filter_content_blocks(&blocks, &mut warnings);
+        assert_eq!(filtered.len(), 1, "only the Text block should remain");
+        assert!(
+            matches!(filtered[0], ContentBlock::Text(_)),
+            "remaining block should be Text"
+        );
+        assert_eq!(warnings.len(), 1, "one warning should be recorded");
+        assert!(
+            warnings[0].contains("empty signature"),
+            "warning should mention empty signature"
+        );
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,7 +1,8 @@
 use dirs;
 use std::collections::BTreeMap;
-use std::fmt::Write;
+use std::fmt::Write as FmtWrite;
 use std::fs;
+use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result, bail};
@@ -369,7 +370,12 @@ pub fn load_config(custom_path: Option<&Path>) -> Result<AppConfig> {
         }
         fs::write(&path, CONFIG_TEMPLATE)
             .with_context(|| format!("Failed to write default config: {}", path.display()))?;
-        eprintln!("Created default config at: {}", path.display());
+        // Intentional stderr write: config creation notice runs before the TUI starts.
+        writeln!(
+            io::stderr(),
+            "Created default config at: {}",
+            path.display()
+        )?;
     }
 
     load_config_from_path(&path)
--- a/src/frontend/stdout.rs
+++ b/src/frontend/stdout.rs
@@ -114,7 +114,7 @@ async fn run_text<W: Write, R: BufRead>(
                 break;
             }
             AgentEvent::Error(msg) => {
-                eprintln!("\nError: {}", msg);
+                crate::logging::log_error(&format!("\nError: {}", msg));
                 anyhow::bail!("LLM error: {}", msg);
             }
             AgentEvent::ToolUseReceived {
@@ -363,14 +363,22 @@ fn handle_confirmation<R: BufRead>(
     input: &serde_json::Value,
 ) -> Result<()> {
     if !is_tty {
-        eprintln!(
+        // Intentional stderr write: stdout.rs runs outside the TUI; this is a fatal user-facing error.
+        writeln!(
+            io::stderr(),
             "Error: tool confirmation required but stdin is not a TTY (tool: {})",
             name
-        );
+        )?;
         let _ = confirm_tx.unbounded_send(ConfirmationResponse::Rejected);
         anyhow::bail!("tool confirmation required in non-TTY mode");
     }
-    eprint!("Allow tool '{}' with input {}? [y/N] ", name, input);
+    // Intentional stderr write: interactive TTY confirmation prompt runs outside the TUI.
+    write!(
+        io::stderr(),
+        "Allow tool '{}' with input {}? [y/N] ",
+        name,
+        input
+    )?;
     io::stderr().flush()?;
     let mut response = String::new();
     stdin.read_line(&mut response)?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,5 @@
+#![deny(clippy::print_stderr, clippy::print_stdout, clippy::dbg_macro)]
+
 pub mod agent;
 pub mod backend;
 pub mod config;
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -4,24 +4,19 @@ use std::path::PathBuf;
 use std::sync::{Mutex, OnceLock};
 
 use crate::types::AgentEvent;
-use anyhow::Result;
+use anyhow::{Context, Result};
 
 static GLOBAL_WRITER: OnceLock<Mutex<Option<BufWriter<File>>>> = OnceLock::new();
 
 /// Initialise the global debug-log writer once.
 pub fn init_global(log_path: Option<PathBuf>) -> Result<()> {
-    GLOBAL_WRITER.get_or_init(|| {
-        let writer = match log_path {
-            Some(path) => {
-                let file = File::create(&path).unwrap_or_else(|_| {
-                    panic!("Failed to create debug log file at {}", path.display())
-                });
-                Some(BufWriter::new(file))
-            }
-            None => None,
-        };
-        Mutex::new(writer)
-    });
+    let writer = match log_path {
+        Some(path) => Some(BufWriter::new(File::create(&path).with_context(|| {
+            format!("Failed to create debug log file at {}", path.display())
+        })?)),
+        None => None,
+    };
+    let _ = GLOBAL_WRITER.set(Mutex::new(writer));
     Ok(())
 }
 
@@ -829,5 +824,43 @@ mod tests {
         logger.log_error("x").expect("Failed to log error");
         logger.log_info("x").expect("Failed to log info");
         // No panic = success.
+    }
+
+    #[test]
+    fn global_writer_path_and_free_log_event() {
+        // GLOBAL_WRITER is a process-global OnceLock. This is the only test that calls
+        // init_global; a single test function avoids races with other test threads.
+        // If the OnceLock was already set (e.g. by a prior call in the same process),
+        // init_global silently no-ops — the "no panic" guarantee still holds.
+        use crate::types::AgentEvent;
+        use tempfile::TempDir;
+
+        let dir = TempDir::new().expect("temp dir");
+        let path = dir.path().join("global-test.log");
+
+        init_global(Some(path.clone())).expect("init_global should succeed");
+
+        log_warn("global warn");
+        log_error("global error");
+        log_event(&AgentEvent::ResponseComplete("assistant text".to_string()));
+        log_event(&AgentEvent::Error("some error".to_string()));
+        flush();
+
+        // Only assert file contents when this process's init_global call won the OnceLock.
+        if path.exists() {
+            let content = std::fs::read_to_string(&path).expect("read log");
+            assert!(content.contains("[WARN]"), "expected [WARN] tag");
+            assert!(content.contains("global warn"), "expected warn message");
+            assert!(content.contains("[ERROR]"), "expected [ERROR] tag");
+            assert!(content.contains("global error"), "expected error message");
+            assert!(
+                content.contains("[ASSISTANT RESPONSE]"),
+                "expected [ASSISTANT RESPONSE] tag"
+            );
+            assert!(
+                content.contains("assistant text"),
+                "expected assistant text"
+            );
+        }
     }
 }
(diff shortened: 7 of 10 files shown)


Fizzizist:
## Implementation Plan

### Problem
`eprintln!`/`println!` written directly to stdio corrupt the TUI (terminal is in alternate-screen/raw mode). The reported symptom is `vertex.rs:385` printing a "dropping thinking block" warning over the TUI canvas. Owner has scoped the fix broadly: silence all such sites unless `--debug` is set, and add a policy + lint so this cannot recur.

### Deliverables

**1. Global logger plumbing** in `src/logging.rs`:
- `static GLOBAL_WRITER: OnceLock<Mutex<Option<BufWriter<File>>>>`.
- `init_global(log_path: Option<PathBuf>) -> Result<()>` — called once from `main.rs` with `Some(path)` when `--debug`, `None` otherwise.
- Free functions: `log_warn(&str)`, `log_error(&str)`, `log_info(&str)`, plus free-function variants of the existing `log_user_input`/`log_tool_use`/`log_tool_result`/`log_assistant_response`/`log_usage`/`log_config`/`log_event`/`flush`. All are no-ops when the global is uninitialised or holds `None`.
- The existing instance-method `Logger` is retained as a thin handle whose `new()` delegates to `init_global` and whose methods forward to the free functions. This keeps the frontend/main API surface unchanged for minimum diff.

**2. Call-site replacements:**

Routed through `logging` (silent without `--debug`):
| File:line | Replacement |
|---|---|
| `src/backend/vertex.rs:385` | Refactor (see test plan), then `logging::log_warn(...)` from production caller. |
| `src/tools/mod.rs:146` | `logging::log_warn(...)` |
| `src/frontend/stdout.rs:117` (error path) | `logging::log_error(...)` + existing `anyhow::bail!` continues to surface the failure as the process exit reason. |
| `src/main.rs:228` (checkpoint_session failure) | `logging::log_error(...)` |
| `src/main.rs:234` (cleanup_empty_session failure) | `logging::log_error(...)` |

Kept user-visible via `write!(io::stderr(), ...)` with a comment justifying intentional terminal I/O (these sites run before the TUI starts or after it tears down, so they do not corrupt the display):
| File:line | Replacement |
|---|---|
| `src/main.rs:237` (Session ID after frontend exits) | `writeln!(io::stderr(), "Session ID: {}", ...)?` |
| `src/config.rs:372` (Created default config notice) | `writeln!(io::stderr(), "Created default config at: {}", ...)?` |
| `src/frontend/stdout.rs:366,373` (TTY confirmation prompt) | `writeln!(io::stderr(), ...)` / `write!(io::stderr(), ...)` |

**3. CLAUDE.md policy update** (Coding section):
> The use of `eprintln!`, `println!`, `eprint!`, `print!`, and `dbg!` is forbidden in `src/`. All diagnostic output must go through the `logging` module. Direct stdout/stderr writes corrupt the TUI. Genuinely interactive prompts or end-of-process user-facing notices (e.g. single-shot TTY confirmation, session-id epilogue) may use `write!(io::stderr(), ...)` and must include a comment justifying why.

**4. Lint enforcement** — Add to `src/main.rs` and `src/lib.rs`:
```rust
#![deny(clippy::print_stderr, clippy::print_stdout, clippy::dbg_macro)]
```
These are clippy restriction lints. `write!(io::stderr(), ...)` is not flagged, so the four deliberate user-facing sites remain clean without `#[allow]` escape hatches.

**5. Regression test** — Refactor `vertex::build_request_body` to extract the per-message content filter into a testable helper:
```rust
fn filter_content_blocks<'a>(blocks: &'a [ContentBlock], warnings: &mut Vec<String>) -> Vec<&'a ContentBlock>
```
The production `build_request_body` collects warnings into a local `Vec<String>` and forwards each to `logging::log_warn` after iteration. Unit test (in the existing `mod tests` of `vertex.rs`) builds a `Message` with one empty-signature `Thinking` block and one `Text` block, calls the helper, asserts (a) the Thinking block is excluded, (b) `warnings.len() == 1`, (c) the warning text matches the documented prefix.

### Acceptance Criteria
- `cargo clippy --all-targets -- -D warnings` passes (the surviving `write!(io::stderr(), ...)` calls are not flagged by `clippy::print_stderr`).
- `cargo test` passes, including the new regression test for the empty-signature filter.
- Running `cargo run` without `--debug`: TUI displays cleanly during a session; the previously-corrupting `eprintln!` sites (vertex thinking-block warning, tools allowlist warning, single-shot error path) produce zero terminal output mid-session.
- Running the REPL without `--debug` and exiting cleanly still prints exactly one line after the TUI restores: `Session ID: <uuid>`.
- First-run with no existing config still prints `Created default config at: <path>` before the TUI starts.
- Running `cargo run -- --debug`: the routed-through-logging messages appear in the debug log file with their respective `[WARN]`/`[ERROR]` tags.
- CLAUDE.md contains the new Coding policy line.

### Files to Create or Modify
- **Modify** `src/logging.rs` — add global writer + free functions; rewire `Logger` instance methods to delegate.
- **Modify** `src/main.rs` — call `logging::init_global` once after parsing `--debug`; replace two error-path `eprintln!` sites with `logging::log_error`; switch Session ID epilogue to `writeln!(io::stderr(), ...)` with comment; add `#![deny(...)]` lint attributes.
- **Modify** `src/lib.rs` — add `#![deny(...)]` lint attributes.
- **Modify** `src/backend/vertex.rs` — extract `filter_content_blocks` helper; replace `eprintln!` with `logging::log_warn`; add regression test in existing `mod tests`.
- **Modify** `src/tools/mod.rs` — replace `eprintln!` in `into_filtered` with `logging::log_warn`.
- **Modify** `src/config.rs` — switch "Created default config at" notice to `writeln!(io::stderr(), ...)` with comment.
- **Modify** `src/frontend/stdout.rs` — replace error-path `eprintln!` with `logging::log_error`; replace interactive-prompt `eprintln!`/`eprint!` with `write!(io::stderr(), ...)`.
- **Modify** `CLAUDE.md` — add Coding policy line under existing `### Coding` section.

### Testing Approach
- **New unit test** in `src/backend/vertex.rs` `mod tests`: `filter_content_blocks_drops_empty_signature_thinking_and_records_warning`. Verifies the regression directly without needing to capture global logger state.
- **Existing unit tests** in `src/logging.rs` `mod tests` continue to pass against the rewired `Logger` instance API.
- **New unit test** in `src/logging.rs` `mod tests`: `log_free_functions_are_noop_when_global_uninitialised` — calls each free function and asserts no panic.
- **Manual smoke tests** (acceptance verification): (a) run REPL without `--debug` and confirm no mid-session stdio corruption; (b) confirm Session ID and config-created notices still appear; (c) run with `--debug` and inspect the log file for the previously-printed lines.

### Risks and Open Questions
- **Global singleton across tests**: `OnceLock<Mutex<...>>` means parallel tests cannot each initialise their own log file. The instance-method `Logger::new(Some(path))` continues to open its own dedicated `BufWriter<File>` (current behaviour preserved); production main calls only `init_global` so there is no double-open. File-content tests stay on the instance API.
- **Lint scope**: `clippy::print_stderr`/`print_stdout` are restriction-level lints. Verify with `cargo clippy --all-targets` that no macro expansions in dependencies cross our boundary (they shouldn't — these lints only fire on calls in the current crate).
- **Behavioural delta**: The two `main.rs` error-path messages (`checkpoint_session failed`, `cleanup_empty_session failed`) become invisible without `--debug`. This matches the owner's stated intent ("zero terminal output unless `--debug`"). The original `cleanup_empty_session` site was already gated on `cli.debug` so no observable change there; the `checkpoint_session` site loses its unconditional stderr line — acceptable per the resolution.

---
*Plan created by mach6*

Fizzizist:
## Progress Update

Implemented the full plan from the plan comment.

**Changes:**

- `src/logging.rs` — Added `GLOBAL_WRITER: OnceLock<Mutex<Option<BufWriter<File>>>>`, `init_global()`, and 11 free functions (`log_warn`, `log_error`, `log_info`, `log_user_input`, `log_tool_use`, `log_tool_result`, `log_assistant_response`, `log_usage`, `log_config`, `log_event`, `flush`). Logger instance methods delegate to free functions when `log_file` is `None` (production thin-handle), write to own file when `Some` (test backward compat). Two new unit tests added.
- `src/backend/vertex.rs` — Extracted `filter_content_blocks` helper; replaced `eprintln!` with `crate::logging::log_warn`; added regression test `filter_content_blocks_drops_empty_signature_thinking_and_records_warning`.
- `src/tools/mod.rs` — `eprintln!` in `into_filtered` → `crate::logging::log_warn`.
- `src/frontend/stdout.rs` — Error path `eprintln!` → `logging::log_error`; TTY confirmation prompt sites → `write!(io::stderr(), ...)` with justification comments.
- `src/config.rs` — `eprintln!` for config creation notice → `writeln!(io::stderr(), ...)` with comment.
- `src/main.rs` — Calls `logging::init_global` at startup; two error-path `eprintln!` sites → `logging::log_error`; Session ID epilogue → `writeln!(io::stderr(), ...)` with comment; deny lint attributes added.
- `src/lib.rs` — `#![deny(clippy::print_stderr, clippy::print_stdout, clippy::dbg_macro)]` added.
- `CLAUDE.md` — New Coding policy prohibiting `eprintln!`/`println!`/`eprint!`/`print!`/`dbg!` in `src/`.
- `src/backend/mod.rs`, `src/types.rs` — Fixed two pre-existing clippy violations that blocked `cargo clippy --all-targets -- -D warnings`.

**Verification:** `cargo clippy --all-targets -- -D warnings` passes. All tests pass (893 lib + integration tests).

**Commit:** `ccb6b63`

---
*Progress tracked by mach6*

Fizzizist:
## Code Review

### Critical

None.

### Important

**1. `init_global` panics on disk error, violating the strict no-panic policy** — `src/logging.rs:13-23`
`File::create(&path).unwrap_or_else(|_| panic!("..."))` is functionally identical to `.unwrap()`, which CLAUDE.md forbids "under all circumstances." If the debug log path is invalid (permissions, full disk, missing parent dir), the binary panics on startup. The function signature `pub fn init_global(...) -> Result<()>` is also a lie — it always returns `Ok(())` because the panic preempts any error path. Fix by hoisting the fallible `File::create` outside `get_or_init`:
```rust
pub fn init_global(log_path: Option<PathBuf>) -> Result<()> {
    let writer = match log_path {
        Some(path) => Some(BufWriter::new(File::create(&path)
            .with_context(|| format!("Failed to create debug log file at {}", path.display()))?)),
        None => None,
    };
    let _ = GLOBAL_WRITER.set(Mutex::new(writer));
    Ok(())
}
```
Flagged by code-reviewer, error-auditor, completeness-checker, and simplifier (4/5 agents independently).

**2. `tools/mod.rs:146` warning string contains 14 literal spaces from indentation** — `src/tools/mod.rs:146-149`
The continuation line is indented to align with code, but `\`-continuation preserves the literal whitespace. The rendered log line is:
```
agent tool allowlist [...] matched no registered tools;              sub-agent will run with an empty tool set
```
Fix by collapsing to a single-line format string or unindenting the continuation.

**3. `BufWriter` in static `OnceLock` is never dropped — buffered writes may be lost on exit** — `src/logging.rs:7`
`static OnceLock` items are never dropped by Rust. The old `impl Drop for Logger` was removed in this PR; nothing replaces it. End-of-session log entries (e.g., the `checkpoint_session` / `cleanup_empty_session` error paths added in `main.rs:233,238`) may sit in the user-space buffer and never reach disk before `main` returns. Fix by calling `logging::flush()` (already exists at `logging.rs:187`) immediately before the `Session ID` epilogue in `main.rs:241`.

**4. Mutex poisoning silently disables all logging permanently** — `src/logging.rs:56-58`
`with_global_writer` catches `lock()` errors with `Err(_) => return`. If any closure ever panics holding the guard, every subsequent `log_warn`/`log_error`/etc. silently no-ops with zero indication. Standard recovery is to call `PoisonError::into_inner()` — the underlying `BufWriter` is still valid.

### Suggestions

**5. `write_or_delegate` is unnecessary indirection** — `src/logging.rs:223-235`
Each Logger method passes two closures that duplicate the same format string. Inlining `if let Some(ref mut w) = self.log_file { writeln!(...) } else { free_fn(); Ok(()) }` is shorter and clearer per method, and removes ~10 lines of helper + ~9 closure pairs.

**6. `Logger::write_event` duplicates the entire `AgentEvent` match against free `log_event`** — `src/logging.rs:96-150` vs `logging.rs:315-355`
Two ~55-line match statements over the same enum. The free version locks the mutex N times per event (once per delegated sub-helper call); the instance version locks once. Collapse the free function to:
```rust
pub fn log_event(event: &AgentEvent) {
    with_global_writer(|w| { let _ = Logger::write_event(w, event); });
}
```
Eliminates ~50 lines and unifies behaviour (single lock per event).

**7. Logger instance is never used with a real file in production** — `src/main.rs:181-184`
In `--debug` mode, `init_global(Some(path))` is followed immediately by `Logger::new(None)`. The Logger's `log_file` is always `None`, so every instance call delegates to the free functions. The instance API exists solely for the test backward-compat path. Either drop the `Option<Logger>` plumbing entirely in production (call free functions directly from frontends) or annotate the API as "test isolation only."

**8. `config.rs:373` first-run notice can abort the binary on a broken stderr pipe** — `src/config.rs:370-378`
`writeln!(io::stderr(), "Created default config at: {}", path.display())?` propagates IO errors via `?`. If stderr is closed/redirected to a broken pipe, the binary aborts before the TUI ever starts. Replace with `let _ = writeln!(...);` — this is purely an FYI notice.

**9. Free-function log errors are silently discarded** — `src/logging.rs:62-103` (every `log_*` helper)
All free functions use `let _ = writeln!(w, ...)`. IO failures on the debug log (deleted file, full disk) are invisible. Acceptable for debug-only logging, but worth a one-shot stderr fallback (`AtomicBool` gate) for visibility on first failure.

**10. Test coverage gaps for the new free-function API**
- `init_global` itself has zero direct tests (only used in `main.rs:179`). Cover success, `None` path, and double-init semantics.
- Free function `log_event` is never called directly in tests — only `Logger::log_event`. Different code path (recursive delegation through free helpers).
- Free function `flush()` is never called directly in tests.
- `Logger::new(None)` delegation test (`logger_none_delegates_to_free_functions`) covers only `log_warn`/`log_error`/`log_info`. Missing: `log_user_input`, `log_tool_use`, `log_tool_result`, `log_assistant_response`, `log_usage`, `log_config`, `log_event`, `flush`.

**11. Logging tests leak files to `/tmp/illustrious-manager-logging-tests/`** — `src/logging.rs:670-680`
Old tests used `tempfile::TempDir` (auto-cleanup). New tests use manual `/tmp/...-<pid>.log` paths with no cleanup. ~28 log files accumulate per test run. `tempfile` is still in `[dev-dependencies]` for other tests. Reverting to `TempDir` is one line per test.

**12. `GLOBAL_WRITER` introduces order-dependent test fragility** — `src/logging.rs:7,815-829`
`OnceLock` is process-global and survives across `cargo test` parallel test threads. The tests `log_free_functions_are_noop_when_global_not_used` and `logger_none_delegates_to_free_functions` rely on no other test ever calling `init_global`. If a future test ever calls `init_global`, these noop tests silently start passing while testing the wrong branch. Mitigation: assert `GLOBAL_WRITER.get().is_none()` at the top of these tests, or gate on `#[serial_test]`.

**13. Regression test for the original bug doesn't exercise the logging path end-to-end** — `src/backend/vertex.rs:1240`
The new test asserts `filter_content_blocks` returns the warning in a `Vec<String>`. It does not assert that `build_request_body` actually routes the warning through `logging::log_warn` (vs accidentally re-introducing `eprintln!`). The `#![deny(clippy::print_stderr)]` lint provides compile-time enforcement, which is arguably stronger than a runtime test for this specific regression — but the test name implies coverage it doesn't fully provide.

### Strengths

- 9/9 `eprintln!`/`eprint!` sites in the original inventory are accounted for. `rg "eprint|println|dbg!" src/` finds only test fixture strings.
- All `write!(io::stderr(), ...)` survivors carry justifying comments per CLAUDE.md policy.
- `#![deny(clippy::print_stderr, clippy::print_stdout, clippy::dbg_macro)]` in both `lib.rs` and `main.rs` provides compile-time enforcement against recurrence — a stronger guarantee than runtime checks.
- All 874 tests pass; `cargo clippy --all-targets -- -D warnings` is clean.
- The `filter_content_blocks` extraction into a pure helper is a nice testability improvement.
- Pre-existing clippy violations (`backend/mod.rs`, `types.rs`) were fixed as collateral cleanup.

---
*Reviewed by mach6*

Fizzizist:
## Review Assessment

Review: https://github.com/Fizzizist/illustrious-manager/pull/215#issuecomment-4509903465

### Classifications

|---|---|---|
| 1. `init_global` panic | **genuine** | `unwrap_or_else(\|_\| panic!(...))` is functionally a forbidden panic per CLAUDE.md ("should never panic"). `OnceLock::get_or_try_init` is stable since Rust 1.83; the fallible work can also be hoisted outside the closure and stored via `OnceLock::set`. Real policy violation. |
| 2. tools/mod.rs whitespace | **false positive** | Verified by reading source: Rust string-continue (`\` at end of line) strips the newline AND all subsequent whitespace per the Rust Reference. The rendered string has a single space between `tools;` and `sub-agent`, not 14. Reviewer misread the diff. |
| 3. BufWriter never flushed on static drop | **nitpick** | Real but minor: `static OnceLock` is never dropped, but `BufWriter`'s buffer is ≤8KB and the OS flushes page cache on normal exit. The only risk is losing the last few log lines on a hard crash — acceptable for a debug-only log. Adding `logging::flush()` at the end of `main()` is good hygiene but not blocking. |
| 4. Mutex poisoning silently kills logging | **false positive** | All closures passed to `with_global_writer` are `let _ = writeln!(...)` — they cannot panic. There is no realistic path to poison the mutex. Defensive code is fine as-is. |
| 5. `write_or_delegate` indirection | **nitpick** | Pure style. Correct behaviour. Inlining is arguably clearer but not blocking. |
| 6. `log_event` / `write_event` duplication | **nitpick** | Real duplication (~50 lines) but format strings are identical and correctness is maintained. Worthwhile follow-up but out of scope. |
| 7. Logger instance unused in prod with real file | **false positive** | The `Logger` instance with `log_file: None` is still actively used by `main.rs` (`log_config`, `log_user_input`) and threaded into `stdout.rs` via `logger.as_mut()`. The delegation pattern is a deliberate design — instance methods route to free functions which write to `GLOBAL_WRITER`. Works correctly. |
| 8. `config.rs` broken pipe via `?` | **nitpick** | Propagating an IO error when stderr is broken is defensible — the consumer is dead and there's nothing useful to do. Replacing `?` with `let _ =` is a judgement call, not a bug. |
| 9. Free-function log errors swallowed | **deferred** | Documented design choice for a fire-and-forget debug logger. Adding an `AtomicBool` one-shot fallback is a worthwhile follow-up but not blocking. No tracking issue exists. |
| 10a. No tests for `init_global` / global writer path | **genuine** | The primary production code path (debug-mode logging via `GLOBAL_WRITER`) has zero direct test coverage. All existing tests exercise `Logger::new(Some(path))` with a private `BufWriter`, which is the back-compat path that production no longer uses. `init_global` and `with_global_writer` are untested. Process-global `OnceLock` makes this hard but not impossible — a single test that calls `init_global(Some(temp_path))` then `log_warn` and reads the file would close the gap. |
| 10b. No test for `flush()` free function | **nitpick** | Indirectly covered via `Logger::flush()` → `flush()` when `log_file` is `None`. Single direct test would be one line. |
| 10c. No test for free `log_event` | **genuine** | Different code path from `Logger::write_event` (recursive delegation through helpers, N lock acquisitions instead of 1). Should have at least one direct test covering 2–3 variants. |
| 10d. Incomplete delegation coverage on `Logger::new(None)` | **nitpick** | Existing test covers 3/9 delegation methods. Extending to the remaining 6 is trivial but low-value (each delegate is one line). |
| 11. Logging tests leak `/tmp` files | **nitpick** | Verified: `tempfile::TempDir` is still needed by ~10 other test modules and stays in `[dev-dependencies]`. Logging tests not using it leaks ~28 files per run to `/tmp/illustrious-manager-logging-tests/`. Minor disk hygiene; trivial fix. |
| 12. `GLOBAL_WRITER` order-dependent test fragility | **false positive** | Verified via `rg "init_global" src/ tests/`: no test calls `init_global`. The two `Logger::new(None)` delegation tests rely on the OnceLock being either uninitialised or holding `None`. Both states produce the no-op behaviour the tests assert. No actual fragility today. Worth a comment in the test file noting the assumption, but not a current bug. |
| 13. Regression test scope | **nitpick** | The `#![deny(clippy::print_stderr, clippy::print_stdout, clippy::dbg_macro)]` lint is a compile-time guarantee — stronger than a runtime test for this exact regression. The unit test on the pure helper is sufficient. |

**Tally:** 2 genuine + 1 deferred + 8 nitpick + 4 false positive = 15 items (F10 expanded into 4).

### Action Plan

1. **F1** — Fix the panic in `init_global`. Use `OnceLock::get_or_try_init` (stable since Rust 1.83) to propagate `File::create` errors as a real `Err`, or hoist the fallible work outside `get_or_init` and use `OnceLock::set`. Either approach removes the `panic!` and makes the `Result<()>` return type honest.

2. **F10a + F10c** — Add direct tests for the global-writer code path:
   - `init_global(Some(temp_path))` + `log_warn("x")` + read file: verify `[WARN]\nx` is written.
   - `init_global(None)` + `log_warn("x")`: verify no panic, no file created.
   - `log_event(&AgentEvent::ThinkingReceived(...))` against the global writer: verify the free-function `log_event` match arms route correctly.
   - Acknowledge the process-global `OnceLock` constraint: these tests need careful ordering (e.g., one combined test) or `#[serial_test]` if added.

That is all that should block merge. Everything else is style, nice-to-have, or non-issue.

---
*Assessment by mach6*

Fizzizist:
## Progress Update

Addressed the two genuine findings and one hygiene nitpick from the review assessment.

**Finding 1 — `init_global` panic removed:**
Hoisted `File::create` outside `get_or_init`, switched to `OnceLock::set`. The function now returns a real `Err` on disk/permissions failure via `anyhow::Context`. No more `panic!`.

**Finding 10a + 10c — Global-writer tests added:**
Single test `global_writer_path_and_free_log_event` calls `init_global(Some(path))`, then exercises `log_warn`, `log_error`, `log_event` (ResponseComplete + Error variants), and `flush`. Asserts `[WARN]`, `[ERROR]`, `[ASSISTANT RESPONSE]` and matching messages appear in the log file. Guards the assertion with `if path.exists()` to remain correct even if the process-global OnceLock was already set by a prior call.

**Finding 3 (nitpick) — Flush before exit:**
Added `logging::flush()` immediately before the Session ID epilogue in `main.rs`, with a comment explaining the static OnceLock drop constraint.

**Commit:** `e14a1cb`

---
*Progress tracked by mach6*

